PR #35, opened 2021-12-12
feature: Support mixup augmentation for detection

yurkovak:
Add mixup augmentation used for training of yolo from [v6.0](https://github.com/ultralytics/yolov5/releases/tag/v6.0) release (M, etc)
Sums 2 mosaic images with a random ratio in [0, 1]. Labels from both images go into ground truth.

With weight 0.5 it looks like this:


shaydeci:
Closing this PR as I integrated most of it in the transfer learning code.

PR #61, opened 2021-12-27
fix the writing of hyperparams in log, tensorboard and wandb

ofrimasad:
fixes the parameters in log file and tensorboard (and wandb) using a structures json format instead of the flatten format.
also asses a __repr__ to anchors so they will be presented in parameters


sefibk (dismissed):
LGTM

shaydeci (approved):
lgtm

PR #62, opened 2021-12-28
remove max_epochs limitation

lkdci:
remove max_epochs limitation in `TRAINING_PARAM_SCHEMA`
Some training recipe i.e, training on cityscapes dataset requires large number of epochs, see for example [RegSeg paper](https://arxiv.org/pdf/2111.09957v1.pdf) which trains for 1000 epochs


shaydeci (approved):
LGTM

PR #70, opened 2021-12-29
Feature/alg 257 save code wandb

ofrimasad:
adding a new important feature to wandb logger - saving the current code as defined in a config file


shaydeci (approved):
LGTM. Nice feature.

PR #84, opened 2021-12-30
Feature/dle 000 shelfnet head replacement

shaydeci:


tests/integration_tests/pretrained_models_test.py
```diff
@@ -326,6 +346,14 @@ def test_pretrained_coco_segmentation_subclass_pretrained_shelfnet34_lw(self):
                            metrics_progress_verbose=True)[0].cpu().item()
         self.assertAlmostEqual(res, self.coco_segmentation_subclass_pretrained_mious["shelfnet34_lw"], delta=0.001)
 
+    def test_transfer_learning_coco_segmentation_subclass_pretrained_shelfnet34_lw(self):
+        trainer = SgModel('coco_segmentation_subclass_pretrained_shelfnet34_lw_transfer_learning', model_checkpoints_location='local',
```
lotem-deci, line 350: @shaydeci  - Please notice that "replace_head" is called implicitly and we need a test for a segmentation Model that only tests the basic functionality (replace head -> verify head was replaced...)

shaydeci: This is exactly the purpose of transfer_learning_test- we have a dataset with a different number of classes and so if the head was not replaced properly the training will crash...

PR #91, opened 2022-01-03
Bugfix/dle 937 download dataset once

jonathan-sha:
don't merge - I want to run tests on the tzag


src/super_gradients/training/utils/distributed_training_utils.py
```diff
@@ -1,7 +1,51 @@
+import itertools
+from contextlib import contextmanager
 import torch
 from torch.cuda.amp import autocast
 import torch.nn as nn
-import itertools
+
+
+@contextmanager
+def run_once(timeout=None, group=None):
```
ofrimasad, line 9: nice. but I think we already have a simple decorator that does that....
src.super_gradients.common.environment.env_helpers.multi_process_safe
just add @multi_process_safe on top of the download_from_cloud function

src/super_gradients/training/datasets/dataset_interfaces/dataset_interface.py
```diff
@@ -75,13 +76,18 @@ def __init__(self, dataset_params={}, train_loader=None, val_loader=None, test_l
...
+
+        # IF WE'RE RUNNING IN DDP, WE SHOULD ONLY DO THIS IN RANK 0
+        with run_once() as should_run:
+            if should_run:
+                env_name = AWS_ENV_NAME
+                downloader = DatasetDataInterface(env=env_name)
+                target_dir = self.dataset_params.dataset_dir
+                if not os.path.exists(target_dir):
+                    os.mkdir(target_dir)
+                downloader.load_remote_dataset_file(self.dataset_params.s3_link, target_dir)
```
ofrimasad, line 90: why aren't we checking if the file itself exist? this will download the same file again even if it exist

ofrimasad (commented):
some comment inline

src/super_gradients/training/utils/distributed_training_utils.py
shaydeci: No it is not. One can use multi_process_safe when there's no need for synchronisation, but that's not the case here.

src/super_gradients/training/datasets/dataset_interfaces/dataset_interface.py
shaydeci: This is the zip file before It's unpacked. Checking that it exists does not imply the dataset files actually do...Sure there's a solution for this, but Im not sure this PR is the place for it.

src/super_gradients/training/utils/distributed_training_utils.py
```diff
@@ -1,7 +1,51 @@
...
+            do_work()
+
+    @run_once()
+    def this_works_as_well(should_run):
+        if not should_run:
+            return
+        do_work()
+    ```
+    """
+    # NOT DDP - WE NEED TO RUN THE JOB, BUT DON'T NEED TO SYNC
```
shaydeci, line 35: This is problematic- note that torch.distributed.is_initialized() will return False, when torch.distributed.init_process_group() has not been called yet (even when running p torch distributed launch).

@ofrimasad did suggest working with multi_process_safe (line 74 in src/super_gradients/common/environment/env_helpers.py) which works with the environment config. It'd be good idea to just add an optional "blocking" argument which would be set to false by default, and we can just add the barrier mechanism there. What do you think ?

shaydeci (commented):
See my comment about multi_process_safe, let me know what you think.

shaydeci:
Im closing this PR for now as it has been stale for a while and our code already supports most of the functionality discussed above, will open a tech dept for this issue.

PR #93, opened 2022-01-03
POST_TRAINING callbacks before closing sg_logger

ofrimasad:
This change should allow POST_TRAINING phase callbacks to still use the sg_logger before it is closed


shaydeci (approved):
LGTM

PR #95, opened 2022-01-04
Feature/dle 1012 pypi installation

shaydeci:


ranrubin (dismissed):
I think it will work! let's see :)

PR #96, opened 2022-01-04
Bugfix/dle 1004 regnet recipes update

shaydeci:
PR for the updated Regent's recipe.

I still need to reproduce results, and then add the proper comments.
2 small features that were used in Algo's coded recipe added as well:
- step_lr with frequency
- warmup as used by timm (with initial warmup lr)


src/super_gradients/training/utils/callbacks.py
```diff
@@ -253,18 +253,20 @@ def update_lr(self, optimizer, epoch, batch_idx=None):
 
 class WarmupLRCallback(LRCallbackBase):
     """
-    LR scheduling callback for lr warmup.
-
-    At each update
+    LR scheduling callback for linear step warmup.
+    LR climbs from warmup_initial_lr with even steps to initial lr. When warmup_initial_lr is None- LR climb starts from
+     1/(1+warmup_epochs).
```
sefibk, line 258: it is not 1/.... but will let it slip

sefibk (dismissed):
LGTM

src/super_gradients/training/utils/callbacks.py
shaydeci: fixed

PR #101, opened 2022-01-09
Bugfix/dle 1004 regnet recipes update

shaydeci:


src/super_gradients/recipes/imagenet_regnetY.yaml
```diff
@@ -1,14 +1,26 @@
...
+#  This example trains with batch_size = 256
+#  Training time on a single GeForce RTX 2080 Ti, and top1 accuracies:
+#  11 days for RegnetY200, 70.88
+#  12 days for RegnetY400, 74.74
+#  19 days for RegnetY600, 76.18
+#  20 days for RegnetY800, 77.07
+#  NOTE: Training should probably be lower as resources were shared among the above runs.
+
+#  Logs and tensorboards at:
+# https://deci-pretrained-models.s3.amazonaws.com/RegnetY800/
```
ofrimasad, line 11: changed to http format. was that intentional?

src/super_gradients/recipes/imagenet_regnetY.yaml
```diff
@@ -1,14 +1,26 @@
...
-#  Reach => 76.05 (Final averaged model)
-#  Log and tensorboard at s3://deci-pretrained-models/regnetY800-imagenet-tensorboard/
+#  RegnetY Imagenet classification training:
+#  This example trains with batch_size = 256
+#  Training time on a single GeForce RTX 2080 Ti, and top1 accuracies:
+#  11 days for RegnetY200, 70.88
+#  12 days for RegnetY400, 74.74
+#  19 days for RegnetY600, 76.18
+#  20 days for RegnetY800, 77.07
+#  NOTE: Training should probably be lower as resources were shared among the above runs.
```
ofrimasad, line 8: I think most users would like to parallel train this. You might want to add a comment stating is this is possible or not and how.

src/super_gradients/recipes/imagenet_regnetY.yaml
```diff
@@ -38,7 +50,7 @@ data_loader_num_workers: 8
 model_checkpoints_location: local
 load_checkpoint: False
 
-experiment_name: regnetY800_imagenet
+experiment_name: regnetY800_imagenet_updated
```
ofrimasad, line 53: intentional?

ofrimasad (commented):
my comments inline

src/super_gradients/recipes/imagenet_regnetY.yaml
```diff
@@ -1,14 +1,26 @@
...
+#  Logs and tensorboards at:
+# https://deci-pretrained-models.s3.amazonaws.com/RegnetY800/
+# https://deci-pretrained-models.s3.amazonaws.com/RegnetY600/
+# https://deci-pretrained-models.s3.amazonaws.com/RegnetY400/
+# https://deci-pretrained-models.s3.amazonaws.com/RegnetY200/
 
 # Instructions:
 # Set the PYTHONPATH environment variable: (Replace "YOUR_LOCAL_PATH" with the path to the downloaded repo):
-#   export PYTHONPATH="YOUR_LOCAL_PATH"/super_gradients/
+#   export PYTHONPATH="YOUR_LOCAL_PATH"/super_gradients/:"YOUR_LOCAL_PATH"/super_gradients/src/
```
ofrimasad, line 18: now that you moved recipes into src you dont need `"YOUR_LOCAL_PATH"/super_gradients/` anymore

src/super_gradients/recipes/imagenet_regnetY.yaml
shaydeci: We haven't tested that if we parallelise we reproduce these results...so I think it's a bad idea to add such comment. If a user wants to try and reproduce theses results while using DDP, I think it should be at their own responsibility.

ofrimasad (commented):
just the experiment name. 
regnetY800_imagenet_updated?  whats next "regnetY800_imagenet_updated_new"?

PR #104, opened 2022-01-12
fix: Allow to override augs in CoCoSegmentation

yurkovak:
`CoCoSegmentationDataSet` didn't copy train/val transformations from `dataset_hyper_params` into `kwargs`, so defaults were silently used even when attempting to override.


src/super_gradients/training/datasets/segmentation_datasets/coco_segmentation.py
```diff
@@ -36,11 +36,15 @@ def __init__(self, dataset_classes_inclusion_tuples_list: list = None, *args, **
...
-                                                                 Rescale(long_size=kwargs["img_size"]),
-                                                                 RandomRescale(scales=(0.5, 2.0)),
-                                                                 PadShortToCropSize(crop_size=kwargs['crop_size']),
-                                                                 CropImageAndMask(crop_size=kwargs['crop_size'], mode="random")])
+        default_image_mask_transforms_aug = transform.Compose([RandomFlip(),
+                                                               Rescale(long_size=kwargs["img_size"]),
+                                                               RandomRescale(scales=(0.5, 2.0)),
+                                                               PadShortToCropSize(crop_size=kwargs['crop_size']),
+                                                               CropImageAndMask(crop_size=kwargs['crop_size'], mode="random")])
+        kwargs["image_mask_transforms_aug"] = dataset_hyper_params.get("image_mask_transforms_aug",
```
yurkovak, line 44: Does `get()` work on params from yamls?

yurkovak: `get(...)` works on configs from `hydra.utils.instantiate` , but doesn't work on `HpmStruct`. @shaydeci do we need to support `dataset_hyper_params`  stored in `HpmStruct`?

shaydeci (commented):
inline

src/super_gradients/training/datasets/segmentation_datasets/coco_segmentation.py
```diff
@@ -32,15 +32,20 @@ def __init__(self, dataset_classes_inclusion_tuples_list: list = None, *args, **
...
-                                                                 PadShortToCropSize(crop_size=kwargs['crop_size']),
-                                                                 CropImageAndMask(crop_size=kwargs['crop_size'], mode="random")])
+        default_image_mask_transforms_aug = transform.Compose([RandomFlip(),
+                                                               Rescale(long_size=kwargs["img_size"]),
+                                                               RandomRescale(scales=(0.5, 2.0)),
+                                                               PadShortToCropSize(crop_size=kwargs['crop_size']),
+                                                               CropImageAndMask(crop_size=kwargs['crop_size'], mode="random")])
+        kwargs["image_mask_transforms_aug"] = get_param(dataset_hyper_params, "image_mask_transforms_aug",
+                                                        default_image_mask_transforms_aug)
+        image_mask_transforms = get_param(dataset_hyper_params, 'image_mask_transforms', False)
```
shaydeci, line 46: I know this seems like a small thing...but lets change this to image_mask_transforms = get_param(dataset_hyper_params, 'image_mask_transforms') which will set it by default to None if the key doesn't exist. Just so when somebody debugs this they won't get confused that image_mask_transforms is some flag for something else.

shaydeci (changes requested):
Just a small change, inline. Thanks.

src/super_gradients/training/datasets/segmentation_datasets/coco_segmentation.py
yurkovak: Agree, added in 20280ca

shaydeci (approved):
lgtm

PR #105, opened 2022-01-16
Added factories

ofrimasad:
- Added factories
- connected DatasetsFactory and MetricsFactory to SgModel
- added an example for factory usage to cifar10_resnet.yaml and cifar10_resnet_train_params.yaml

TODO - use factories in all recipes (in next PR)


src/super_gradients/common/factories/datasets_factory.py
```diff
@@ -0,0 +1,27 @@
...
+            "library_dataset": LibraryDatasetInterface,
+            "classification_dataset": ClassificationDatasetInterface,
+            "cifar_10": Cifar10DatasetInterface,
+            "cifar_100": Cifar100DatasetInterface,
+            "imagenet": ImageNetDatasetInterface,
+            "tiny_imagenet": TinyImageNetDatasetInterface,
+            "coco2017_detection": CoCoDetectionDatasetInterface,
+            "coco2014_detection": CoCo2014DetectionDatasetInterface,
+            "coco2017_segmentation": CoCoSegmentationDatasetInterface,
+            "pascal_voc": PascalVOC2012SegmentationDataSetInterface,
```
shaydeci, line 24: We should state the exact dataset (i.e year, voc/aug, segmentation/detection) from pascal. Note that we also have PascalVOCUnifiedDetectionDataSetInterface

src/super_gradients/common/factories/metrics_factory.py
```diff
@@ -0,0 +1,13 @@
...
+from super_gradients.training.metrics import Accuracy, Top5, DetectionMetrics
+
+
+class MetricsFactory(BaseFactory):
+
+    def __init__(self):
+        type_dict = {
+            'Accuracy': Accuracy,
+            'Top5': Top5,
+            'DetectionMetrics': DetectionMetrics,
```
shaydeci, line 11: We have more then these...IoU for example...

src/super_gradients/common/factories/base_factory.py
```diff
@@ -0,0 +1,41 @@
...
+    def get(self, conf: Union[str, dict]):
+        """
+        Get an instantiated object.
+            :param conf: a configuration. either a type name (str) or a dictionary {type_name(str): {parameters...}}
+        """
+        raise NotImplementedError
+
+
+class BaseFactory(AbstractFactory):
+
```
shaydeci, line 18: This is a bit hard to follow, can you please add documentation that's a bit more detailed ?
Also im not sure I understand properly what the behaviour would be for instantiated objects inside a list ( such as train_metrics_list = [Accuracy(), Top5()]).

It is important we don't limit our users to only use our built in objects...

shaydeci (changes requested):
Inline.

Besides that, we need to be very careful not to break our recipes, so please add a unit test that verifies that the same factory calls in them do not crash.

src/super_gradients/common/factories/base_factory.py
ofrimasad: added documentation.
list factory allows the user to generate lists as well (explained in docs)

src/super_gradients/common/factories/datasets_factory.py
ofrimasad: fixed.
PascalVOCDetectionDataSet does not have an interface currently

src/super_gradients/common/factories/metrics_factory.py
ofrimasad: added IoU and PixelAccuracy

ofrimasad (commented):
addressed

src/super_gradients/common/factories/datasets_factory.py
shaydeci: Sure it does, https://github.com/Deci-AI/super-gradients/blob/2aa4045d9236b45623de4d892dd7c767c8398566/src/super_gradients/training/datasets/dataset_interfaces/dataset_interface.py#L697

shaydeci (changes requested):
Thanks. Can we just add a unit test, so we know this won't break our recipes?
Something that takes a  yaml file and checks that it can be instantiated using the factory.

ofrimasad (commented):
added a test

src/super_gradients/common/factories/datasets_factory.py
shaydeci: Also cityscapes

PR #111, opened 2022-01-18
Fixed missing checkpoints_dir_path piping to base logger

NatanBagrov:
Version 1.4.0 has one extra parameter passed to `BaseSGLogger`'s `__init__`: `checkpoints_dir_path`.
It was added in: https://github.com/Deci-AI/super-gradients/commit/26497f3891d668b32ddfba066959c35c61bac0eb#diff-ebb69576351b4be27ebcfe050406dfb79ef52590cc77f213e829d7c30225ce30.

Since it is not apparent in `WandBSGLogger`, which initializes `BaseSGLogger` - an exception is thrown.


shaydeci (approved):
lgtm

PR #117, opened 2022-01-25
Feature/dle 1081 add detection viz callback

shaydeci:


src/super_gradients/training/utils/callbacks.py
```diff
@@ -450,6 +448,34 @@ def __call__(self, context: PhaseContext):
...
+        self.post_prediction_callback = post_prediction_callback
+        self.batch_idx = batch_idx
+        self.classes = classes
+        self.last_img_idx_in_batch = last_img_idx_in_batch
+
+    def __call__(self, context: PhaseContext):
+        if context.epoch % self.freq == 0 and context.batch_idx == self.batch_idx:
+            # SOME CALCULATIONS ARE IN-PLACE IN NMS, SO CLONE THE PREDICTIONS
+            preds = (context.preds[0].clone(), None)
+            preds = self.post_prediction_callback(preds)
```
ofrimasad, line 472: but what if a DetectionPostPredictionCallback is applied on the predictions before this callback?
maybe add a warning in the documentation...

ofrimasad (commented):
just one minor issue. this callback should not be used together with the regular DetectionPostPredictionCallback... since in include such a callback inside it

src/super_gradients/training/utils/callbacks.py
shaydeci: This cannot be the case as we take the predictions from context.preds. 
In SgModel code we update the context with raw predictions always.

ofrimasad: great

PR #120, opened 2022-01-27
- added a new factory for transforms

ofrimasad:
- cleaned cityscapes_stdc_seg50 training from redundant _target_ calls as an example for the new factory
- fixed a bug in the factory decorator
- added stdc_loss to the losses list

### DIFF
--- a/src/super_gradients/recipes/cityscapes_stdc_seg50.yaml
+++ b/src/super_gradients/recipes/cityscapes_stdc_seg50.yaml
@@ -52,44 +52,42 @@ dataset_params:
   _convert_: all
   batch_size: 8
   val_batch_size: 8
-  color_jitter: 0.5
-  random_scales:
-    - 0.125
-    - 1.5
   crop_size:
     - 1024
     - 512
   eval_scale: 0.5
   image_mask_transforms_aug:
-    _target_: torchvision.transforms.Compose
-    transforms:
-      - _target_: super_gradients.training.utils.segmentation_utils.ColorJitterSeg
-        brightness: ${dataset_params.color_jitter}
-        contrast: ${dataset_params.color_jitter}
-        saturation: ${dataset_params.color_jitter}
+    Compose:
+      transforms:
+        - ColorJitterSeg:
+            brightness: 0.5
+            contrast: 0.5
+            saturation: 0.5
 
-      - _target_: super_gradients.training.utils.segmentation_utils.RandomFlip
+        - RandomFlipSeg
 
-      - _target_: super_gradients.training.utils.segmentation_utils.RandomRescale
-        scales: ${dataset_params.random_scales}
+        - RandomRescaleSeg:
+            scales:
+              - 0.125
+              - 1.5
 
-      - _target_: super_gradients.training.utils.segmentation_utils.PadShortToCropSize
-        crop_size: ${dataset_params.crop_size}
-        fill_mask: ${dataset_params.cityscapes_ignored_label}
+        - PadShortToCropSizeSeg:
+            crop_size: ${dataset_params.crop_size}
+            fill_mask: ${dataset_params.cityscapes_ignored_label}
 
-      - _target_: super_gradients.training.utils.segmentation_utils.CropImageAndMask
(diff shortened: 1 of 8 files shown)


src/super_gradients/training/utils/segmentation_utils.py
```diff
@@ -264,7 +264,7 @@ def __init__(self, crop_size: Union[float, Tuple, List], fill_mask: int = 0, fil
         :param fill_image: grey value to fill image padded background.
         """
         # CHECK IF CROP SIZE IS A ITERABLE OR SCALAR
-        self.crop_size = crop_size
+        self.crop_size = tuple(crop_size)
```
lkdci, line 267: Why do you need this? if crop size is a single integer as legacy used it will failed.
Transforming crop_size to tuple is already handled in self.check_valid_arguments().
Is it about PIL require only tuple type? If that the case check_valid_arguments() is the place to handle that, after crop_size is validated as iterable.

lkdci, line 267: Why do you need this? if crop size is a single integer as legacy used it will failed.
Transforming crop_size to tuple is already handled in self.check_valid_arguments().
Is it about PIL require only tuple type? If that the case check_valid_arguments() is the place to handle that, after crop_size is validated as iterable.

src/super_gradients/training/datasets/segmentation_datasets/segmentation_dataset.py
```diff
@@ -6,12 +6,18 @@
...
+from super_gradients.common.decorators.factory_decorator import resolve_param
+from super_gradients.common.factories.transforms_factory import TransformsFactory
 from super_gradients.training.datasets.sg_dataset import DirectoryDataSet, ListDataset
 from super_gradients.training.utils.segmentation_utils import RandomFlip, Rescale, RandomRotate, PadShortToCropSize, \
     CropImageAndMask, RandomGaussianBlur, RandomRescale
 
 
 class SegmentationDataSet(DirectoryDataSet, ListDataset):
+
+    @resolve_param('image_mask_transforms', factory=TransformsFactory())
```
lkdci, line 19: @shaydeci  do we really need both `image_mask_transforms` and `image_mask_transforms`, the actual one that is used is determined by `augment: bool`, the other is redundant.
IMO there should be only `image_mask_transforms`, `augment` is redundant and only complicating this dataset class.

src/super_gradients/recipes/cityscapes_stdc_seg50.yaml
```diff
@@ -52,44 +52,42 @@ dataset_params:
...
     - 512
   eval_scale: 0.5
   image_mask_transforms_aug:
-    _target_: torchvision.transforms.Compose
-    transforms:
-      - _target_: super_gradients.training.utils.segmentation_utils.ColorJitterSeg
-        brightness: ${dataset_params.color_jitter}
-        contrast: ${dataset_params.color_jitter}
-        saturation: ${dataset_params.color_jitter}
+    Compose:
```
lkdci, line 60: Why do we need Compose -> transforms? Can it be handled in the factory class, and not in YAML?

lkdci (commented):
Inline

src/super_gradients/training/utils/segmentation_utils.py
ofrimasad: removed. was a mistake

src/super_gradients/recipes/cityscapes_stdc_seg50.yaml
ofrimasad: to keep the general logic of all factories:

CalssName:
    parameter_name: parameter_value

src/super_gradients/training/datasets/segmentation_datasets/segmentation_dataset.py
shaydeci: I personally don't mind it. I think it can come in handy if you want to cancel out augmentations at some point (for example fro debugging). Both ways are okay on my end.

shaydeci (approved):
lgtm

PR #122, opened 2022-01-31
fix pad short to crop size

ofrimasad:
fix augmentation when passing a list instead of a tuple (which happens when working with yamls)


shaydeci (approved):
LGTM

NatanBagrov:
LGTM

PR #123, opened 2022-02-02
Feature/dle 1127 dali integration

shaydeci:
Optional DALI support for imagenet. 
For 4 X A5000 DDP training on our Resnet recipe, we  get: 
- 4 % shorter epoch time for Resnet50
- 15 % shorter epoch time for Resnet18


src/super_gradients/recipes/dataset_params/imagenet_dataset_params.yaml
```diff
@@ -12,3 +12,5 @@ imagenet_pca_aug: 0.0 # imagenet pca augmentation (applied only to trainset)
 train_interpolation: default # interpolation mode
 rand_augment_config_string: # randaugment config string (see super_gradients/training/datasets/auto_augment.py)
 random_erase_prob: 0.0 # random erase probability (applied only to trainset)
+use_dali: False # whether to use dali pipeline
```
lotem-deci, line 15: @shaydeci  - Let's be a little more explicit for this param...
 Consider - "nvidia_dali_data_loading" or something else, so it will be clear to our users...

src/super_gradients/training/datasets/dali_datasets/dali_dataloaders.py
```diff
@@ -0,0 +1,18 @@
+import collections
+
+
+class DaliClassificationDataLoader(collections.Iterator):
```
lotem-deci, line 4: @shaydeci - I think a more suitable location for this class is inside dataset_interface, since this is not a dataset for DL training... Let's move this please

src/super_gradients/training/datasets/dali_datasets/dali_pipelines.py
```diff
@@ -0,0 +1,77 @@
+try:
+    from nvidia.dali.pipeline import pipeline_def
+    import nvidia.dali.types as types
+    import nvidia.dali.fn as fn
+except ImportError:
+    raise ImportError("Please install DALI from https://www.github.com/NVIDIA/DALI.")
+
+
+@pipeline_def
+def imagenet_dali_pipeline(data_dir: str, crop: int, resize_size: int, shard_id: int, num_shards: int, dali_cpu=False, is_training=True):
```
lotem-deci, line 10: @shaydeci  - I think that this design pattern by nvidia could be easily modified to a "DaliDataSetInterface"... 

I would expect that Dali integration could be implemented generically so it can wrap every dataset interface we already have in deci, and this will be much cleaner, leaner and less error prone.

Also - It will be more future-proof, and will be much easier to maintain.

lotem-deci (commented):
LGTM

PR #124, opened 2022-02-03
Progressed the latest README.md to docs/

avideci:


avideci:
I also removed docs/.doctrees (used by sphinx) because it contained a pickled environment, which is a bad practice.
Added it to .gitignore

shaydeci (approved):
lgtm

PR #126, opened 2022-02-03
fix add_image img_tensor arg

lkdci:
Fix argument name from `image` to `img_tensor`, as can be see in torch [doc](https://pytorch.org/docs/stable/_modules/torch/utils/tensorboard/writer.html#SummaryWriter.add_image)

Error before fix:
```
TypeError: add_image() got an unexpected keyword argument 'image'
```


shaydeci (approved):
LGTM

PR #128, opened 2022-02-07
Added DropPath to ResNet, retrained, achieved 79.47 top-1

NatanBagrov:


src/super_gradients/training/models/classification_models/resnet.py
```diff
@@ -228,6 +240,7 @@ def replace_head(self, new_num_classes=None, new_head=None):
 
 def ResNet18(arch_params, num_classes=None, backbone_mode=None):
     return ResNet(BasicBlock, [2, 2, 2, 2], num_classes=num_classes or arch_params.num_classes,
+                  droppath_prob=get_param(arch_params, "droppath_prob", 0),
```
NatanBagrov, line 243: I think for future proof is better to pass `arch_params` dict instead of "flattening" it here.
On one hand, it might be "better" not to mix names and dicts in `ResNet` class;
On the other hand, if we're interested in adding another arch feature (such as AntiAliasing), we're doomed to to all the piping instead of just adding it to the `arch_params` and extracting it inside `ResNet`.
Up to you.

shaydeci (approved):
LGTM

PR #130, opened 2022-02-09
Updated & regenrated docs from master

avideci:
How to generate docs:

```python
$  ./scripts/generate_docs.sh

# Docs are ready, Let's check them out:
$  cd docs && python -m http.server 8080
```


shaydeci (approved):
LGTM

PR #133, opened 2022-02-14
New Transfer Learning Object Detection Notebook Version

lotem-deci:
* Changed the flow of the notebook to put more emphasis on the methods and the value proposition
* Added inline comments for the code blocks 
* Cleaned all outputs of the cells
* Added more references to allow the user to continue learning and diving deeper into concepts we did not have time to drill down into


yurkovak:
1. A comment to *Generating new anchors*, to the sentence
> We are using a different DataSetInterface for anchors generation even though we are using the same images and labels since the parameters for the Anchors generation process are different from the ones for training/validation

You actually do expect to fit anchors for the data as it's going to be in training/validation, so it sounds slightly fishy and raises a question. If the parameters that differ are batch size and stuff like that, best to say it explicitly. And generally to explain which ones differ.

2. `dataset_interface` has the same name for both uses of the interface. It doesn't break anything, because anchors are calculated after `connect_dataset_interface`, but I would still rename the anchors one to prevent [this situation.](https://harpomaxx.github.io/post/deploying-ml-model/images/notebooksvsservices.jpeg) 

3. This comment of mine
> Both bullets below are either redunant or not very clear. 

wasn't removed, huge apologies again for leaving comments like this in italic!

4. Following the discussion of yaml, I have an idea how to make the so called story better: do all the stuff related to defining params, datasets, anchors, and so on in the cells one by one (like it it already) *but* move all the SG boilerplate code to one cell close to the end, e.g.

```python
# create a trainer, yadda yadda 
trainer = SgModel(experiment_name="yolov5n_pascal_finetune", ckpt_root_dir="/home/data/notebook_ckpts/")
# connect our data, etc etc
trainer.connect_dataset_interface(...)
# build a pre-trained model with new anchors
trainer.net.to('cpu')
trainer.build_model(architecture="yolo_v5n", arch_params={"pretrained_weights": "coco", "anchors": anchors})
# and so on and so on
``` 
This way a user would:
a. learn clearly which actions transfer learning requires
b. recall again the code of training in SG and what's new in it for this task

PR #136, opened 2022-02-16
Feature/sg 36 yolox model

yurkovak:
Adding YoloX architectures, Nano: Tiny, S, M, L, X.

Since most of the code overlaps with V5, I'm reusing it and tweaking with the yolo_type arch_param. However, with this flag having all the models (both V5 and X) in one file called yolov5.py becomes ridiculous. That's why I'd like to rename the current yolov5.py into yolo_base.py and move YoLoV5(from N to X) into a small yolov5.py that will have only definition of these models. 
On the other side, it won't look that good too, because we also have yolov3.py and yolo_base won't be a base for it. 

Let me know what you think is best


yurkovak (commented):
Question inline

shaydeci:
> current yolov5.py into yolo_base.py and move YoLoV5(

How about renaming it to yolov5_base.py? then moving N-L to yolov5.py and x as you did ?

shaydeci (changes requested):
See inline answers and comment

PR #137, opened 2022-02-16
feature: Add YoloX configs

yurkovak:
Configs used for training of YoloX. Should be merged the last of all the PRs. Will add cfgs for the models later too.


src/super_gradients/recipes/coco2017_yolox.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+sg_model:
+  _target_: super_gradients.SgModel
+  experiment_name: ${experiment_name}
+  model_checkpoints_location: ${model_checkpoints_location}
+
+training_hyperparams:
+  criterion_params:
+    strides: [8, 16, 32]  # output strides of all yolo outputs
+    im_size: 640
```
yurkovak, line 36: To be fixed, should come from `dataset_params.train_image_size`

yurkovak: However! If a user runs with overriding of this dataset field, it won't impact the criterion_params, will it? Requires a better solution

shaydeci: Sure it will, as dataset_params is compiled before this config.
Lets change it to ${dataset_params.train_image_size}.

shaydeci (changes requested):
Minor inline comment

PR #141, opened 2022-03-06
Feature/sg 34 regseg transfer learning

shaydeci:
Added example code for Regseg transfer learning, which will later be used for an example notebook.
This PR holds several necessary features for it to run:

- BCE with Dice loss.
- Binary IOU metric object (I.e IOU only for target class).
- Binary segmentation visualisation callback.
- Supervisely dataset interface.
- Different lr assignment for head and backbone for RegSeg.


src/super_gradients/training/metrics/segmentation_metrics.py
```diff
@@ -111,3 +111,16 @@ def update(self, preds, target: torch.Tensor):
             preds = preds[0]
         _, preds = torch.max(preds, 1)
         super().update(preds=preds, target=target)
+
+
+class BinaryIOU(torchmetrics.IoU):
```
lotem-deci, line 116: @shaydeci  - Did we make a decision on what measures we are going to take to handle the SegmentationMetrics issue we've had with mIOU (caused by ignore_index...)?

src/super_gradients/training/utils/callbacks.py
```diff
@@ -477,6 +478,30 @@ def __call__(self, context: PhaseContext):
             context.sg_logger.add_images(tag=tag, images=batch_imgs[:self.last_img_idx_in_batch], global_step=context.epoch, data_format='NHWC')
 
 
+class BinarySegmentationVisualizationCallback(PhaseCallback):
```
lotem-deci, line 481: @shaydeci  - Why did we implement a specific VisualizationCallback for BinaryIOU? Can't we have a generic implementation for SegmentationIOU and have BinaryIOU as a specific edge case of it?

lotem-deci: If it's complicated and will slow us down let's open an issue in the backlog for it - But I am not sure it cannot be implemented easily for more than 2 classes

src/super_gradients/training/utils/segmentation_utils.py
```diff
@@ -350,3 +373,100 @@ def to_one_hot(target: torch.Tensor, num_classes: int, ignore_index: int = None)
         one_hot = torch.cat([one_hot[:, :ignore_index], one_hot[:, ignore_index + 1:]], dim=1)
 
     return one_hot
+
+
+def undo_image_preprocessing(im_tensor: torch.Tensor) -> np.ndarray:
```
lotem-deci, line 378: @shaydeci  - Can we give this function a better name?

src/super_gradients/training/utils/segmentation_utils.py
```diff
@@ -21,6 +26,19 @@ def __repr__(self):
         return self.__class__.__name__ + str(self.__dict__).replace('{', '(').replace('}', ')')
 
 
+class ResizeSeg(SegmentationTransform):
```
lotem-deci, line 29: @shaydeci  - We need to create a proper place in the modules hierarchy for all of the Augmentations instead of putting everything inside utils...

lotem-deci:
Nice work.
We have some required fixes inline...

src/super_gradients/training/metrics/segmentation_metrics.py
shaydeci: Yes as we discussed- I checked and IOU with ignored index behaviour is as right.
For binary segmentation though it is more common to just take the target IOU, but I'm also returning mIOU and BG IOU for logging.

src/super_gradients/training/utils/callbacks.py
shaydeci: Regarding the first question- IOU metric returns the mean across all classe including background, where for binary segmentation though it is more common to just take the target IOU, but I'm also returning mIOU and BG IOU for logging.

Regarding the second question- you are probably right, it just requires some thought about how to display the images (see the way I present TPs, FN, and FPs). Not sure wed want to do that for all classes....maybe we do, but anyways, Il add it to the backlog since I think it will slow us down.

lotem-deci (approved):
LGTM

PR #145, opened 2022-03-13
Feature/sg 71 kd module

shaydeci:
This PR holds:
- KDModule implementation
- SgModel's new function build_kd_model which instantiates teacher and student models, then sets its net attribute and takes care of checkpoint loading (both for resuming kd training and loading checkpoints to the teacher prior to training).
- Unit tests for the different scenarios when building kd models. 
- Refactoring of SgModel's build_model method and corresponding fixes in method calls.


src/super_gradients/training/models/kd_models/kd_module.py
```diff
@@ -0,0 +1,62 @@
...
+class KDModule(SgModule):
+    """
+    KDModule
+
+    class implementing Knowledge Distillation logic as an SgModule
+
+    attributes:
+        student: torch.nn.Module - the student model
+        teacher: torch.nn.Module- the teacher model
+        freeze_teacher_eval_mode: bool- whether to freeze self.teacher at eval mode regardless of self.train(mode)
```
shairoz-deci, line 21: why would someone want to train the teacher? when will this be set to True?

src/super_gradients/training/models/kd_models/kd_module.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    def __init__(self, student: torch.nn.Module, teacher: torch.nn.Module,
+                 freeze_teacher_eval_mode: bool = False):
+        super(KDModule, self).__init__()
+        self.student = student
+        self.teacher = teacher
+        self.freeze_teacher_eval_mode = freeze_teacher_eval_mode
+        self._freeze_teacher()
+
+    def _freeze_teacher(self):
```
shairoz-deci, line 31: @shaydeci can you explain this. The teacher's params are set with require_grad=False. and if freeze_teacher_eval_mode then will be switched to train mode but still with require_grad=False?

src/super_gradients/training/models/kd_models/kd_module.py
```diff
@@ -0,0 +1,62 @@
+from super_gradients.training.models.sg_module import SgModule
+import torch
+from super_gradients.training.utils.utils import HpmStruct
+
+
+class KDOutput:
+    def __init__(self, student_output: torch.Tensor = None, teacher_output: torch.Tensor = None):
```
shairoz-deci, line 7: is it wise to have a default value for student_output? shouldn't it be obligatory?

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -320,6 +313,87 @@ def build_model(self,  # noqa: C901 - too complex
...
+        kd_model_arch_params['num_classes'] = student_arch_params['num_classes']
+        teacher_arch_params['num_classes'] = student_arch_params['num_classes']
+
+        student_arch_params = core_utils.HpmStruct(**student_arch_params)
+        teacher_arch_params = core_utils.HpmStruct(**teacher_arch_params)
+
+        student_net, _ = sg_model_utils.instantiate_net(student_architecture, student_arch_params)
+        teacher_net, _ = sg_model_utils.instantiate_net(teacher_architecture, teacher_arch_params)
+
+        if teacher_checkpoint_path is not None:
```
shairoz-deci, line 362: why do you allow teacher_checkpoint_path to be None? I'd think that we should assert if None because then its not distillation

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -320,6 +313,87 @@ def build_model(self,  # noqa: C901 - too complex
...
+                                 "number of classes.")
+            if teacher_checkpoint_path is not None:
+                if teacher_pretrained_weights:
+                    logger.warning(
+                        teacher_checkpoint_path + " checkpoint is "
+                                                  "overriding " + teacher_pretrained_weights + "for teacher model")
+
+            load_pretrained_weights(teacher_net, teacher_architecture, teacher_pretrained_weights)
+
+        architecture = KDModule(student=student_net,
```
shairoz-deci, line 386: At this point if neither pretrained_weights nor teacher_checkpoint_path were provided the teacher will start from scratch. This should be asserted IMO

src/super_gradients/training/models/kd_models/kd_module.py
shaydeci: freeze_teacher_eval_mode control's whether self.teacher should also transfer to train=True, but that doesn't change the fact the teacher won't be trained (it just controls dropout and bn statistics).
Essentially what this flag means is whether the teacher should have dropout and running bn stats on when training the student or not.

src/super_gradients/training/models/kd_models/kd_module.py
shaydeci: See above comment regarding freeze_teacher_eval_mode flag

src/super_gradients/training/models/kd_models/kd_module.py
shaydeci: Will fix it.

src/super_gradients/training/sg_model/sg_model.py
shaydeci: Good point, will fix it.

src/super_gradients/training/models/kd_models/kd_module.py
shairoz-deci: This is not clear from the name or documentation. if the difference is between no_grad and eval() than perhaps call the flag "run_teacher_on_eval" or something of that sort and explain in the doc. "Freeze" IMO gives a different connotation. 
and if the teacher will run either with no_grad or eval why do you set all of its params with require_grad=False?

shaydeci: When setting the teacher's parameters with requires_grad=False we reduce the network's number of parameters and memory consumption ( https://stackoverflow.com/questions/63785319/pytorch-torch-no-grad-versus-requires-grad-false)
It is true that we don't need to wrap then the teacher's forward with no_grad(), that ill fix now.

src/super_gradients/training/sg_model/sg_model.py
shaydeci: fixed

src/super_gradients/training/models/kd_models/kd_module.py
shaydeci: fixed

src/super_gradients/training/models/kd_models/kd_module.py
shaydeci: fixed

src/super_gradients/training/models/kd_models/kd_module.py
shaydeci: removed with torch no_grad()  in forward pass

shairoz-deci (commented):
LGTM
@lotem-deci do you want to go over as well?

src/super_gradients/training/models/kd_models/kd_module.py
```diff
@@ -0,0 +1,63 @@
...
+from super_gradients.training.utils.utils import HpmStruct
+
+
+class KDOutput:
+    def __init__(self, student_output: torch.Tensor, teacher_output: torch.Tensor):
+        self.student_output = student_output
+        self.teacher_output = teacher_output
+
+
+class KDModule(SgModule):
```
lotem-deci, line 12: @shaydeci  - Please notice you haven't implemented replace_head method

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -320,6 +313,91 @@ def build_model(self,  # noqa: C901 - too complex
                 self.arch_params.num_classes = num_classes_new_head
                 self.net.to(self.device)
 
+    def build_kd_model(self,
```
lotem-deci, line 316: @shaydeci  - Something does not feel right...
All of the checkpoints mechanism for KD needs to be polished and adapted accordingly...

BTW - I suggest to have all of the code that handles KD in the same place, and not spanned across all of our code base... namely - SG_Model SHOULD NOT BE CHANGED in order to support an extended use case... We should inherit from sg_model and implement kd_model that overrides the build_model method according to the required behaviour...

This is just one use case, but this is relevant for all of the required changes.. like save_checkpoint, changes to the metrics or anything else that is *specific* to a KDModel and not relevant for an SGModel...

We need to be more aware of this design since we will eventually want to take apart the SGModel monolith into smaller modules and we do not want to feed the monster

lotem-deci:
This looks good - We need to make some more comprehensive design changes to the surrounding code (SGModel) in order for us to merge this new feature...

Nice Work!

shaydeci:
Im closing this down as we move our kd related pr's to https://github.com/Deci-AI/super-gradients/tree/feature/SG-000_KD_master_branch

PR #149, opened 2022-03-14
Feature/sg 75 refactor train from recipe

shaydeci:
The purpose of this PR is to reduce time to start training from recipe for users, and to do so from the command line after just installing SG, with a relatively simple command.
For example:
python -m super_gradients.train_from_recipe  --config-name=imagenet_regnetY architecture=regnetY800 ckpt_root_dir=<my_checkpoints_dir>

This means that since users won't necessarily clone the repo, I added the option to pass the ckpt_root dir when running recipes as well.


shairoz-deci (commented):
LGTM

PR #151, opened 2022-03-17
Adding a notebook with upload to the Deci Lab

daniel-deci:


src/super_gradients/examples/deci_lab_export_example/deci_lab_export_example.py
```diff
@@ -45,17 +48,28 @@
...
+                "valid_metrics_list": [Accuracy(), Top5()],
+                "loss_logging_items_names": ["Loss"],
+                "metric_to_watch": "Accuracy",
+                "greater_metric_to_watch_is_better": True,
+                "phase_callbacks": phase_callbacks}
 
 # RUN TRAINING. ONCE ALL EPOCHS ARE DONE THE OPTIMIZED MODEL FILE WILL BE LOCATED IN THE EXPERIMENT'S
 # CHECKPOINT DIRECTORY
-model.train(train_params)
+if __name__ == '__main__':
```
daniel-deci, line 74: I added this as a commodity for make it easy to debug from PyCharm

daniel-deci:
@shaydeci @lotem-deci 
This can be reviewed but can't be merged until we publish the [deci-lab-client](https://github.com/Deci-AI/deci-lab-client/pull/21)

src/super_gradients/training/utils/callbacks.py
```diff
@@ -195,30 +197,76 @@ def __init__(self, email, model_meta_data, optimization_request_form, password=N
...
+            This function will attempt to upload the trained model and schedule an optimization for it.
+            Args:
+                context (PhaseContext): Training phase context
+            Returns:
+                bool: whether or not the optimized model has been benchmarked
+            """
         try:
             model = copy.deepcopy(context.net)
-            model_state_dict_path = os.path.join(context.ckpt_dir, self.ckpt_name)['net']
-            model.load_state_dict(model_state_dict_path)
```
shaydeci, line 202: This line shouldn't be removed.  There's a a reason we load the weights from model_state_dict_path-
The callback is fired at the end of training and the weights at that point might not be optimal (or just not the ones we meant when specifying them in ckpt_name.

src/super_gradients/examples/deci_lab_export_example/deci_lab_export_example.py
```diff
@@ -21,19 +23,20 @@
 model.build_model(net)
```
shaydeci, line 23: We need to make sure conversion works for our SG supported architectures (for example calling model.build_model("yolo_v5n")).
We need to at least do this for some architecture families in each task. You can take the code from our Quickstart notebooks for the different tasks and modify the number of epochs to something small, and add the deci lab upload callback to the phase callbacks as done here.

src/super_gradients/training/utils/callbacks.py
```diff
@@ -195,30 +197,79 @@ def __init__(self, email, model_meta_data, optimization_request_form, password=N
...
+        logger.info(f"We couldn't finish your model optimization. Visit https://console.deci.ai for details")
+
+    def upload_model(self, model):
+        """
+            This function will upload the trained model to the Deci Lab
+
+            Args:
+                model: The resulting model from the training process
+        """
+        self.platform_client.add_model(
```
shaydeci, line 211: For some architectures we must call prep_model_for_conversion().
Please check if we need to do so before passing the model through local_loaded_model arg by:
model = model.module.cpu()
if hasattr(model, "prep_model_for_conversion"):
    model.prep_model_for_conversion(input_size=self.model_met_data.input_dimensions)

(not sure this is how to get the input dimensions from ModelMetaData but you get the principle...)

shaydeci (changes requested):
See inline comments.
We need to check this works for at least most of the architectures, and definitely the ones the require prep_model_for_conversion prior to the call.
Make sure when these architectures are built, that the call actually does some manipulation (by passing the right arguments through arch params when building them, for instance take a look at repvgg prep_model_for_conversion)

shaydeci (approved):
LGTM

lotem-deci (approved):
LGTM

PR #154, opened 2022-03-22
Add criterion to context before train starts

lkdci:
Added due to that context.criterion is None when accessing context with phase `TRAIN_EPOCH_START`


shaydeci (approved):
lgtm

PR #159, opened 2022-03-27
Feature/sg 82 seperate ckpt params from arch params

shaydeci:
- This PR holds the separation of checkpoint related parameters from arch_params in sg_model.build_model().

*pretrained weights arg moved to checkpoint params as well


src/super_gradients/training/kd_model/kd_model.py
```diff
@@ -20,7 +20,7 @@ class KDModel(SgModel):
 
     def build_model(self,  # noqa: C901 - too complex
                     architecture: Union[str, nn.Module] = 'kd_module',
-                    arch_params={},
+                    arch_params={}, checkpoint_params={},
```
lotem-deci, line 23: @shaydeci  - You've worked really hard to move the checkpoint loading mechanism of sg_model outside of build_model, yet in KDModel you break this and add this unneeded coupling again.

Can we implement the same logic as we did in SGModel or is this something I'm missing?

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -222,18 +223,21 @@ def connect_dataset_interface(self, dataset_interface: DatasetInterface, data_lo
     # FIXME - we need to resolve flake8's 'function is too complex' for this function
     def build_model(self,  # noqa: C901 - too complex
                     architecture: Union[str, nn.Module],
-                    arch_params={}, *args, **kwargs):
+                    arch_params={}, checkpoint_params={}, *args, **kwargs):
```
lotem-deci, line 226: @shaydeci  - Let's discuss tomorrow why are we calling load_checkpoint from within build_model... If we have successfully decoupled both of these methods, we should probably be able to call these two methods independently from SGModel...

Let's discuss this tomorrow

src/super_gradients/training/kd_model/kd_model.py
shaydeci: The logic isn't changed.
Load checkpoint was always called inside build_model, decoupling it (in SG or KD) is not something that was discussed when we got Algo's comments regarding KD implementation, but rather the fact that arch_params holds extra information which is not architecture related....

lotem-deci: This is correct... The problem is with the end result of we have right now on our plate:
1. Inside SGModel -> De-Coupled logic for "build_model"(A) and "load_checkpoint"(B) but B is invoked from A
2. Inside KDModel -> logic for "load_checkpoint"(B) is manipulated because it is called from within A... So it feels like a good place to "add just a little more business logic to not touch load_checkpoint in the inheriting class" when in fact we want the exact opposite... We want the inheriting class to make the proper adjustments inside "load_checkpoint" and than he can (if this is what's needed) to call super.load_checkpoint after running the new BL...

src/super_gradients/training/kd_model/kd_model.py
```diff
@@ -18,101 +19,194 @@ class KDModel(SgModel):
...
 
-        # DERIVE NUMBER OF CLASSES FROM DATASET INTERFACE IF NOT SPECIFIED
-        if 'num_classes' not in student_arch_params.keys():
-            if self.dataset_interface is None:
-                raise Exception('Error', 'Number of classes not defined in students arch params and dataset is not '
-                                         'defined')
-            else:
-                student_arch_params['num_classes'] = len(self.classes)
+        if get_param(checkpoint_params, 'pretrained_weights') is not None:
+            raise ValueError("pretrained_weights is ambiguous for KD models.")
```
lotem-deci, line 91: @shaydeci  - Great work on all of the exception handling!

I want to point out that all over this code you are using vanilla exceptions... We should consider using custom exceptions since their uniqueness allows the stack trace to point to the location in the codebase easily.

For example:
Creating a base class for exception (KDModelException that inherits from Exception) will allow the user to search for it in the code base and find its source file... 
Then we can inherit from this base exception class and raise more specific exceptions for eachof the relevant use cases...

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -242,37 +250,34 @@ def build_model(self,  # noqa: C901 - too complex
                 arch_params['num_classes'] = len(self.classes)
 
         self.arch_params = core_utils.HpmStruct(**arch_params)
+        self.checkpoint_params = core_utils.HpmStruct(**checkpoint_params)
 
-        self.net, self.architecture_cls = sg_model_utils.instantiate_net(architecture, self.arch_params)
+        # FIXME: MAKE CLASS METHOD OUT OF INSTANTIATE NET AND PASS STUDENT AND TEEACHER THROUGH KWARGS
```
lotem-deci, line 255: @shaydeci  - Do not forget to clean #FIXMEs and #TODOs

src/super_gradients/training/kd_model/kd_model.py
```diff
@@ -18,101 +19,194 @@ class KDModel(SgModel):
...
+        student_architecture = get_param(kwargs, "student_architecture")
+        teacher_architecture = get_param(kwargs, "teacher_architecture")
+        student_arch_params = get_param(kwargs, "student_arch_params")
+        teacher_arch_params = get_param(kwargs, "teacher_arch_params")
+        student_arch_params = core_utils.HpmStruct(**student_arch_params)
+        teacher_arch_params = core_utils.HpmStruct(**teacher_arch_params)
+        student_pretrained_weights = get_param(checkpoint_params, 'student_pretrained_weights')
+        teacher_pretrained_weights = get_param(checkpoint_params, 'teacher_pretrained_weights')
+
+        student, _ = SgModel.instantiate_net(student_architecture, student_arch_params,
```
lotem-deci, line 177: @shaydeci  - I think that the IDE's recommendation to define instantiate_net as a staticmethod is wrong since it breaks the inheritance...

If we remove the "@staticmethod" from SGModel.instantiate_net than we can simply replace 
SgModel.instantiate_net(student_architecture, student_arch_params,
                                             {"pretrained_weights": student_pretrained_weights}) 
with 
super().instantiate_net(student_architecture, student_arch_params,
                                             {"pretrained_weights": student_pretrained_weights})

Which is the behavior we are after...


WDYT?

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -1653,3 +1669,41 @@ def evaluate(self, data_loader: torch.utils.data.DataLoader, metrics: MetricColl
...
+        else:
+            net = architecture
+
+        if pretrained_weights:
+            load_pretrained_weights(net, architecture, pretrained_weights)
+            if num_classes_new_head != arch_params.num_classes:
+                net.replace_head(new_num_classes=num_classes_new_head)
+                arch_params.num_classes = num_classes_new_head
+
+        return net, architecture_cls
```
lotem-deci, line 1709: @shaydeci  - Why are we returning both net and architecture_cls?

Running the parsing of the str:
architecture_cls = ARCHITECTURES[architecture] is not relevant any more...

The only usage of self.architecture_cls is in re_build_model... (line 1234) - but we can (and probably should) replace this call with instantiate_net...

shaydeci: As discussed- we need to return them as the nets are used for initialising the teacher and student (KDModel.instantiate_net())

lotem-deci: @shaydeci  - Why do we need to return architecture_cls? We need to change re_build_model to use instantiate_net... And we can remove the self.architecture_cls model class attribute since it is not needed anymore...

shaydeci: done

lotem-deci (approved):
LVGTM

PR #161, opened 2022-03-28
fix stdc context conversion

lkdci:
Fixed when replacing to static size context embedding, the weights are transferred to the new module.


shaydeci (approved):
LGTM

PR #173, opened 2022-04-17
Exponential LR scheduler

lkdci:
Exponential LR scheduler function. Good option for research purposes, doesn't depend on max_epochs so the training can be resumed. Only `lr_decay_factor` as hyperparameter.


shaydeci (approved):
LGTM

PR #174, opened 2022-04-17
Add Dice and BinaryDice metrics

lkdci:
Added Dice and binary Dice metrics, by implementing `_dice_from_confmat` similar format as `_jaccard_from_confmat` torchmetrics implementation.
* Added new metrics to MetricsFactory
* Created new `_preprocess_segmentation_inputs` to prevent duplicate code in all segmentation metrics.
* IoU now inherits from `torchmetrics.JaccardIndex` instead of `torchmetrics.IoU` which is no longer  supported in torchmetrics 0.8.0.


src/super_gradients/training/metrics/segmentation_metrics.py
```diff
@@ -101,26 +167,73 @@ def compute(self):
         return pix_acc
 
 
-class IoU(torchmetrics.IoU):
-    def __init__(self, num_classes, dist_sync_on_step=True, ignore_index=None):
-        super().__init__(num_classes=num_classes, dist_sync_on_step=dist_sync_on_step, ignore_index=ignore_index)
+class IoU(torchmetrics.JaccardIndex):
```
shaydeci, line 170: Notice we are currently using 0.7.3, did you check inheriting from torchmetrics.JaccardIndex works with that version ?

src/super_gradients/training/metrics/segmentation_metrics.py
```diff
@@ -77,6 +124,26 @@ def intersection_and_union(im_pred, im_lab, num_class):
     return area_inter, area_union
 
 
+def _preprocess_segmentation_inputs(preds, target: torch.Tensor,
```
shaydeci, line 127: We are getting more and more of these for each new scenario...
My suggestion- lets make a callable class (named MetricsArgsPrepFn or something similar) that we can just pass to our Metrics objects, and would apply some transformations to input and targets before applying the metrics update.
In this case wed inherit from it and construct a class named PreprocessSegmentationMetricsArgsFn that applies the below code accordingly.

src/super_gradients/training/metrics/segmentation_metrics.py
```diff
@@ -101,26 +167,73 @@ def compute(self):
...
-        super().update(preds=torch.sigmoid(preds), target=target.long())
+        preds, target = _preprocess_segmentation_inputs(preds, target, apply_sigmoid=True)
+        super(IoU, self).update(preds=preds, target=target.long())
 
     def compute(self):
         ious = super(BinaryIOU, self).compute()
         return {"target_IOU": ious[1], "background_IOU": ious[0], "mean_IOU": ious.mean()}
+
+
+class Dice(torchmetrics.JaccardIndex):
```
shaydeci, line 203: above comment regarding 0.7.3 torch metrics version

shaydeci (changes requested):
inline

src/super_gradients/training/metrics/segmentation_metrics.py
lkdci: Yes, IoU was renamed JaccardIndex at version `0.7.0`, see [here](https://github.com/PyTorchLightning/metrics/pull/662)

src/super_gradients/training/metrics/segmentation_metrics.py
lkdci: same as above

lkdci:
Added `AbstractMetricsArgsPrepFn` and `PreprocessSegmentationMetricsArgs` as preprocess inputs callable function to be called on the metrics update.

shaydeci (approved):
lgtm

PR #177, opened 2022-04-18
Bugfix/sg 99 stdc single class bug

shaydeci:


src/super_gradients/training/losses/ohem_ce_loss.py
```diff
@@ -11,7 +11,8 @@ def __init__(self,
                  threshold: float,
                  mining_percent: float = 0.1,
                  ignore_lb: int = -100,
-                 num_pixels_exclude_ignored: bool = True):
+                 num_pixels_exclude_ignored: bool = True,
+                 criteria: torch.nn.Module = None):
```
lkdci, line 15: perhaps the type should be `_Loss` instead of `nn.Module`

src/super_gradients/training/losses/ohem_ce_loss.py
```diff
@@ -52,3 +53,21 @@ def forward(self, logits, labels):
         else:
             loss = loss[:num_mining]
         return torch.mean(loss)
+
+
+class OhemBCELoss(OhemCELoss):
+    """
+    OhemBCELoss - Online Hard Example Mining Binary Cross Entropy Loss
+    """
+    def __init__(self,
```
lkdci, line 62: Binary cross entropy can be used with multi-classes not only with binary tasks. In case of multi-classes the target tensor is with shape [BxCxHxW], and usually a one hot tensor.

lkdci (commented):
BTW  TODO that `DDRNetLoss` shouid be fixed as well to support binary task. DDRNet loss is actually a private case of STDCLoss when `num_detail_heads == 0`

src/super_gradients/training/losses/ohem_ce_loss.py
shaydeci: Yes thats one of the changes I wanted to make.
Ill change the class hierarchy so that we have OhemLoss which gets the criteria as an argument in the constructor, then also allow to pass a different criteria to StdcLoss as well

lkdci: In case of multi-classes we should keep the option of ignore labels

src/super_gradients/training/losses/ohem_ce_loss.py
```diff
@@ -52,3 +57,40 @@ def forward(self, logits, labels):
...
+                 threshold: float,
+                 mining_percent: float = 0.1,
+                 num_pixels_exclude_ignored: bool = True, ):
+        super(OhemBCELoss, self).__init__(threshold=threshold,
+                                          mining_percent=mining_percent,
+                                          num_pixels_exclude_ignored=num_pixels_exclude_ignored,
+                                          criteria=nn.BCEWithLogitsLoss(reduction='none'))
+
+    def forward(self, logits, labels):
+        return super(OhemBCELoss, self).forward(logits.squeeze(1), labels.float())
```
lkdci, line 96: This line will failed if num_classes > 1

lkdci (commented):
Inline

src/super_gradients/training/losses/ohem_ce_loss.py
```diff
@@ -24,9 +29,9 @@ def __init__(self,
         assert 0 <= mining_percent <= 1, "mining percent should be a value from 0 to 1"
         self.thresh = -torch.log(torch.tensor(threshold, dtype=torch.float))
         self.mining_percent = mining_percent
-        self.ignore_lb = -100 if ignore_lb is None or ignore_lb < 0 else ignore_lb
+        self.ignore_lb = ignore_lb
         self.num_pixels_exclude_ignored = num_pixels_exclude_ignored
-        self.criteria = nn.CrossEntropyLoss(ignore_index=self.ignore_lb, reduction='none')
+        self.criteria = criteria
```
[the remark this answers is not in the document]
shaydeci: added informative exceptions

src/super_gradients/training/losses/ohem_ce_loss.py
shaydeci: fixed.

src/super_gradients/training/losses/ohem_ce_loss.py
shaydeci: fixed

src/super_gradients/training/losses/ohem_ce_loss.py
shaydeci: fixed

PR #182, opened 2022-04-24
Init optimizer with class type

lkdci:
Add the option to pass a `torch.optim.Optimizer` class type in traininig params to initiate an optimizer object through the `build_optimizer`.

This way a custom / not supported optimizer class can be passed, while still make use of the utils functions `initialize_param_groups` and `zero_weight_decay_on_bias_and_bn`


shaydeci:
Im closing this PR down as our optimisers factory already supports external classes (see tests/unit_tests/factories_test.py
 as an example)

PR #188, opened 2022-04-27
docs: Comment all the stages of the loss

yurkovak:
As requested, adding more comments explaining every step of the loss

### DIFF
--- a/src/super_gradients/training/losses/yolox_loss.py
+++ b/src/super_gradients/training/losses/yolox_loss.py
@@ -14,6 +14,14 @@ from super_gradients.training.utils.detection_utils import calculate_bbox_iou_ma
 
 
 class IOUloss(nn.Module):
+    """
+    IoU loss with the following supported loss types:
+        * 'iou' for
+            (1 - iou^2)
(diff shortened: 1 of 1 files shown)


src/super_gradients/training/losses/yolox_loss.py
```diff
@@ -114,11 +150,13 @@ def compute_loss(self, predictions: List[torch.Tensor], targets: torch.Tensor) \
                 obj_target = transformed_outputs.new_zeros((total_num_anchors, 1))
                 fg_mask = transformed_outputs.new_zeros(total_num_anchors).bool()
             else:
+                # GT boxes to image coordinates
                 gt_bboxes_per_image = labels_im[:, 2:6] * self.im_size
                 gt_classes = labels_im[:, 1]
                 bboxes_preds_per_image = bbox_preds[batch_idx]
 
                 try:
```
yurkovak, line 158: Do we want to keep this? It's supposed to help when during training there is not enough memory for assignment. In this case it will be attempted again, but on CPU. I actually saw it happen a few times

shaydeci: Sure, lets keep it.

shaydeci (approved):
lgtm

PR #191, opened 2022-05-03
Feature/sg 117 yolox multistage training cb

shaydeci:


yurkovak:
What's up with the files diff, why does it include so many unrelated changes? Not sure how to review this

tests/unit_tests/train_stage_switch_test.py
```diff
@@ -0,0 +1,57 @@
...
+        train_params = {"max_epochs": 4, "lr_updates": [], "lr_decay_factor": 0.1, "lr_mode": "step",
+                        "lr_warmup_epochs": 0, "initial_lr": 1, "loss": "cross_entropy", "optimizer": 'SGD',
+                        "criterion_params": {}, "optimizer_params": {"weight_decay": 1e-4, "momentum": 0.9},
+                        "train_metrics_list": [Accuracy()], "valid_metrics_list": [Accuracy()],
+                        "loss_logging_items_names": ["Loss"], "metric_to_watch": "Accuracy",
+                        "greater_metric_to_watch_is_better": True, "ema": False, "phase_callbacks": phase_callbacks}
+
+        expected_reductions = ['mean', 'mean', 'sum', 'sum']
+        model.train(train_params)
+        self.assertListEqual(criterion_reductions, expected_reductions)
```
yurkovak, line 53: Nice idea for the test!
Minor: I don't like that you rely on the fact that editing `self.placeholder` in the callback will result in a change in the list passed into init, it's trickier to read. I think it's better to define `CriterionReductionAttributeCollectorCallback([])` in this scope as an object, pass it into the list and do
`self.assertListEqual(callback.placeholder, expected_reductions)`

Oh, and please rename it into something more informative than a "placeholder".

yurkovak (commented):
Neat! Minor inline

PR #196, opened 2022-05-18
fix: Build a correct YOLOX-nano with depthwise separable convs

yurkovak:
Adding new arch_param 'depthwise': False, which is needed for YOLOX-nano. When its True the following layers are replaced with depthwise-separable ones:
* `Conv`s* and blocks** in the backbone (all except `Focus`);
* Blocks and 2 last downsizing  `Conv`s in the head;
* `Conv`s in two branches in `DetectX` (all except Stem and predicting layers).

*Depthwise-separable Conv = `DepthWiseConv`
**Depthwise-separable block = a block with `DepthWiseConv` in Bottleneck

Changes in `C3` and `BottleneckCSP aren't backward-compatible, but these modules are used only in yolov5_base.py and csp_darknet53.py anyway.

### DIFF
--- a/src/super_gradients/training/models/detection_models/csp_darknet53.py
+++ b/src/super_gradients/training/models/detection_models/csp_darknet53.py
@@ -83,7 +83,8 @@ class DepthWiseConv(nn.Module):
                  padding: int = None):
         super().__init__()
 
-        self.dconv = Conv(input_channels, input_channels, kernel, stride, activation_type, padding, input_channels)
+        self.dconv = Conv(input_channels, input_channels, kernel, stride, activation_type, padding,
+                          groups=input_channels)
         self.conv = Conv(input_channels, output_channels, 1, 1, activation_type)
 
     def forward(self, x):
--- a/src/super_gradients/training/models/detection_models/csp_darknet53.py
+++ b/src/super_gradients/training/models/detection_models/csp_darknet53.py
@@ -77,14 +77,29 @@ class Conv(nn.Module):
         return self.act(self.conv(x))
 
 
-class Bottleneck(nn.Module):
-    # STANDARD BOTTLENECK
-    def __init__(self, input_channels, output_channels, shortcut: bool, activation_type: Type[nn.Module], groups=1):
+class DepthWiseConv(nn.Module):
+    # STANDARD CONVOLUTION
+    def __init__(self, input_channels, output_channels, kernel, stride, activation_type: Type[nn.Module],
+                 padding: int = None):
         super().__init__()
 
+        self.dconv = Conv(input_channels, input_channels, kernel, stride, activation_type, padding, input_channels)
+        self.conv = Conv(input_channels, output_channels, 1, 1, activation_type)
+
+    def forward(self, x):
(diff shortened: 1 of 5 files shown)


src/super_gradients/training/models/detection_models/csp_darknet53.py
```diff
@@ -77,14 +77,29 @@ def fuseforward(self, x):
         return self.act(self.conv(x))
 
 
+class DepthWiseConv(nn.Module):
+    # STANDARD CONVOLUTION
+    def __init__(self, input_channels, output_channels, kernel, stride, activation_type: Type[nn.Module],
+                 padding: int = None):
+        super().__init__()
+
+        self.dconv = Conv(input_channels, input_channels, kernel, stride, activation_type, padding, input_channels)
```
ofrimasad, line 86: can you please add the arg name `groups=imput_channels`

src/super_gradients/training/models/detection_models/yolov5_base.py
```diff
@@ -243,35 +248,39 @@ def __init__(self, arch_params):
...
 
         if arch_params.yolo_type != 'yoloX':
             self._modules_list.append(
                 Detect(num_classes, anchors, channels=[width_mult(v) for v in (256, 512, 1024)]))  # 24
         else:
             strides = anchors.stride
             self._modules_list.append(
-                DetectX(num_classes, strides, activation_type, channels=[width_mult(v) for v in (256, 512, 1024)]))  # 24
+                DetectX(num_classes, strides, activation_type, channels=[width_mult(v) for v in (256, 512, 1024)],
+                        depthwise=arch_params.depthwise))  # 24
```
ofrimasad, line 283: no need to check `arch_params.depthwise`. you already have local variable `depthwise` here

src/super_gradients/training/models/detection_models/yolov5_base.py
```diff
@@ -414,7 +423,7 @@ def _initialize_weights(self):
             elif t is nn.BatchNorm2d:
                 m.eps = 1e-3
                 m.momentum = 0.03
-            elif t in [nn.LeakyReLU, nn.ReLU, nn.ReLU6, nn.Hardswish]:
+            elif t in [nn.LeakyReLU, nn.ReLU, nn.ReLU6, nn.Hardswish, nn.SiLU]:
```
ofrimasad, line 426: is this related to the depthwise change somehow? or a separate bugfix?

ofrimasad (commented):
few minor comment inline

src/super_gradients/training/models/detection_models/csp_darknet53.py
yurkovak: Sure, added in d17de50

src/super_gradients/training/models/detection_models/yolov5_base.py
yurkovak: Separate indeed. I was just comparing the Nano model definition in our codebase vs the source and noticed that SiLU isn't inplace, unlike in the original implementation

PR #201, opened 2022-05-29
Feature/sg 39 improve train valid format

Louis-Dupont:


src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -102,6 +105,20 @@ class EvaluationType(str, Enum):
...
+
+@dataclass
+class EpochStats:
+    current: float = None
+    previous: float = None
+    best: float = None
+    change_from_previous: float = None
+    change_from_best: float = None
+    is_better_than_previous: bool = None
+    is_best_epoch: bool = None
```
Louis-Dupont, line 119: This could be use to standardize the format of the dictionaries "self.train_epoch_stats" and "self.train_epoch_stats" (instantiated line 895)

Pros:
- forces the format
- less repetition when creating new instances (just call EpochStats() instead of initializing all the fields 1 by 1)

Cons:
- Different coding style (Introduces dataclasses which are not used in other places)
- "Only" used in 3 places, and not necessary to have understandable code.

requirements.txt
```diff
@@ -25,4 +25,7 @@ pip-tools>=6.4.0
 pyparsing==2.4.5
 einops==0.3.2
 protobuf~=3.19.0
-deci-lab-client==2.22.1
\ No newline at end of file
+deci-lab-client==2.22.1
```
ranrubin, line 28: Double requirement given: deci-lab-client==2.21.1 (from -r requirements.txt (line 31)) (already in deci-lab-client==2.22.1 (from -r requirements.txt (line 28)), name='deci-lab-client')

ranrubin (changes requested):

src/super_gradients/training/utils/sg_model_utils.py
```diff
@@ -2,17 +2,136 @@
...
+    current: float = None
+    previous: float = None
+    best: float = None
+    change_from_previous: float = None
+    change_from_best: float = None
+    is_better_than_previous: bool = None
+    is_best_value: bool = None
+
+
+def update_monitored_value(previous_monitored_value: MonitoredValue, new_value: float,
```
shaydeci, line 40: Please add pep8 documentation.

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -1719,6 +1737,26 @@ def evaluate(self, data_loader: torch.utils.data.DataLoader, metrics: MetricColl
 
         if self.multi_gpu == MultiGPUMode.DISTRIBUTED_DATA_PARALLEL:
             logging_values = reduce_results_tuple_for_ddp(logging_values, next(self.net.parameters()).device)
+
+        logging_values = get_logging_values(loss_avg_meter, metrics, self.criterion)
+        pbar_message_dict = get_train_loop_description_dict(logging_values,
+                                                            metrics,
+                                                            self.loss_logging_items_names)
+
+        for monitored_value_name in self.valid_monitored_values.keys():
```
shaydeci, line 1746: Lets move this to a helper class method.

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -1719,6 +1737,26 @@ def evaluate(self, data_loader: torch.utils.data.DataLoader, metrics: MetricColl
 
         if self.multi_gpu == MultiGPUMode.DISTRIBUTED_DATA_PARALLEL:
             logging_values = reduce_results_tuple_for_ddp(logging_values, next(self.net.parameters()).device)
+
+        logging_values = get_logging_values(loss_avg_meter, metrics, self.criterion)
```
shaydeci, line 1741: You are overwriting the reduced result for logging_values with the local one when running in DDP (i.e the logging_values in line 1739).

We don't reduce while looping to save time, but in the end of course we'd want the average across processes (metrics are already reduced out of the box since we use torchmetrics, but not the loss loggings).

shaydeci (changes requested):
Minor comments.

src/super_gradients/training/sg_model/sg_model.py
Louis-Dupont: I moved it to a helper function because I had no need for "self" in it, so I thought it would be better like that to avoid overloading the SgModel class

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -1719,6 +1734,21 @@ def evaluate(self, data_loader: torch.utils.data.DataLoader, metrics: MetricColl
...
+        pbar_message_dict = get_train_loop_description_dict(logging_values,
+                                                            metrics,
+                                                            self.loss_logging_items_names)
+
+        self.valid_monitored_values = sg_model_utils.update_monitored_values_dict(
+            monitored_values_dict=self.valid_monitored_values, new_values_dict=pbar_message_dict)
+
+        if not silent_mode:
+            progress_bar_data_loader.write("===========================================================")
+            sg_model_utils.display_epoch_summary(epoch=context.epoch, n_digits=4,
```
shaydeci, line 1747: You are calling "display epoch summary" inside our evaluate function, which is used both by validate, and test. For the rather, it does not make sense. You need to make sure such prints are not done when calling SgModel.test

shaydeci (changes requested):
One minor inline comment.

shaydeci (approved):
LGTM

shaydeci (approved):

shaydeci (approved):

ranrubin (approved):

PR #204, opened 2022-05-30
Feature/sg 131 check env compliance

Louis-Dupont:


src/super_gradients/compliance/check_compliance.py
```diff
@@ -0,0 +1,122 @@
...
+
+import os
+import sys
+from pip._internal.operations.freeze import freeze
+from logging import getLogger, DEBUG
+from typing import List, Dict
+from pathlib import Path
+from packaging.version import Version
+
+logger = getLogger('deci-compliance')
```
ofrimasad, line 13: please change to SG-complience

src/super_gradients/compliance/check_compliance.py
```diff
@@ -0,0 +1,122 @@
...
+    logger.info(f'OS Name: {name}')
+    logger.info(f'OS Platform: {platform}')
+    if 'linux' not in platform.lower():
+        return ['Deci officially supports only Linux kernels. Some features may not work as expected.']
+    return []
+
+
+def get_libs_requirements() -> List[str]:
+    """Read requirement.txt from the root, and split it as a list of libs/version"""
+    with open(Path(__file__).parent.parent.parent.parent / "requirements.txt", "r") as f:
```
ofrimasad, line 30: I am sure that there is a nicer way than .parent.parent.parent.parent
perhaps use pkg_resources

src/super_gradients/compliance/check_compliance.py
```diff
@@ -0,0 +1,122 @@
...
+        requirements_str = f.read()
+    return requirements_str.split("\n")
```
ofrimasad, line 32: why not use f.readlines() instead? this implementation may fail if someone uses other type of line breaker

src/super_gradients/compliance/check_compliance.py
```diff
@@ -0,0 +1,122 @@
...
+def check_compliance() -> None:
+    """Run all the compliance test and log everything that does not meet requirements"""
+
+    requirement_checkers = {
+        'operating_system': verify_os,
+        'libraries': verify_installed_libraries,
+    }
+
+    compliance_errors = {}
+    logger.info('Deci Compliance Check Started')
```
ofrimasad, line 95: change to SuperGradients Compliance Check Started

src/super_gradients/compliance/check_compliance.py
```diff
@@ -0,0 +1,122 @@
...
+    logger.info(f'Checking the following components: {list(requirement_checkers.keys())}')
+    logger.info('_' * 20)
+    for test_name, test_function in requirement_checkers.items():
+        logger.info(f"Verifying {test_name}...")
+        try:
+            errors = test_function()
+            if errors:
```
ofrimasad, line 102: but the tests always return a list... this will always be true?

src/super_gradients/compliance/check_compliance.py
```diff
@@ -0,0 +1,122 @@
...
+            raise
+
+    if compliance_errors:
+        logger.fatal(
+            f'The current environment does not meet Deci\'s needs, errors found in: {", ".join(list(compliance_errors.keys()))}')
+    else:
+        logger.info('Great, Looks like the current environment meet\'s Deci\'s requirements!')
+
+
+if __name__ == '__main__':
```
ofrimasad, line 121: didn't we want this tool to be called automatically on `__init__`?

ofrimasad (commented):
my comments inline

src/super_gradients/compliance/check_compliance.py
Louis-Dupont: @shaydeci seemed to prefer to have the user call it explicitly when he wants, to avoid having all these checks pop everytime it is imported. You don't think so ?

src/super_gradients/compliance/check_compliance.py
Louis-Dupont: I didnt find any satisfying solution using pkg_resources, did you have something in mind ?
What blocked me is that pkg_resources refers to the python package "super-gradients/src/super_gradients" and not to the root "super-gradients" which includes the "requirements.txt" file.

The other approach I can think of is to use the path from where the script is run and then move to "requirements.txt". This is great if we run from the root, but if it is run from somewhere else then the path will change.
For this reason I prefered to refer to the path of "requirements.txt" using the compliance script path as a reference, and then move to the requirements path using ".parent".

I just splited "Path(__file__).parent.parent.parent.parent / "requirements.txt"" into 3 subpart to explicit a bit more what is done.

ofrimasad: wait... so this will not work if SG was installed using pip?

src/super_gradients/compliance/check_compliance.py
ofrimasad: I admit I did not know that. but to keep a clean readable code, please make it `if len(errors) > 0`

src/super_gradients/compliance/check_compliance.py
Louis-Dupont: Done in 5a61647931ffb0a9db54398e0ccd0beb91c18449

src/super_gradients/compliance/check_compliance.py
Louis-Dupont: It will work now. I added a step in the ci to copy the requirements.txt into the source code and also added it to the setup.py to have it installed along with the rest of the package

.circleci/config.yml
```diff
@@ -39,6 +39,10 @@ jobs:
       - deci-common/checkout_and_skip_build:
           check_version_file: true
       - deci-common/get_persisted_version_info
+      - run:
+          name: add requirements.txt to source code
+          command: |
+            cp requirements.txt src/super_gradients/.
```
ranrubin, line 45: no need for .
would suggest 
```
cp requirements.txt src/super_gradients/requirements.txt 
```

ranrubin (approved):

ranrubin (changes requested):
Please add the copy command before uploading the package as well
on line 97 

https://github.com/Deci-AI/super-gradients/blob/5ec80041b3aed846118badae35e19bbe44bbb591/.circleci/config.yml#L97

ranrubin (approved):

src/super_gradients/__init__.py
```diff
@@ -1,9 +1,18 @@
...
 
 __all__ = ['ARCHITECTURES', 'losses', 'utils', 'datasets_utils', 'DataAugmentation',
            'TestDatasetInterface', 'SgModel', 'KDModel', 'SegmentationTestDatasetInterface', 'DetectionTestDatasetInterface',
-           'ClassificationTestDatasetInterface', 'init_trainer', 'is_distributed', 'train_from_recipe', 'train_from_kd_recipe']
+           'ClassificationTestDatasetInterface', 'init_trainer', 'is_distributed', 'train_from_recipe', 'train_from_kd_recipe',
+           'env_sanity_check']
+
+logger = get_logger(__name__)
+
+if logger.isEnabledFor(logging.INFO):
```
ofrimasad, line 17: please fix the logic:
if logging.INFO:
   run and run write output to screen and log
else:
   run and write output to log only

ofrimasad (commented):
please fix the activation logic

src/super_gradients/sanity_check/env_sanity_check.py
```diff
@@ -0,0 +1,165 @@
...
+    else:
+        logger.info('Great, Looks like the current environment meet\'s Deci\'s requirements!')
+
+    # The last message needs to be displayed independently of DISPLAY_SANITY_CHECK_VAR
+    logger.setLevel(keep_log_level)
+    if display_sanity_check:
+        logger.info(f'**This check can be hidden by setting the env variable {DISPLAY_SANITY_CHECK_VAR}=False prior to import**')
+    else:
+        logger.info(f'**A sanity check is done when importing super_gradients for the first time.'
+                    f'You can see the details by setting the env variable {DISPLAY_SANITY_CHECK_VAR}=True prior to import**')
```
ofrimasad, line 161: you mean =False

src/super_gradients/sanity_check/env_sanity_check.py
```diff
@@ -0,0 +1,165 @@
...
+from super_gradients.common.abstractions.abstract_logger import get_logger
+
+logger = get_logger(__name__, log_level=logging.INFO)
+
+DISPLAY_SANITY_CHECK_VAR = "HIDE_SANITY_CHECK"
+LIB_CHECK_IMPOSSIBLE_MSG = 'Library check is not supported when super_gradients installed through "git+https://github.com/..." command'
+
+
+def _init_sanity_logger_verbosity():
+    """If the user has the env variable HIDE_SANITY_CHECK, the logs will only be stored in the log files.
```
ofrimasad, line 18: fix name of variable and default setting

ofrimasad (commented):
fix the naming the documentations please

src/super_gradients/sanity_check/env_sanity_check.py
Louis-Dupont: I changed variable names, now it should be good

ofrimasad (approved):
LGTM

PR #205, opened 2022-05-31
interact with the platform through the lab client SDK

daniel-deci:
This only addresses the immediate fix.

This is still not ideal. From this small fix it felt not intuitive to start building the mode name by myself.
What if the #train  returned a training context - one that returns data about the process that just ran. Something like
```python
  class ResultsContext(Enum):
    TRAINNING = 'training'
    CONVERSION = 'conversion'
    LAB_UPLOAD = 'lab_upload'
  
  class SgModel:
     ...
     def train(self, training_params: dict = dict()):  -> Dict[ResultsContext, dict]
``` 
That way, I'd just have to access the corresponding context in order to get the model name that the server returned....otherwise, what happens when tomorrow the server changes its logic and `_1_1` suffixes won't be a thing anymore?

### DIFF
--- a/tests/integration_tests/deci_lab_export_test.py
+++ b/tests/integration_tests/deci_lab_export_test.py
@@ -1,16 +1,19 @@
 import unittest
-from super_gradients import SgModel, \
-    ClassificationTestDatasetInterface
+from super_gradients import SgModel, ClassificationTestDatasetInterface
 from super_gradients.training.metrics import Accuracy, Top5
 from super_gradients.training.models import ResNet18
 from torch.optim import SGD
 from super_gradients.training.utils.callbacks import DeciLabUploadCallback, ModelConversionCheckCallback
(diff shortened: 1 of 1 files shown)


daniel-deci:
97% of the changes here are black related (formatting). The real change lays from line 57

daniel-deci:
FYI @shaydeci @ofrimasad The description is just an idea. Don't stick to the details of the implementation

daniel-deci:
@shaydeci why closing this

shaydeci:
> @shaydeci why closing this

Im closing it as it has been stale for a while, and I didn't really understand the importance of it (its also stated that this is just an idea?)
Once clarified and discussed we can reopen it, but we've been having too many stale PR's...

daniel-deci:
Yeah, the description is an idea that came as the result of the fix itself the PR is taking care of to a ticket you opened.

PR #209, opened 2022-06-01
Feature/sg 13 detection metrics update

shaydeci:
3rd PR as part of Yolox integration to SG- holds minor changes in detection metrics to support Yolox training.


src/super_gradients/training/metrics/detection_metrics.py
```diff
@@ -89,19 +91,30 @@ def ap_per_class(tp, conf, pred_cls, target_cls):
 
 
 class DetectionMetrics(Metric):
-    def __init__(self, num_cls,
-                 post_prediction_callback: DetectionPostPredictionCallback = None,
-                 iou_thres: IouThreshold = IouThreshold.MAP_05_TO_095,
-                 dist_sync_on_step=False):
-        """
+    """
+    DetectionMetrics
```
ofrimasad, line 95: don't you think it will make more sense to define normalize_targets in the callback instead?

Metrics are not expected to change the data, but a post prediction callback is

ofrimasad (commented):
if I understand correctly, the entire change was the normalize_targets.

I think that normalization belongs in the callback (in fact it is the callback that scales the predictions to image size, we can just tell him not to do that maybe)

shaydeci:
> if I understand correctly, the entire change was the normalize_targets.
> 
> I think that normalization belongs in the callback (in fact it is the callback that scales the predictions to image size, we can just tell him not to do that maybe)

Notice we are normalising the GT targets. This is due to the fact that our Yolox loss expects them to be unnormalized, so either this  piece of code would goes here, or we add denormalising in Yolox loss.

I could also add something targets_prep_fn but I honestly think that since this code is soon to be legacy this won't matter so much and just add more code to get rid of...

ofrimasad (commented):
ok. LGTM

ofrimasad (dismissed):

PR #216, opened 2022-06-13
Widen deci-lab-client dependency range

roikoren755:


roikoren755:
> why not == ? we had some bad experiences with the build process being broken without us changing anything (just by an update in the required libs)

Since it's an internal package, I thought we'd be ok with `~=`, which lets us use newer versions of the library as long the change isn't breaking (so v3 won't get installed, but v2.27 will). We can change it to `==`, just wanted it to be as non-disruptive as possible :)

ofrimasad (approved):
LGTM

PR #221, opened 2022-06-15
refactor: Readable scale

yurkovak:
The function `_norm_relative_box` does the inverse of what happens with the predicted values. 
The predicted values are **multiplied** by scale, so the targets should be **divided**. It is not a bug, just much more clear to read the formula like this. Confused me for some time, because I didn't notice the inverse of scale happening in init


shaydeci (approved):
lgtm

PR #222, opened 2022-06-15
fix: Get rid of steps

yurkovak:
Steps is the hardcoded variable that defines output stride of each predicting level. But we already have input size and feat side `DefaultBoxes`, so steps are calculable and it's error-prone to hard-code them. 
It's not used anywhere outside, only in
`fk = fig_size / np.array(steps)` 
where it calculates exactly feat_size, but with some rounding errors. 

Bottom line: it's more correct to just use the `feat_size` list directly instead of `fig_size / steps`


shaydeci (approved):
LGTM

PR #233, opened 2022-07-03
Feature/alg 511 reproduce ssd

yurkovak:
Changes:

- Add a training yaml of MobileNet v2 SSD;
- Document and refactor the loss;
- Fix GT to anchors matching: every GT is guaranteed to get a pair;
- Take anchor options from code into yaml;
- Fix steps (level strides) in `DefaultBoxes`;
- Fix targets offset issue;
- Move constants into arguments


src/super_gradients/recipes/coco_ssd_lite_mobilenet_v2.yaml
```diff
@@ -0,0 +1,131 @@
+defaults:
+  - training_hyperparams: default_train_params
+  - dataset_params: coco_detection_dataset_params
+  - arch_params: default_arch_params
+  - checkpoint_params: default_checkpoint_params
```
shaydeci, line 5: Don't we start from a pretrained backbone anymore?
If we do please also add it to the model zoo, and add whatever else needed in the yaml file for it to be loaded.

In case we do- also add support for normalising by imagenet mean and std in coco dataset interface.

src/super_gradients/recipes/coco_ssd_lite_mobilenet_v2.yaml
```diff
@@ -0,0 +1,131 @@
...
+experiment_suffix: res${dataset_params.val_image_size}
+experiment_name: ${architecture}_coco_${experiment_suffix}
+
+sg_model:
+  _target_: super_gradients.SgModel
+  experiment_name: ${experiment_name}
+  model_checkpoints_location: ${model_checkpoints_location}
+  multi_gpu: AUTO
+
+anchors_resolution: ${dataset_params.val_image_size}x${dataset_params.val_image_size}
```
shaydeci, line 86: Please add on top how one can use the right anchors for the different resolutions from the command line.

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -568,7 +568,7 @@ def box_area(box):
 
 
 def non_max_suppression(prediction, conf_thres=0.1, iou_thres=0.6, merge=False, classes=None,
-                        agnostic=False):  # noqa: C901
+                        agnostic=False, multi_label_per_box=None):  # noqa: C901
```
shaydeci, line 571: Add docs for multi_label_per_box

src/super_gradients/training/losses/ssd_loss.py
```diff
@@ -1,66 +1,129 @@
...
+            best_cell_per_target, best_cell_per_target_index = ious.max(1)
+            # make sure EACH target has a grid cell assigned
+            best_target_per_cell_index[best_cell_per_target_index] = torch.arange(len(targets)).to(device)
+            best_target_per_cell[best_cell_per_target_index] = 2.
 
-            target_locations[:, mask] = targets[indices[mask], 2:].T
-            target_labels[mask] = targets[indices[mask], 1]
+            mask = best_target_per_cell > self.iou_thresh
+            each_cell_target_locations[:, mask] = target_boxes[best_target_per_cell_index[mask]].T
+            each_cell_target_labels[mask] = target_labels[best_target_per_cell_index[mask]] + 1
```
shaydeci, line 117: Where are we taking care of this when computing mAP ?

shaydeci (changes requested):
Inline comments, mostly about documentation so please remember to edit the docs when you add/remove params.

Besides that- we need to add the pretrained checkpoint to the model zoo, and add an integration test for the pretrained checkpoint (pretrained_models_test)

src/super_gradients/training/losses/ssd_loss.py
```diff
@@ -1,66 +1,129 @@
...
 from super_gradients.training.utils.ssd_utils import DefaultBoxes
 
 
+class HardMiningCrossEntropyLoss(_Loss):
+    """
+    L_cls = [CE of all positives] + [CE of the hardest backgrounds]
+    where the second term is built from [3 * positive pairs] background cells with the highest CE
+    (the hardest background cells)
+    """
+
```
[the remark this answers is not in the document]
yurkovak: Added in e8ccb06

src/super_gradients/training/utils/ssd_utils.py
```diff
@@ -71,54 +98,17 @@ def __call__(self, order="xyxy"):
...
     post prediction callback module to convert and filter predictions coming from the SSD net to a format
     used by all other detection models
     """
 
-    def __init__(self, conf: float = 0.1, iou: float = 0.45, classes: list = None, max_predictions: int = 300,
+    def __init__(self, dboxes: DefaultBoxes, conf: float = 0.1, iou: float = 0.45, classes: list = None,
+                 max_predictions: int = 300,
                  nms_type: NMS_Type = NMS_Type.ITERATIVE,
-                 dboxes: DefaultBoxes = DefaultBoxes.dboxes300_coco(), device='cuda'):
+                 multi_label_per_box=True):
```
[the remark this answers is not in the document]
yurkovak: That was always a default with out NMS and using False degrades mAP. It is obviously slower with True, but not noticeably without a proper benchmarking. Yet another reason to allow switching it off

src/super_gradients/training/losses/ssd_loss.py
```diff
@@ -1,66 +1,129 @@
...
+            # one best GT for EACH cell (does not guarantee that all GTs will be used)
+            best_target_per_cell, best_target_per_cell_index = ious.max(0)
 
-            values, indices = torch.max(ious, dim=0)
-            mask = values > 0.5
+            # one best grid cell for EACH target
+            best_cell_per_target, best_cell_per_target_index = ious.max(1)
+            # make sure EACH target has a grid cell assigned
+            best_target_per_cell_index[best_cell_per_target_index] = torch.arange(len(targets)).to(device)
+            best_target_per_cell[best_cell_per_target_index] = 2.
```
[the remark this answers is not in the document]
yurkovak: Explained in e8ccb06

src/super_gradients/recipes/coco_ssd_lite_mobilenet_v2.yaml
yurkovak: Extensive experiments show that there is no improvement from using ImageNet pre-trained backbone for COCO detection, or we aren't doing it right. 
As for my run with pre-training and image normalization, it landed 0.4 mAP lower. 
We can still add `pretrained: False` to the yaml and make sure True would work correctly. WDYT?

src/super_gradients/training/losses/ssd_loss.py
yurkovak: I'm still unhappy about using offset, because it requires taking care of the shift in the post-prediction callback, any other external code like COCO API where ids are mapped to classes, implementing offset in any new potential dataset. And adding more code = more things can break. While this tiny `+1` a few lines below does the trick. 
LMK if you still want me to add the offset 🙃

yurkovak: Not really, NMS accepts objectness and other classes, then during execution it doesn't treat objectness as a class, as expected, and returns the class id starting from 0 (ignoring objectness). So whenever indices are involved, they are 0-based, accept for the target constructed in the loss here

shaydeci: I know, all Im saying is to let the users be aware that once we feed class labels to SSD post prediction callback, they should beware that the returned class ids (**for predictions**) will be 0 based. Just a small note in docs thats all.

src/super_gradients/training/pretrained_models.py
```diff
@@ -24,6 +24,7 @@
               "yolo_v5m_coco": "https://deci-pretrained-models.s3.amazonaws.com/YOLOv5+m/ckpt_best.pth",
               "yolo_v5l_coco": "https://deci-pretrained-models.s3.amazonaws.com/YOLOv5+l/ckpt_best.pth",
               "yolo_v5n_coco": "https://deci-pretrained-models.s3.amazonaws.com/YOLOv5+n/ckpt_best.pth",
+              "ssd_lite_mobilenet_v2": "https://deci-pretrained-models.s3.amazonaws.com/ssd_lite_mobilenet_v2/ckpt_best.pth",
```
shaydeci, line 27: Can you please add the usual transfer learning and mAP verification tests to tests/integration_tests/pretrained_models_test.py?

shaydeci (changes requested):
Let's just add the tests we talked about and a clarification note regarding the zero based classes returned from SSD post prediction callback, then we can LGTM.

src/super_gradients/recipes/coco_ssd_lite_mobilenet_v2.yaml
yurkovak: Added in 9508619

src/super_gradients/recipes/coco_ssd_lite_mobilenet_v2.yaml
shaydeci: If we didn't experience any improvement then sure, lets leave it.

src/super_gradients/training/pretrained_models.py
yurkovak: Added in b1e22b9, 00814be, 471143a

PR #234, opened 2022-07-07
Feature/sg 137 add coco-like mAP implementation

Louis-Dupont:


src/super_gradients/training/metrics/metric_utils.py
```diff
@@ -39,6 +39,40 @@ def calc_batch_prediction_detection_metrics_per_class(metrics, dataset_interface
...
+        ap, precision, recall, f1 = ap.cpu().numpy(), precision.cpu().numpy(), recall.cpu().numpy(), f1.cpu().numpy()
+        _, unique_classes = unique_classes.cpu().numpy().astype(np.int64)
+
+        if iou_thres.is_range():
+            precision, recall, f1 = precision[:, 0], recall[:, 0], f1[:, 0]
+        mean_ap, mean_precision, mean_recall, mean_f1 = ap.mean(), precision.mean(), recall.mean(), f1.mean()
+    else:
+        targets_per_class = torch.zeros(1)
+
+    if not silent_mode:
```
shaydeci, line 59: What are all of these prints ? We should only print what we usually do during training...

src/super_gradients/training/metrics/metric_utils.py
```diff
@@ -39,6 +39,40 @@ def calc_batch_prediction_detection_metrics_per_class(metrics, dataset_interface
     return results_tuple
 
 
+def calc_batch_prediction_detection_metrics_per_classV2(matching, dataset_interface, iou_thres, silent_mode,
+                                                      per_class_verbosity, class_names, test_loss):
+    images_counter = len(matching)
```
shaydeci, line 44: Is there any case that len(matching) is 0 ?
If not, we need to remove that if statement.
If there is, then line 72: results_tuple = (mean_precision, mean_recall, mean_ap, mean_f1, *test_loss.average)

might crash.

src/super_gradients/training/metrics/metric_utils.py
```diff
@@ -39,6 +39,40 @@ def calc_batch_prediction_detection_metrics_per_class(metrics, dataset_interface
     return results_tuple
 
 
+def calc_batch_prediction_detection_metrics_per_classV2(matching, dataset_interface, iou_thres, silent_mode,
```
shaydeci, line 42: Please add docs.

src/super_gradients/training/transforms/transforms.py
```diff
@@ -618,22 +618,23 @@ def __init__(self, input_dim, swap=(2, 0, 1), max_targets=50):
...
 
-        boxes = new_target[:, :4]
-        labels = new_target[:, 4]
         img, r = rescale_and_pad_to_size(img, self.input_dim, self.swap)
-        boxes = xyxy2cxcywh(boxes)
-        boxes *= r
-        boxes = cxcywh2xyxy(boxes)
-        new_target = np.concatenate((boxes, labels[:, np.newaxis]), 1)
+
+        def _pad_rescale_target(targets_in):
```
shaydeci, line 627: This isn't clean, lets take it out and pass r as an argument.
Also ad docs to it.

src/super_gradients/training/datasets/datasets_conf.py
```diff
@@ -35,3 +35,8 @@
     'cell phone', 'microwave', 'oven', 'toaster', 'sink', 'refrigerator', 'book', 'clock', 'vase', 'scissors',
     'teddy bear', 'hair drier', 'toothbrush'
 ]
+
+COCO_91_INDEX = [
```
shaydeci, line 39: Don't think we really need this (the name is pretty confusing as well).
We can just take the class indices from the segmentation classes- COCO_DEFAULT_CLASSES_TUPLES_LIST without the background class, can't we ?)

shaydeci (changes requested):
inline comments

src/super_gradients/training/datasets/detection_datasets/coco_detection.py
```diff
@@ -1,5 +1,27 @@
...
+
+import numpy as np
+import torch
+
 from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataSet
-from super_gradients.training.datasets.datasets_conf import COCO_DETECTION_CLASSES_LIST
+from super_gradients.training.datasets.datasets_conf import COCO_DETECTION_CLASSES_LIST, COCO_91_INDEX
+
+
+def convert_xywh_to_cxcywh(bbox: List[int], img_width: float, img_height: float):
```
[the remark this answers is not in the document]
Louis-Dupont: There is no function that does it directly, but we do have "convert_xywh_bbox_to_xyxy" and "xyxy2cxcywh" that could be combined.
That being said, these 2 functions must be applied over np.arrays of shape (batch_size x 6) and have regular lists of shape (6).
I moved it to detection_utils in commit 11c4126ddb1d4c64e58acdc6f40940598ad61967

src/super_gradients/training/metrics/metric_utils.py
Louis-Dupont: this function was removed because outdated and never used: f5182088bd85175c86d345622c1242a924c989e7

src/super_gradients/training/metrics/metric_utils.py
Louis-Dupont: same, this function was removed because outdated and never used: f5182088bd85175c86d345622c1242a924c989e7

src/super_gradients/training/metrics/metric_utils.py
Louis-Dupont: Same as previous comment, this function was removed because outdated and never used: f5182088bd85175c86d345622c1242a924c989e7

src/super_gradients/training/datasets/datasets_conf.py
Louis-Dupont: I removed COCO_91_INDEX  from datasets_conf and generated it like that:
`
        coco_detection_category_id_list = [
            category_id
            for category_id, category_name in COCO_DEFAULT_CLASSES_TUPLES_LIST
            if category_name in COCO_DETECTION_CLASSES_LIST
        ]
`

see https://github.com/Deci-AI/super-gradients/blob/d0c79c703fb824c3ec4c6c4f01b241c27cdfaeae/src/super_gradients/training/datasets/detection_datasets/coco_detection.py

src/super_gradients/training/transforms/transforms.py
Louis-Dupont: Changing the function to be a method here d1a317718140d2b6e8ebcd47456823a3fb9a1ee7
Doc added here 15164be538e7291878fc9f0a2367b4e20a0f15d2

src/super_gradients/training/datasets/dataset_interfaces/dataset_interface.py
```diff
@@ -597,7 +598,7 @@ def __init__(self, dataset_params=None):
 
 class CoCoDetectionDatasetInterface(CoCoDataSetInterfaceBase):
     def __init__(self, dataset_params=None, cache_labels=False, cache_images=False, train_list_file='train2017.txt',
-                 val_list_file='val2017.txt'):
+                 val_list_file='val2017.txt', with_crowd=False):
```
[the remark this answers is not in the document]
Louis-Dupont: Is it all that needs to be changed to work with it as a dataset_param ? 4cc1bfe92cba1ce6725cd8b14d5df9749f89aeaa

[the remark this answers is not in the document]
shaydeci: yes

shaydeci (approved):
LGTM

PR #237, opened 2022-07-07
fix: Support any model width multiplier

yurkovak:
Fixing the usage of width multiplier for any YOLO-based model. We used to be able to construct only certain multiplier which aren't know in advance, but now any can be used, which is verified in the test I'm adding.

The problem was 2-fold:
1. Say someone attempts to use width_mult of 1.33. In this case the last head layer had input width_mult(1024), which is meant operate on two feature maps of width_mult(512) concatenated prior. But `int(1.33*512)` =  680, so the width after concat will be 680*2 = 1360, but `int(1.33*1024)` =  1361, which creates channel mismatch.
2. It's problematic to aways do `width_mult(connector[1])` where connector is a pre-calculated number of channels after concat between a head and a backbone layers. In such case using this head with a fixed backbone (no width scaling in it) becomes very cumbersome. You have to multiply backbone channels with a width back and forth to reach a correct connector (obviously not backbone layers per se, just numbers). This causes rounding issues and makes many multipliers impossible to use.

Simple solution:
Define input channels of layers after concat as a sum of two scaled* inputs, allow not scaling the backbone   

*scaled = width multiplier applied

### DIFF
--- a/src/super_gradients/training/models/detection_models/yolov5_base.py
+++ b/src/super_gradients/training/models/detection_models/yolov5_base.py
@@ -14,6 +14,7 @@ from super_gradients.training.utils.detection_utils import non_max_suppression, 
 from super_gradients.training.utils.export_utils import ExportableHardswish, ExportableSiLU
 from super_gradients.training.utils.utils import HpmStruct, check_img_size_divisibility, get_param
 
+
 COCO_DETECTION_80_CLASSES_BBOX_ANCHORS = Anchors([[10, 13, 16, 30, 33, 23],
                                                   [30, 61, 62, 45, 59, 119],
                                                   [116, 90, 156, 198, 373, 326]],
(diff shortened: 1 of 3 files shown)


yurkovak:
I ran pretrained_models_test to make sure models didn't change shape. 
This change shouldn't change old models that always "survived" rounding, just enable construction of new ones.

src/super_gradients/training/models/detection_models/yolov5_base.py
```diff
@@ -240,45 +242,49 @@ def __init__(self, arch_params):
...
-            block(width_mult(512), width_mult(512), depth_mult(3), activation_type, False, depthwise))  # 20
+            block(2 * width_mult(256), width_mult(512), depth_mult(3), activation_type, False, depthwise))  # 20
 
         self._modules_list.append(DownConv(width_mult(512), width_mult(512), 3, 2, activation_type))  # 21
         self._modules_list.append(Concat(1))  # 22
         self._modules_list.append(
-            block(width_mult(1024), width_mult(1024), depth_mult(3), activation_type, False, depthwise))  # 23
+            block(2 * width_mult(512), width_mult(1024), depth_mult(3), activation_type, False, depthwise))  # 23
 
+        detect_input_channels = [width_mult(v) for v in (256, 512, 1024)]
```
ofrimasad, line 280: why are 256, 512, 1024 hard coded here?

ofrimasad (dismissed):
LGTM just had a minor question there

yurkovak:
Forgot to update the yamls, please re-approve

ofrimasad (approved):
LGTM

PR #238, opened 2022-07-10
Fix failing yolo param groups test

yurkovak:
This test is supposed to make sure that yolo's `initialize_param_groups(...)` method doesn't leave some parameters out. Can happen if you change an architecture, but forget about this function. But params that don't require grad, like anchors, etc, shouldn't be in optimizer's params. That's a fix


ofrimasad (approved):
LGTM

PR #240, opened 2022-07-10
fixed

shaydeci:
update_param_groups was accidentally removed in https://github.com/Deci-AI/super-gradients/commit/27e472e4d0b360828f165fbd4e322ae8e5bc5edf


ofrimasad (commented):
not that fast... :) 
please add a test to prevent this from happening again

PR #242, opened 2022-07-10
refactor: Separate function for regnet slope to blocks

yurkovak:
Minor but useful refactor. We often need this parsing outside of RegNetX class, so people just end up copy-pasting this code every time. Would like to be able to import it


ofrimasad (approved):
LGTM

PR #243, opened 2022-07-10
feature: Add new options for anchorless detect module

yurkovak:
* Custom intermediate channels and groups in DetectX branches;
* GroupedConvBlock instead of DWConv (support group sizes other than one)

I double-checked that these changes don't change state dicts of YOLOX-N and YOLOX-S (two models that use the affected `DetectX`, one with depthwise True and one with False)


shaydeci (approved):
It'd be great if you can be a good citizen, and add these new params that can be tweakeed through arch params in docs.

Other than that, LGTM

shaydeci (approved):

PR #249, opened 2022-07-14
Backbone mode from arch_params

yurkovak:
Fixing the unexpected initialization: we usually put backbone_mode in arch_params


shaydeci (approved):
LGTM

PR #251, opened 2022-07-14
Save student only in ckpt_best for KD

shaydeci:
Checkpoint saving for best checkpoints logic changed in KDModel to support weight loading in SgModel for the student.


ofrimasad (approved):
LGTM

PR #252, opened 2022-07-17
Feature/sg 157 supprt natice torch loader

tomerkeren42:


ofrimasad (commented):
looks very nice. just a very minor comment.
@shaydeci, should we remove Anchors from here altogether or keep this functionality?

PR #257, opened 2022-07-20
allow using an external Optimizer (not initialized outside)

ofrimasad:


shaydeci (changes requested):
Can we please just add a small unit test verifying this feature works ?

shaydeci (approved):
LGTM

PR #263, opened 2022-07-20
Fix compute of precision & recall

Louis-Dupont:
Fix a bug where the computation of recall/precision would crash when all the confidence scores are above the "score_threshold".

### DIFF
--- a/src/super_gradients/training/utils/detection_utils.py
+++ b/src/super_gradients/training/utils/detection_utils.py
@@ -1672,6 +1672,24 @@ def compute_detection_metrics_per_cls(
     # Reversed cummax to only have decreasing values
     rolling_precisions = rolling_precisions.flip(0).cummax(0).values.flip(0)
 
+    # ==================
+    # RECALL & PRECISION
+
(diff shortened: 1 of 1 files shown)


src/super_gradients/training/utils/detection_utils.py
```diff
@@ -1672,6 +1672,24 @@ def compute_detection_metrics_per_cls(
...
 
+    # ==================
+    # RECALL & PRECISION
+
+    # We want the rolling precision/recall at index i so that: preds_scores[i-1] < score_threshold <= preds_scores[i]
+    # Note: torch.searchsorted works on increasing sequence and preds_scores is decreasing, so we work with "-"
+    highest_score_idx_below_thresh = torch.searchsorted(-preds_scores, -score_threshold, right=False)
+
+    # If all predictions are above the threshold, there is no i so that score_threshold <= preds_scores[i],
+    # So we need to take into account ALL the predictions (i.e. use the last index in the rolling precision/recall)
```
Louis-Dupont, line 1683: @ido-deci I just want to make sure this is the right behavior because it is different to the suggested fix

ofrimasad (approved):
LGTM

PR #279, opened 2022-07-27
feature: Allow custom warmup

yurkovak:


src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -907,7 +907,13 @@ def train(self, training_params: dict = dict()):  # noqa: C901
                                                            update_param_groups=self.update_param_groups,
                                                            **self.training_params.to_dict()))
         if self.training_params.lr_warmup_epochs > 0:
-            warmup_callback_cls = LR_WARMUP_CLS_DICT[self.training_params.warmup_mode]
+            warmup_mode = self.training_params.warmup_mode
+            if isinstance(warmup_mode, str):
+                warmup_callback_cls = LR_WARMUP_CLS_DICT[warmup_mode]
+            elif isinstance(warmup_mode, type) and issubclass(warmup_mode, PhaseCallback):
```
shaydeci, line 913: Let's be more specific- LRCallbackBase, its important so that lr scheduling won't intersect with warmup (should have different logic in is_lr_scheduling_enabled).

While at it, please add a unit test for training with external warmup.

src/super_gradients/training/sg_model/sg_model.py
```diff
@@ -1529,7 +1535,7 @@ def _initialize_sg_logger_objects(self):
 
             self.sg_logger = SG_LOGGERS[sg_logger](**sg_logger_params)
         else:
-            raise RuntimeError('sg_logger can be either an sg_logger name (str) or a subcalss of AbstractSGLogger')
+            raise RuntimeError('sg_logger can be either an sg_logger name (str) or an instance of AbstractSGLogger')
```
shaydeci, line 1538: Mistake I guess ?

shaydeci (changes requested):

src/super_gradients/training/sg_model/sg_model.py
yurkovak: Was a mistake, just sneaking in a fix into this PR

src/super_gradients/training/sg_model/sg_model.py
yurkovak: Base class fixed in 952bf7b
In 444e689 added a unit test for a custom warmup and also for an initial LR for linear warmup

shaydeci (approved):
LGTM

PR #292, opened 2022-08-02
Feature/sg 172 reproduce ssd with new coco dataset interface

shaydeci:
- Updated SSD lite mobilenet V2 recipe with new coco dataset intercase implementation.
- Optional bounding boxes filtering after affine transform.
- Updated pretrained model URL & mAP test.


ofrimasad (approved):
LGTM

PR #299, opened 2022-08-04
DDRNet train recipe

lkdci:
Improved recipe for DDRNet training.
Results:
| model | mIoU | previous %mIoU diff |
| ------ | ----- | --------------- |
| DDRNet23 | 80.26 | 1.61 |
| DDRNet23-slim | 78.01| 1.41 |

Differences:
* Fixed backbone pretrained weights loading. Only weight for context path in DDRNet were loaded with ImageNet weights while the ImageNet ckpt includes also weights for spatial path and context-spatial interchanged layers.
* Different LR for encoder and decoder.
* Dice + CE-Edge combo loss instead of Hard mining loss.


src/super_gradients/recipes/cityscapes_stdc_base.yaml
```diff
@@ -1,7 +1,7 @@
 # STDC Base training params
 
 defaults:
-  - training_hyperparams: cityscapes_stdc_train_params
+  - training_hyperparams: cityscapes_train_params
```
shaydeci, line 4: Lets rename them to cityscapes_default_train_params

shaydeci (changes requested):
Very Nice! just a tiny comment and then LGTM.

src/super_gradients/recipes/cityscapes_stdc_base.yaml
lkdci: Done

lkdci:
Addressed.

shaydeci (approved):
LGTM

PR #307, opened 2022-08-09
Hotfix/sg 000 fix image loading new detectiondataset

Louis-Dupont:
Changes done:
- Cache resized images instead of original image
- When loading image from cache, remove the pad (cached images need to be padded but we don't want to pad images at this step, this affects future steps)
- Copy the annotation as well, not only the image (This is problematic with augmentations that "accumulate" on the targets leading to False information)


src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -301,20 +305,24 @@ def get_sample(self, index: int) -> Dict[str, Union[np.ndarray, Any]]:
...
+        either from cache or by loading from disk, based on self.cached_imgs_padded
         :param index:  Image index
-        :return:       Resized image
+        :return:       rescaled image
         """
         if self.cache:
-            return self.cached_imgs[index].copy()
+            padded_image = self.cached_imgs_padded[index]
+            rescaled_height, rescaled_width = self.annotations[index]["rescaled_img_shape"]
+            rescaled_image = padded_image[:rescaled_height, :rescaled_width, :]
```
shaydeci: Is there ever going to be padding here?

src/super_gradients/training/datasets/detection_datasets/pascal_voc_detection.py
```diff
@@ -71,17 +71,17 @@ def _load_annotation(self, sample_id: int) -> dict:
...
         target[:, :4] *= r
 
-        initial_img_shape = (width, height)
-        resized_img_shape = (int(width * r), int(height * r))
+        initial_img_shape = (height, width)
+        rescaled_img_shape = (int(height * r), int(width * r))
 
         return {"img_path": img_path, "target": target,
-                "initial_img_shape": initial_img_shape, "resized_img_shape": resized_img_shape}
+                "initial_img_shape": initial_img_shape, "rescaled_img_shape": rescaled_img_shape}
```
shaydeci: Do we use at all "initial_img_shape" ?

shaydeci (changes requested):
Some comments inline.

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
Louis-Dupont: The cache does pad because it expects a fixed size tensor, it is done here https://github.com/Deci-AI/super-gradients/blob/6810c3c5888ae39b233a351ee459961d32bce3a3/src/super_gradients/training/datasets/detection_datasets/detection_dataset.py#L234-L235
So in the lines you commented we undo this padding, because otherwise it would affect the transforms

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -119,11 +120,12 @@ def __init__(
         if "target" not in self.target_fields:
             raise KeyError('"target" is expected to be in the fields to subclass but it was not included')
 
+        self.REQUIRED_ANNOTATION_FIELDS = {"target", "img_path", "rescaled_img_shape"}
```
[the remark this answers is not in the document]
Louis-Dupont: It's a because all the other attributes are variables (at least they depend on the init params), and this one is the only constant, so I used the convention for constants to emphasize on it.
But I don't mind, we can just make it private, I guess it is good as well

src/super_gradients/training/datasets/detection_datasets/pascal_voc_detection.py
Louis-Dupont: Not in the current code, I can remove it

Louis-Dupont: removed in 1205d4148559afc71221c13d55e4c940ec4a1c54

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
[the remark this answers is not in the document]
Louis-Dupont: changed in 1205d4148559afc71221c13d55e4c940ec4a1c54

shaydeci (approved):
LGTM

PR #309, opened 2022-08-10
Fix scale between rescaling batches

Louis-Dupont:
DetectionMultiscalePrePredictionCallback was changed to keep in memory the state of the new rescale size.

Tests done:
- The unit test "MultiScaleTest" was added.
- The recipe "coco2017_yolox" was run without changing any parameter for 1 epoch (it uses DetectionMultiscalePrePredictionCallback)


shaydeci (approved):
LGTM

PR #310, opened 2022-08-11
Feature/sg 221 reproduce yolox fastloss

Louis-Dupont:
The code is ready, we only need to train the different architectures and to fill the recipe with the results


ofrimasad (approved):
looks very nice!
If i don't mistake, YoloXFastDetectionLoss was already reviewed in previous PR (I see changes that I have requested)
LGTM

PR #314, opened 2022-08-14
added two assertions for num classes must be bigger than 1

tomerkeren42:
When trying to use IoU or Dice with num of classes == 1, things will collapse and a lot of bleeding can happen - unless you use the dedicated Binary equivalence. 
One (me) might not know the equivalence exist and sweat out trying to use the general ones.


NatanBagrov (dismissed):
LGTM, should save time :)

shaydeci (changes requested):
Don't use asserts, raise informative errors instead.

tomerkeren42:
> 

d1d91212

shaydeci (approved):
LGTM

shaydeci (approved):

PR #323, opened 2022-08-18
Feature/hydra output dir

danbochman:
Change behaviour of Hydra to output logs to user's checkpoints/experiment_name directory
Current behaviour spams output directories with unique timestamps for every run

This solves the bug where datasets such as cifar10, mnist, etc. were downloaded and extracted from scratch in each run.
In addition, this allows for easy loading of trained model, since now checkpoints are stored together with the YAML config that created them. A simple script can rebuild the model and load the weighs to it.

### DIFF
--- a/src/super_gradients/common/environment/env_helpers.py
+++ b/src/super_gradients/common/environment/env_helpers.py
@@ -2,6 +2,7 @@ import argparse
 import os
 import sys
 from functools import wraps
+from omegaconf import OmegaConf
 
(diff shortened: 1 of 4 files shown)


shaydeci (approved):
LGTM

ofrimasad (changes requested):
looks like this PR includes many redundant code changes which originate in your PyCharm lint settings. 
please note that we are using the 160 chars per line limit (not 80) and that strings can use ' and not only "

PR #336, opened 2022-08-29
fix: Prevent loosing weight_decay from params

yurkovak:
Fixed a crash in case a model's `initialize_param_groups` modifies `training_params`


shaydeci (approved):
LGTM

shaydeci (approved):
LGTM

PR #319, opened 2022-08-17
Fix kd-ema when using teacher_input_adapter

Louis-Dupont:

PR #340, opened 2022-08-30
teacher adpator freeze + map location cpu for all pretrained models

shaydeci:
Imagenet resent50 KD recipe was broken because of changes in https://github.com/Deci-AI/super-gradients/pull/319 .
(Input adaptor's parameters had require_grad=True which led to CUDA OOM).
Additionally map location=None in load_pretrained_weights caused to having extra GPU memory allocated in rank:0 when training in DDP. I removed it since the net is moved to the proper device in Trainer anyways.


yurkovak (approved):
LGTM

PR #341, opened 2022-08-30
[SG-251] Added CIFAR10/100 dataloaders, configs and tests

spsancti:


src/super_gradients/recipes/dataset_params/cifar100_dataset_params.yaml
```diff
@@ -0,0 +1,54 @@
+-batch_size: 256 # batch size for trainset
```
shaydeci, line 1: why the - ?

src/super_gradients/recipes/dataset_params/cifar100_dataset_params.yaml
```diff
@@ -0,0 +1,54 @@
...
+-val_batch_size: 512 # batch size for valset in DatasetInterface
+
+# TODO: REMOVE ABOVE, HERE FOR COMPATIBILITY UNTIL WE REMOVE DATASET_INTERFACE
+
+train_dataset_params:
+  root: /data/cifar100
+  train: True
+  transform:
```
shaydeci, line 9: It would look nicer if we align the "transform" arg as done in the other yaml files (i.e transforms).
This can be done if we reimplement the cifar10/100 datasets (that would inherit from the original ones of course) and resolve transforms with ListFactory(TransformsFactory()).

And while also speaking of resolving the transforms- lets take out the _target_ and use the factories for that.

src/super_gradients/recipes/dataset_params/cifar100_dataset_params.yaml
```diff
@@ -0,0 +1,54 @@
...
+        std: (0.2023, 0.1994, 0.2010)
+  target_transform: null
+  download: True
+
+train_dataloader_params:
+  batch_size: 256
+  num_workers: 8
+  batch_sampler: True
+  sampler:
+    InfiniteSampler:
```
shaydeci, line 28: This is not the default sampler thats used for cigar....we only use infinite sampler for YoloX training,
Just leave it blank and the sampler will be deririved by wether we're in DDP or not,

src/super_gradients/recipes/dataset_params/cifar10_dataset_params.yaml
```diff
@@ -1,2 +1,54 @@
...
+        mean: (0.4914, 0.4822, 0.4465)
+        std: (0.2023, 0.1994, 0.2010)
+  target_transform: null
+  download: True
+
+val_dataloader_params:
+  batch_size: 512
+  num_workers: 8
+  batch_sampler: True
+  sampler:
```
shaydeci, line 50: Same comment as above regarding the sampler.

shaydeci (changes requested):
Inline.

src/super_gradients/recipes/dataset_params/cifar100_dataset_params.yaml
spsancti: That's what was there in the beginning

PR #342, opened 2022-08-30
Feature/sg 223 cityscapes dataset refactor

danbochman:
- Refactored SegmentationDataset, Cityscapes Dataset
- Added params from CityscapesDatasetInterface to dataset YAML
- Changes to the way transforms are defined in YAML and expected in SegmentationDataset are not backwards compatible
- Cleaned a lot of unused code/methods in SegmentationDataset
- Moved hard-coded transformations (e.g. ToTensor, Normalize) to YAML for more visibility
- Many tests for dataset and dataloader
- Performed manual tests that all cityscapes training recipes are working


src/super_gradients/common/factories/base_factory.py
```diff
@@ -16,13 +18,14 @@ def get(self, conf: Union[str, dict, list]):
 
                 If provided value is not one of the three above, the value will be returned as is
         """
-        raise NotImplementedError
+        pass
```
shaydeci, line 21: Why did you remove the Error? This will lead to unexpected behaviour when users use the factory and will get a none value back, resulting in some other error at some point...

src/super_gradients/recipes/cityscapes_ddrnet.yaml
```diff
@@ -50,66 +52,33 @@ training_hyperparams:
...
 
-architecture: ddrnet_23
+
 experiment_name: ${architecture}_cityscapes
 
 model_checkpoints_location: local
 ckpt_root_dir:
 
-multi_gpu: DDP
+multi_gpu: "Off"
```
shaydeci, line 77: Should be DDP.

src/super_gradients/training/datasets/segmentation_datasets/segmentation_dataset.py
```diff
@@ -1,102 +1,55 @@
...
-    @resolve_param('image_mask_transforms_aug', factory=TransformsFactory())
+    @resolve_param('transforms', factory=TransformsFactory())
     def __init__(self, root: str, list_file: str = None, samples_sub_directory: str = None,
                  targets_sub_directory: str = None,
-                 img_size: int = 608, crop_size: int = 512, batch_size: int = 16, augment: bool = False,
-                 dataset_hyper_params: dict = None,
-                 cache_labels: bool = False, cache_images: bool = False, sample_loader: Callable = None,
-                 target_loader: Callable = None, collate_fn: Callable = None, target_extension: str = '.png',
-                 image_mask_transforms: transform.Compose = None, image_mask_transforms_aug: transform.Compose = None):
+                 cache_labels: bool = False, cache_images: bool = False,
```
shaydeci, line 18: These are breaking changes which you did not take care of on the dataset_interface side.
It can't be that we changed the signature and nothing is changed in dataset_interfaces.py.
This might cause the current train_from_recipe to crash (and even worse- it might continue the training and  just takes some weird defaults).

src/super_gradients/training/datasets/segmentation_datasets/segmentation_dataset.py
```diff
@@ -1,102 +1,55 @@
...
 
-        self.image_mask_transforms_aug = image_mask_transforms_aug or default_image_mask_transforms_aug
-        # FIXME: CROP SIZE CANNOT BE PASSED WHEN LIST
-        if image_mask_transforms is None:
-            image_mask_transforms = transform.Compose([Rescale(short_size=self.crop_size),
-                                                       CropImageAndMask(crop_size=self.crop_size, mode="center")
-                                                       ])
-
-        self.image_mask_transforms = image_mask_transforms
+        self.transforms = transform.Compose(transforms) if transforms else []
```
shaydeci, line 52: [] won't be subscriptable...

src/super_gradients/training/datasets/segmentation_datasets/cityscape_segmentation.py
```diff
@@ -89,16 +92,15 @@ def get_train_ids_color_palette(self):
         return self.train_id_color_palette
 
     @staticmethod
-    def target_transform(target):
+    def transform_target(target):
```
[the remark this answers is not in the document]
danbochman: Action -> Target is a more sensible convention imo
e.g. load_dataset, apply_transforms, train_model, load_checkpoint....

src/super_gradients/training/datasets/segmentation_datasets/cityscape_segmentation.py
```diff
@@ -89,16 +92,15 @@ def get_train_ids_color_palette(self):
...
         """
-        target_transform - Transforms the sample image
+        Transforms the sample image
```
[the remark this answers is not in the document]
danbochman: This is an automated refactor, the comment is the same as before - you want to change it?

src/super_gradients/common/factories/base_factory.py
danbochman: It is not removed, it is automatically applied by the @abstractmethod decorator and the @ABC inheritance
This is a better practice for abstract classes and methods, and gives additional useful properties when working with them (e.g. instructions by IDE, auto-complete, etc.)

src/super_gradients/training/datasets/segmentation_datasets/segmentation_dataset.py
danbochman: With the exception of the `transforms`, which were fixed in the latest commit for Cityscapes, nothing else was being used by any dataset that inherits from `SegmentationDataset`

src/super_gradients/training/datasets/segmentation_datasets/segmentation_dataset.py
danbochman: [] is passed to `Compose` so you can still do:
`transformed = self.transforms({"image": image, "mask": mask})`
Even if there are no transforms passed to the Dataset.
Otherwise you would need some kind of `if else` to handle it.
What is the scenario you are worried about?

shaydeci: I don't see where it will be resolved. Note that the factory resolved the argument, not the attribute (I looked up self.transforms usages, and didn't see it being passed to compose after this line).

danbochman: Got it. Nice catch!

PR #347, opened 2022-09-01
Feature/sg 252 coco dataset refactor

danbochman:
- coco segmentation dataset and datasetinterface fixed to work with new SegmentationDataset format
 - bugfix in coco seg train params
 - preparation to implement new dataset params format


shaydeci (approved):
LGTM

PR #359, opened 2022-09-08
Feature/sg 000 model download mock

ofrimasad:


src/super_gradients/training/utils/checkpoint_utils.py
```diff
@@ -272,3 +272,25 @@ def load_pretrained_weights(model: torch.nn.Module, architecture: str, pretraine
...
+    Loads pretrained weights from the MODEL_URLS dictionary to model
+    @param architecture: name of the model's architecture
+    @param model: model to load pretrinaed weights for
+    @param pretrained_weights: path tp pretrained weights
+    @return: None
+    """
+
+    map_location = torch.device('cpu')
+
+    pretrained_state_dict = torch.load(pretrained_weights, map_location=map_location)
```
shaydeci, line 289: I think we can avoid here some code duplication apearing  in load_pretrained_weights_local and load_pretrained_weights if we extract 290-296 to a helper method.

shaydeci (changes requested):
On SG side- one minor comment regarding code duplicates.

src/super_gradients/training/utils/checkpoint_utils.py
ofrimasad: fixed

ofrimasad:
moved to other PR

PR #370, opened 2022-09-12
fix: Plot detection in RGB

yurkovak:
The images are loaded in detection dataset in BGR, so I think it's safe to assume whatever is plotted in the same class is always BGR and should be converted to RGB before displaying.
R.n. we are plotting creepy blue pics


shaydeci (approved):
LGTM

PR #376, opened 2022-09-13
LAB-4040 Get model from platform

roikoren755:


requirements.txt
```diff
@@ -24,7 +24,8 @@ pyparsing==2.4.5
 einops==0.3.2
 pycocotools==2.0.4
 protobuf~=3.19.0
-deci-lab-client==2.38.0
+deci-common==3.4.1
+deci-lab-client==2.55.0b16158
```
ofrimasad, line 28: please use a released version

src/super_gradients/common/plugins/deci_client.py
```diff
@@ -1,24 +1,42 @@
...
+    def __init__(self, lab_client: DeciPlatformClient):
+        self.lab_client = lab_client
         GlobalHydra.instance().clear()
+        self.super_gradients_version = None
+        try:
+            import pkg_resources
+
+            self.super_gradients_version = pkg_resources.get_distribution("super_gradients").version
+        except Exception as e:
+            print(e)
```
ofrimasad, line 21: printing won't do any good. 
instead add `self.super_gradients_version = 3.0.0` as a fallback

src/super_gradients/common/plugins/deci_client.py
```diff
@@ -1,24 +1,42 @@
...
+except (ImportError, NameError):
+    from typing import Any as DeciPlatformClient
 
-    def __init__(self):
-        self.bucket_name = 'deci-model-repository-research'
-        self.prefix = 'mock_client'
-        self.s3 = boto3.client('s3')
+
+class DeciClient:
+    def __init__(self, lab_client: DeciPlatformClient):
```
ofrimasad, line 12: why not instantiate the DeciPlatformClient inside? why do we need to pass it all the way from outside?

ofrimasad (commented):
my comment inline

requirements.txt
roikoren755: yes, just waiting on a new version of the API to be released

src/super_gradients/common/plugins/deci_client.py
roikoren755: 👍

src/super_gradients/common/plugins/deci_client.py
roikoren755: The lab client needs to be logged in for anything to work.
With `deci-adk`, we provide colabs and snippets that initialize the client and log in with it first, and then pass it to the class to fetch the model using it

ofrimasad: but login can happen on `DeciPlatformClient.__init__()`. it's not like we expect the user to pass a user name or token in the code. these are taken from env_variable or local file (like with was cli)

roikoren755: The snippets we currently have do expect the user to pass a token in the code, though

ofrimasad: snippets are ok, but real code does not include tokens.
you can have an optional parameter `DeciPlatformClient.__init__(self, user: Optional[str], token: Optional[str])` but the DeciPlatformClient needs to know how to init without it (like wandb, was and other clients).
either take the token from env/file or prompt for login

roikoren755: 👍

ofrimasad:
moved to other PR

PR #393, opened 2022-09-18
Augment HSV will allow configuring bgr_channels in case there are oth…

NatanBagrov:
…er channels in the input tensor


ofrimasad (commented):
looks like setting the order of the colors is a bit too redundant. 
there is RGB and BGR but I don't think anyone uses GRB or RBG...
why not just add a flag 'rgb:bool = False' and then use 
```cv2.cvtColor(img, cv2.COLOR_RGB2HSV)```

NatanBagrov:
Sorry for not elaborating. The intention is to allow using this
augmentation with inputs of more than 3 channels (e.g, RGB+IR) while
applying it to the relevant channels.

ofrimasad (approved):
ok.
LGTM

PR #396, opened 2022-09-19
Trainer constructor cleanup

shaydeci:


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -76,30 +74,20 @@ class Trainer:
...
-                 train_loader: DataLoader = None, valid_loader: DataLoader = None, test_loader: DataLoader = None,
-                 classes: List[Any] = None):
+    def __init__(self, experiment_name: str, device: str = None,
+                 multi_gpu: Union[MultiGPUMode, str] = MultiGPUMode.OFF, ckpt_name: str = 'ckpt_latest.pth',
+                 ckpt_root_dir: str = None):
         """
 
         :param experiment_name:                      Used for logging and loading purposes
         :param device:                          If equal to 'cpu' runs on the CPU otherwise on GPU
         :param multi_gpu:                       If True, runs on all available devices
```
ofrimasad, line 84: can't we move multi gpu mode to training params as well?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -76,30 +74,20 @@ class Trainer:
...
-        :param model_checkpoints_location:      If set to 's3' saves the Checkpoints in AWS S3
                                                 otherwise saves the Checkpoints Locally
-        :param overwrite_local_checkpoint:      If set to False keeps the current local checkpoint when importing
                                                 checkpoint from cloud service, otherwise overwrites the local checkpoints file
         :param ckpt_name:                       The Checkpoint to Load
```
ofrimasad, line 87: why do we need ckpt_name here?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -859,6 +834,7 @@ def forward(self, inputs, targets):
 
         self.train_loader = train_loader or self.train_loader
         self.valid_loader = valid_loader or self.valid_loader
+        self._set_dataset_params()
```
ofrimasad, line 837: on each backward step? why?

ofrimasad: why not once in the beginning of train()

ofrimasad (commented):
very nice. but I think we can maybe remove even more parameters. 
my comment inline

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: ?

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: No, its important that in the constructor we specify whether we use DDP or not, as we initialise DDP in the constructor We do so because of the upcoming objects that would be initialised accordingly (for instance the dataloaders will use an appropriate sampler after doing so).

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: moved to train params in https://github.com/Deci-AI/super-gradients/pull/396/commits/b1d2e101f09d58e6e3374f72aef3d020d73e44ce

src/super_gradients/training/sg_trainer/sg_trainer.py
ofrimasad: ok

ofrimasad (approved):
LGTM

PR #405, opened 2022-09-22
Feature/sg 292 update register model

Louis-Dupont:
The main part of this PR is to adapt the "get" function to also be able to load custom models that are registered by anyone (and not implemented within SG)

Basically, SgModule all work with a single arg: arch_params. But we want to support any model, and to not force the user to adapt his module to our signature. This is especially needed because the user might want to use a model he did not define himself.
For that reason we have need 2 different behaviors: if it's SgModule or if it's not.
Ideally I think we should unify these cases (for instance by removing arch_params) because it can add confusion and will make the code harder to maintain but this would be a big task, so for now I came up with an "if/else" approach to handle both cases.


src/super_gradients/training/models/model_factory.py
```diff
@@ -20,98 +22,108 @@
...
+    :param pretrained_weights:  Describe the dataset of the pretrained weights (for example "imagenent")
+
+    :return:
+        - architecture_cls:     Class of the model
+        - arch_params:          Might be updated if loading from remote deci lab
+        - pretrained_weights:   Might be updated if loading from remote deci lab
+        - is_remote:            True if loading from remote deci lab
+    """
+    is_remote = False
+    if not isinstance(model_name, str):
```
ofrimasad, line 40: if we expect only a string, then let's write exactly that.
TypeError('parameter model_name is expected to be a string')

src/super_gradients/training/models/model_factory.py
```diff
@@ -20,98 +22,108 @@
...
+        raise ValueError("Unsupported model model_name " + str(model_name) + ", see docs or all_architectures.py for supported nets.")
+    elif isinstance(model_name, str) and model_name not in ARCHITECTURES.keys():
```
ofrimasad, line 42: no need to check isinstance(model_name, str) here (otherwise it would have go into the if)

src/super_gradients/training/models/model_factory.py
```diff
@@ -20,98 +22,108 @@
...
+            raise ValueError("Unsupported model name " + str(model_name) + ", see docs or all_architectures.py for supported nets.")
+        _arch_params = hydra.utils.instantiate(_arch_params)
+        _arch_params = HpmStruct(**_arch_params)
+        _arch_params.override(**arch_params.to_dict())
+        model_name, arch_params, is_remote = _arch_params["model_name"], _arch_params, True
+        pretrained_weights = deci_client.get_model_weights(model_name)
+    return ARCHITECTURES[model_name], arch_params, pretrained_weights, is_remote
+
+
+def instantiate_model(model_name: str, arch_params: dict, num_classes: int, pretrained_weights: str = None) -> SgModule:
```
ofrimasad, line 56: now this function does not return SgModule but nn.Module

ofrimasad (commented):
some comments inline

src/super_gradients/training/models/model_factory.py
Louis-Dupont: Changed in dd505a03c6cd553696f7d3e80ef8dfc25992af4c

src/super_gradients/training/models/model_factory.py
Louis-Dupont: Updated in dd505a03c6cd553696f7d3e80ef8dfc25992af4c

src/super_gradients/training/models/model_factory.py
```diff
@@ -16,102 +18,115 @@
...
-            arch_params.num_classes = num_classes_new_head
+        if pretrained_weights:
+            num_classes_new_head = core_utils.get_param(arch_params, "num_classes", PRETRAINED_NUM_CLASSES[pretrained_weights])
+            arch_params.num_classes = PRETRAINED_NUM_CLASSES[pretrained_weights]
 
+        # Most of the SG models work with a single params names "arch_params" of type HpmStruct, but a few take **kwargs instead
+        if "arch_params" not in get_callable_param_names(architecture_cls):
+            net = architecture_cls(**arch_params.to_dict(include_schema=False))
+        else:
+            net = architecture_cls(arch_params=arch_params)
```
Louis-Dupont, line 98: This is how I handle the case when our model does not work with arch_params.
 Is there a reason to replace:
`"arch_params" not in get_callable_param_names(architecture_cls)`
by
`get_callable_param_names(architecture_cls)[1] = "arch_params" ` ?

ofrimasad: no. i think this is ok

ofrimasad (approved):
LGTM

PR #407, opened 2022-09-28
PPLiteSeg recipes

lkdci:
Recipes to reproduce the pplite-seg IoU results.
This implementation has several differences compared to the original [paper](https://arxiv.org/abs/2204.02681):
* The original implementation used SSLD (self-supervised method) ImageNet pretraining which acc. to the paper " obtain enhanced pre-trained weights". I simply used the STDC weigths with regular supervised training.
* The location of the auxilary heads is a little different. I attached aux. heads straight from the encoder, while original attached after decoder UAFM blocks.
* Minor diffs in Hyper-params and Rnadom scaled in data-augmentations.
* The main difference and mIoU gain come from the ** Loss function**. I used our unified Combo Dice-CEEdge loss as implemented in STDC and DDRNet recipes. The original implementation used Hard mining cross-entropy only. For sanity check I reproduced the paper results with the same loss function for PPLite-T-Seg50 (~73 mIoU)


src/super_gradients/recipes/cityscapes_pplite_seg50.yaml
```diff
@@ -0,0 +1,100 @@
...
+      num_classes: 19
+      ignore_index: 19
+      num_aux_heads: 3
+      num_detail_heads: 0
+      weights: [ 1., 1., 1., 1. ]
+      dice_ce_weights: [ 1., 1. ]
+      ce_edge_weights: [ .5, .5 ]
+      edge_kernel: 5
+
+  sg_logger: wandb_sg_logger
```
ofrimasad, line 80: please remove wandb_sg_logger

ofrimasad: and the sg_logger_params of course

ofrimasad (commented):
looks beautiful. just one minor comment

src/super_gradients/recipes/cityscapes_pplite_seg50.yaml
lkdci: oops, removed [5d25844](https://github.com/Deci-AI/super-gradients/pull/407/commits/5d2584404c98135da02650f0c84b7913188c17fd)

ofrimasad (approved):
LGTM

PR #408, opened 2022-09-28
Fix load_backbone

Louis-Dupont:
Basically, we used to load the Trainer and now we load directly the module, but the code for load_backbone was not changed accordingly which broke the code when setting load_backbone=True


ofrimasad (approved):
LGTM

PR #409, opened 2022-09-29
FIX: Rename memory_cached

Louis-Dupont:
memory_cached is now deprecated by pytorch and we should use memory_reserved.
This will avoid a lot of warnings that we get in every run


ofrimasad (commented):
will this support older cuda/pytorch?
maybe move the logic to utils.get_gpu_mem_utilization 
and there check if torch.cuda has an attribute memory_reserved

Louis-Dupont:
> will this support older cuda/pytorch? maybe move the logic to utils.get_gpu_mem_utilization and there check if torch.cuda has an attribute memory_reserved
I introduced this check, but I think that it might not be necessary since memory_reserved was introduced in torch 1.4 and ask for 1.9 in requirements.

shaydeci (approved):
LGTM

PR #410, opened 2022-09-29
Remove AWS env check

Louis-Dupont:
It seems that the env variable that is checked is never used, am I missing something ?
If it is never used, I don't see a reason to keep the check in the code, especially since it prints a message every time we import SG.


ofrimasad (approved):
LGTM

ofrimasad (approved):
LGTM

PR #412, opened 2022-10-01
Mute logging for subprocesses

Louis-Dupont:
Currently, when running DDP, we see the logs of every subprocess.
There is a line in the code that should prevent this, but apparently it only mutes the prints, not the logs/warnings.
This PR aims at muting everything except for the errors for the non master nodes (rank>1), while leaving the master node untouched


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1393,15 +1392,14 @@ def _initialize_ddp(self):
...
+        if local_rank > 0:
+            mute_current_process()
+
+        logger.info("Distributed training starting...")
         if not torch.distributed.is_initialized():
             torch.distributed.init_process_group(backend='nccl', init_method='env://')
 
-        if local_rank > 0:
-            f = open(os.devnull, 'w')
-            sys.stdout = f  # silent all printing for non master process
```
Louis-Dupont, line 1403: Apparently this only muted the prints, not logs/warnings

shaydeci (approved):
LGTM

PR #413, opened 2022-10-01
Replace torch.tensor by my_tensor.clone().detach()

Louis-Dupont:
PyTorch raises a warning when using torch.tensor instead of my_tensor.clone().detach().


ofrimasad (approved):
LGTM

PR #414, opened 2022-10-02
Feature/sg 236 integrate clear ml

Louis-Dupont:
Note: there is a workaround that avoids the following error, and I am not sure if to keep the code like that


src/super_gradients/common/sg_loggers/clearml_sg_logger.py
```diff
@@ -0,0 +1,226 @@
...
+        self.save_logs = save_logs_remote
+
+    @multi_process_safe
+    def setup(self, project_name, experiment_name):
+        from multiprocessing.process import BaseProcess
+
+        # Prevent clearml modifying os.fork and BaseProcess.run, which can cause a DataLoader to crash (if num_worker > 0)
+        # Issue opened here: https://github.com/allegroai/clearml/issues/790
+        default_fork, default_run = os.fork, BaseProcess.run
+        self.task = Task.init(
```
ofrimasad, line 79: This will cause an ugly crash if clearml is not installed but this function is called.
please check if Task is defined and raise a nice error message (suggest installing clearml).

src/super_gradients/common/sg_loggers/clearml_sg_logger.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @multi_process_safe
+    def add_text(self, tag: str, text_string: str, global_step: int = 0):
+        super().add_text(tag, text_string, global_step)
+        self.clearml_logger.report_text(text_string)
+
+    @multi_process_safe
+    def add_figure(self, tag: str, figure: plt.figure, global_step: int = 0):
+        super().add_figure(tag, figure, global_step)
+        logger.warning(
```
ofrimasad, line 175: instead, just save the figure into a temp image and upload it as an image

ofrimasad (commented):
looks great. 
minor comments

src/super_gradients/common/sg_loggers/clearml_sg_logger.py
Louis-Dupont: Check added in f8a57a1711901a0a9a18a8dcb8c0fe579af0db4d

src/super_gradients/common/sg_loggers/clearml_sg_logger.py
Louis-Dupont: Added here: c99c2639fa3718156470d6b77a31091e8ac3ce37

ofrimasad (approved):
LGTM

PR #417, opened 2022-10-03
Update requirements.txt

tomerkeren42:
Pillow version was upgraded from 6.2.0 to 9.2.0, because of following bug was found while using pillow 8.2.0: Using transformations of random scale, which can be lower than expected crop-size, and then pad short size could lead to a bug in pillow version of padding - it was using a drawing rectangle (of zeros) instead of new method in newer versions. This method could create the following bug:
    self.draw.draw_rectangle(xy, ink, 0, width)
TypeError: an integer is required (got type tuple)


shaydeci (approved):
LGTM

PR #418, opened 2022-10-06
Improve cross-platform compatibility when parsing a readme description

BloodAxe:
Since README.md is in UTF-8 encoding, open() fails to read the file on non-mac platforms without explicit encoding argument.


shaydeci (approved):
LGTM

PR #419, opened 2022-10-07
Improve cross-platform compatibility when parsing a readme description

BloodAxe:
Follow-up to https://github.com/Deci-AI/super-gradients/pull/418 that addes explicit encoding to two remaining get_XXX functions that read contents from file.


ofrimasad (approved):
LGTM

PR #420, opened 2022-10-09
added the ability to download and import external code for models fro…

ofrimasad:
…m the deci-lab

fixed a bug in all_architectures.py (model names were missing or changed)


shaydeci (dismissed):
Let's just check that when we pip install this works, then LGTM.

I will also add to the backlog adding an integration tests as another part of the test plan for sg, that tests this mechanism.

shaydeci (approved):
lgtm

PR #422, opened 2022-10-10
Improve Windows platform support. 

BloodAxe:
Ensure that config directory & config file name does not contain backward slash symbol (Hydra fails to find config name in this case).


ofrimasad (approved):
Congrats on first PR!!

PR #426, opened 2022-10-11
Rename shelfnet classes_num param

Louis-Dupont:
For some reason shelfnet was implementing with the naming classes_num instead of num_classes.
This breaks the logic defined in model_factory, which expects num_classes.


shaydeci (approved):
LGTM

PR #428, opened 2022-10-18
Feature/sg 371 move sync bn to training params

shaydeci:
- sync_bn moved to training params so it can be ingested in Sg trainer (need to be there as we have to initialise ddp first..)
- Recipes using sync_bn updated- sync_bn is moved to their training params.
- Default arch_params yaml file is removed as it was the only flag there.
- Additional redundant reference in arch params in pretrained models test is removed.


ofrimasad (approved):
LGTM

PR #430, opened 2022-10-18
Feature/SG 344 - CSPResNet

BloodAxe:


src/super_gradients/training/models/detection_models/csp_resnet.py
```diff
@@ -0,0 +1,317 @@
...
+        self,
+        layers=[3, 6, 6, 3],
+        channels=[64, 128, 256, 512, 1024],
+        activation_type: Type[nn.Module] = torch.nn.SiLU,
+        return_idx=[1, 2, 3],
+        use_large_stem: bool = True,
+        width_mult=1.0,
+        depth_mult=1.0,
+        use_alpha: bool = False,
+    ):
```
ofrimasad, line 227: please add documentation

ofrimasad: please extract the default values into a yaml to at least a dictionary (to be later moved to yaml)

ofrimasad: you can use your ActivationsTypeFactory with `resolve_param('activation_type', ActivationsTypeFactory)`

src/super_gradients/training/models/detection_models/csp_resnet.py
```diff
@@ -0,0 +1,317 @@
...
+        block_fn: Type[nn.Module],
+        in_channels: int,
+        out_channels: int,
+        num_blocks,
+        stride: int,
+        activation_type: Type[nn.Module],
+        use_attention: bool = True,
+        use_alpha: bool = False,
+    ):
+        super().__init__()
```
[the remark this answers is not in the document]
ofrimasad: you can use your ActivationsTypeFactory with `resolve_param('activation_type', ActivationsTypeFactory)`

ofrimasad (commented):
looks very nice :)
please reuse available blocks and modules when possible. 
please add documentation to Model __init__ 
my comments inline

src/super_gradients/common/factories/activations_type_factory.py
```diff
@@ -0,0 +1,31 @@
...
+        """
+        Get a type.
+           :param conf: a configuration
+           if string - assumed to be a type name (not the real name, but a name defined in the Factory)
+           a dictionary is not supported, since the actual instantiation takes place elsewhere
+
+           If provided value is not one of the three above, the value will be returned as is
+        """
+        if isinstance(conf, str):
+            return get_activation_type(conf)
```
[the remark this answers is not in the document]
BloodAxe: The line of thinking for `ActivationsTypeFactory` was to benefit from `resolve_param`. 

A `get_activation_type()` method serves a few news:
* to being able to import existing activation functions from `torch.nn` and allowing some version-dependent dispatching (That's for future use. As an example: `if user asked for SiLU and torch version is < 1.6 import our own SiLU, otherwise use torch.nn.SiLU`).
* Handle lower-case activation functions names and well-known aliases (swish -> SiLU, leaky_relu, lrelu -> LeakyReLU)

We def. can merge them in a one class. It's just I'm not quite sure that implementation of `get_activation_type()` should be put inside the `ActivationsTypeFactory`. When used outside of configuration-based model instantiation it is easier to use `get_activation_type()`.

src/super_gradients/training/models/detection_models/csp_resnet.py
BloodAxe: > please extract the default values into a yaml to at least a dictionary (to be later moved to yaml)
Can you please elaborate on this?
CSPResNet was mean to be instantiated inside PPYolo, this is where extraction of params from config dict may take place. Would that work?

ofrimasad: regarding the parameters extraction, i meant that prefer not to have default parameters inside the code. many of our models already have those, but we plan on moving them all to yaml. We did the same process on DataLoaders a few weeks ago, so you can see now that dataloaders params are not defined as default values in the init().
the idea is to expose which parameters each block except, all in one place (the yaml). 

regarding the `resolve_param()`, I think it would be best to add it both here and in PP-Yolo. If an nn.Module is passed from PP-Yolo, it will just pass through, but if someone wants to use CSPResNet as the backbone to anything else (or even on its own), it will translate a string into nn.Module

ofrimasad (commented):
answered

src/super_gradients/training/models/detection_models/csp_resnet.py
BloodAxe: TY, I've added corresponding YAML files and indeed code became much cleaner :)

shaydeci (changes requested):
Looks really good!
One small thing- please add the unit test to deci_core_unit_test_suite_runner as we do with the other unit tests then we can merge (:

shaydeci (approved):
LGTM

PR #433, opened 2022-10-25
Feature/sg 143 black formatter

BloodAxe:
This PR adds a black formatter to check diff of a PR to have all code-formatting correct from black point of view (In addition to existing flake-8). There are some extra steps we can add in future (isort, mypy), but let's move step by step.


ranrubin (approved):
nice

ofrimasad (approved):
I loved using apostrophes so much :(
LGTM

PR #434, opened 2022-10-25
add deci requirements to ensure compatibility

Louis-Dupont:
deci-lab-client is not part of the requirements. It was removed not too long ago because we don't want all the super-gradients users to have it installed on their device, this is meant for pro users only.
But we need a simple way for the pro users to still be able to install super-gradients with all the "pro requirements".

To solve this, this PR introduces extra parameters:
"pip install super-gradients" will keep only install the basic requirements (no change)
"pip install super-gradients[pro]" will install super-gradients with all the requirements of pro users

Tested:
- pip install -e .[pro]
- pip install git+https://github.com/Deci-AI/super-gradients@hotfix/SG-000-add-deci-requirements-back-to-ensure-comptability#egg=super-gradients[pro]

Not tested:
- pip install super-gradients[deci-client] (from PyPI, requires a release)

TODO: change env_sanity_check accordingly


BloodAxe:
Does these new dependencies are actually required for every user? If we bringing a set of dependencies that only a part of the users require, maybe it's more justified to install them via `super_gradients[deci-lab]` to keep default distribution free of optional dependencies?

ofrimasad (commented):
@BloodAxe, these were removed not long ago. and we kept having issues with non-matching SG and client versions. 
what do you mean by `super-gradients[deci-client]`?

Louis-Dupont:
So we can chose to keep the same approach (requirements.txt/setup.py) and to add something like premium_requirements.txt, following https://hanxiao.io/2019/11/07/A-Better-Practice-for-Managing-extras-require-Dependencies-in-Python/
Or, we can change the current approach to work with pyproject.toml which seems to slowly become the standard. check: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

README.md
```diff
@@ -109,6 +109,11 @@ ________________________________________________________________________________
 pip install super-gradients
 ```
 
+Note: If you are a premium user of Deci, please run instead
```
ofrimasad, line 112: please remove it from README.
we don't want people installing this by mistake.

ofrimasad (commented):
just remove from README and we are ready to go

ofrimasad (approved):
LGTM

shaydeci (approved):

PR #435, opened 2022-10-26
when env sanity check fails - print error

ofrimasad:
when env sanity check fails with an error, always print it to the screen and in red


shaydeci (approved):
LGTM

PR #440, opened 2022-10-30
Feature-280/ throw error when using sequential sampler with DDP

Louis-Dupont:
Ready


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -923,10 +925,12 @@ def forward(self, inputs, targets):
         global logger
         if training_params is None:
             training_params = dict()
-
         self.train_loader = train_loader or self.train_loader
         self.valid_loader = valid_loader or self.valid_loader
         self._set_dataset_params()
+        if self.multi_gpu == MultiGPUMode.DISTRIBUTED_DATA_PARALLEL and isinstance(self.train_loader.sampler, SequentialSampler):
```
BloodAxe, line 931: Why don't we change `not isinstance(train_sampler, DistributedSampler)` instead? 
I imagine having a some kind of random sampler would also lead to processing through the entire dataset on each node.

Louis-Dupont: Definitly, I changed accordingly in b11ce3da6219b4f14c33dd8ea3aa4b7ccf86d2de

shaydeci: No @BloodAxe, the first option is the one we should go for.
We use many samplers that are intended be used in DDP setting (for example, repeated augmentations sampler).
In such cases, the user will pass the sampler explicitly to the DataLoader constructor. But this will raise an unwanted error.
The reason we check for SequentialSampler is that it is the default sampler when none is passed to the DataLoader constructor, which is most likely to be a mistake of the user. But if they pass something else- thats completely on them.

shaydeci (changes requested):
see comment in @BloodAxe and mine convo.

src/super_gradients/training/sg_trainer/sg_trainer.py
Louis-Dupont: Ok you're right, but in that case why not create a list of samplers that we support on DDP (InfiniteSampler, RepeatAugSampler and DistributedSampler) and raise a warning if it's not one of the three ?

shaydeci: Good idea, but since SequentialSampler shouldn't be used in DDP setting whatsoever- only in that case throw an error.

shaydeci (approved):
LGTM

PR #441, opened 2022-10-30
add list resolvers to OmegaConf to get items from list

ofrimasad:


BloodAxe:
Maybe use `getitem` instead of `list_get` to be consistent with python's naming scheme for getting a key from container?

src/super_gradients/common/environment/env_helpers.py
```diff
@@ -96,6 +96,9 @@ def register_hydra_resolvers():
     OmegaConf.register_new_resolver("class", lambda *args: get_cls(*args), replace=True)
     OmegaConf.register_new_resolver("add", lambda *args: sum(args), replace=True)
     OmegaConf.register_new_resolver("cond", lambda boolean, x, y: x if boolean else y, replace=True)
+    OmegaConf.register_new_resolver("getitem", lambda lst, i: lst[i], replace=True)                    # get the i'th object from a list (0 based)
```
shaydeci, line 99: I agree with @BloodAxe, lets rename it to list_get.

shaydeci (changes requested):
Just small renaming comment, then LGTM.

shaydeci (changes requested):
see inline comment regarding the docs

src/super_gradients/common/environment/env_helpers.py
ofrimasad: fixed in [0917439](https://github.com/Deci-AI/super-gradients/pull/441/commits/091743989a9a486ed41d5cbd86556bbb105823b0)

shaydeci (approved):
LGTM

PR #446, opened 2022-11-02
Adding some documentation on how to define a custom metric

Louis-Dupont:


src/super_gradients/common/registry/README.md
```diff
@@ -33,6 +33,12 @@ Recipes support out of the box every model, metric or loss that is implemented i
 
 ## 3. Examples
 ### A. Metric
+SuperGradients works with torchmetrics.Metric .
+To write your own metric you need to implement update() and compute() methods.
+
+In order to work on DDP you also need to define states, which are basically attributes that you want to share between your nodes, using add_state(). An example of state would be the number of correct predictions. You can see an example below.
```
shaydeci, line 39: I think this doesnt describe the state's purpose (also pretty inaccurate, the states aren't shared but reduced).
A better statement in my opinion would be:
"...which are attributes to be reduced, and broadcasted among the different ranks in compute() when training in distributed setting..."
then
"...would be the number of correct predictions, which will be summed across the different processes, broadcasted to all of them before computing the metric value..."

shaydeci (changes requested):
Small suggestion for change then LGTM

src/super_gradients/common/registry/README.md
Louis-Dupont: Yeah that's better, I changed it in 652c45f0179d82928b53b16ad36c9e2106114bd1

shaydeci (approved):
LGTM

PR #450, opened 2022-11-03
Feature/sg 321 ddp sampler handling for external dataloaders

shaydeci:
- Added the option for users to use dataloaders.get() with instantiated Dataset object. When doing so and not specifying the sampler argument- it will be derived according to the multi-gpu mode (i.e DistributedSampler in DDP setting and SequentialSampler otherwise) to reduce friction for inexperienced users.

- Unit test for the basic single gpu case added. DDP setting was tested manually and resulted in DistributedSampler as the data loader's sampler.


ofrimasad (approved):
LGTM

PR #451, opened 2022-11-05
Feature/sg 385 save console stream

Louis-Dupont:


ofrimasad (commented):
I don't think that keeping the file open for the entire run is a good practice.
instead, maybe we should use `with open(path) as file: file.write(...)`
I am not even sure how will the file copy work when the file is open.

@shaydeci , @BloodAxe what do you think?

src/super_gradients/common/auto_logging/console_logging.py
```diff
@@ -0,0 +1,112 @@
...
+class ConsoleSink:
+    """Singleton responsible to sink the console streams (stdout/stderr) into a file."""
+
+    def __init__(self):
+        self._setup()
+
+    @multi_process_safe
+    def _setup(self):
+        """On instantiation, setup the default sink file."""
+        default_path = Path.home() / "sg_logs" / "console.log"
```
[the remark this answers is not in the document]
BloodAxe: Pathlib works fine on Windows, so should not be an issue. I think the biggest advantage of Pathlib is this nice syntax sugar to allow writing `path / to / something` instead of os.path.join(path,to,something)`. But yes, it is cross-platform.

src/super_gradients/common/auto_logging/console_logging.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __init__(self, file):
+        self.file = file
+        self.stderr = sys.stderr
+        sys.stderr = self
+
+    def __del__(self):
+        sys.stderr = self.stderr
+        self.file.close()
+
```
Louis-Dupont, line 42: Note: delete stderr as well

src/super_gradients/common/auto_logging/console_logging.py
```diff
@@ -0,0 +1,112 @@
...
+    def write(self, data):
+        self.file.write(data)
+        self.stderr.write(data)
+
+    def flush(self):
+        self.file.flush()
```
Louis-Dupont, line 48: Open a file and write the buffer to it

src/super_gradients/common/auto_logging/console_logging.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+FILE_BUFFER_SIZE = 10_000
+
+
+class BufferWriter:
+    """File writer buffer that opens a file only when flushing and under the condition that threshold buffersize was reached."""
+
+    def __init__(self, filename: str, buffer_size: int):
+        self.buffer = StringIO()
```
ofrimasad, line 18: please add documentation. is buffer size in MB or in number of chars?

src/super_gradients/common/auto_logging/console_logging.py
```diff
@@ -0,0 +1,145 @@
...
+        self.filename = filename
+        self.buffer_size = buffer_size
+        self.open = open
+
+    def write(self, data: str):
+        self.buffer.write(data)
```
ofrimasad, line 24: if you have a defined buffer size, the point is to check if a flush is required after each write.
so you accumulate the strings into the buffer, but every once in a while it is written to file

src/super_gradients/common/auto_logging/console_logging.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def flush(self, force: bool = False):
```
ofrimasad, line 26: this implementation is not thread-safe.
you are replacing self.buffer without locking it first to writing.
so one thread might be writing to self.buffer while the other is dumping it.
you have to use 
```
# create a lock
lock = Lock()
# acquire the lock
lock.acquire()
# ...
# release the lock
lock.release()
```

ofrimasad (commented):
please see my comment inline. 
this requires a bit of a 'backend' coding way of thinking. 
maybe have @roikoren755 take a look

src/super_gradients/common/auto_logging/console_logging.py
Louis-Dupont: Mmhh I might be wrong on that point, but I think that self.buffer is not shared among the threads.

Also if I create a lock I need to pass it to my threads when creating the threads for it to be shared among the threads, no ?

src/super_gradients/common/auto_logging/console_logging.py
Louis-Dupont: Added in 8bfd183beb1ba34012bd58dd0ae3c3a91cf164c1 (number of chars)

src/super_gradients/common/auto_logging/console_logging.py
Louis-Dupont: I changed it like that: https://github.com/Deci-AI/super-gradients/blob/8bfd183beb1ba34012bd58dd0ae3c3a91cf164c1/src/super_gradients/common/auto_logging/console_logging.py#L22-L38

ofrimasad: looks ok

src/super_gradients/common/auto_logging/console_logging.py
Louis-Dupont: I pushed this c4569d78aff59d1acd0974e654d24019c3bd0e4a and it worked all fine on Coco dataset with DDP (x3).

ofrimasad (dismissed):
LGTM
just make sure this was tested with DDP for a few epochs

ofrimasad (dismissed):
LGTM

PR #452, opened 2022-11-06
Bug/sg 318 default optimizer params not taken with zero wd on bn and bias

shaydeci:
- With zero wd on bn and bias- we get unexpected behaviour since we inject the weight decay directly to the param groups. I updated the code so that we always override the default optimiser parameters with the user defined ones (if there are any, which is also something I took care of here by specifying "optimiser_params" to be an empty dictionary in case the user does not wish to pass any..)
- Added unit tests.


ofrimasad (approved):
LGTM

PR #454, opened 2022-11-06
Feature/sg 387 upload console logs for pro users

Louis-Dupont:
## Summary

**Flow**
- **On import**: if `deci_lab_client` is installed, a deci client will be instantiated and we wrap the exception handler to catch errors. Possibility to stop this behavior with `UPLOAD_LOGS=FALSE`
- **On exception** (from anywhere in the code): The exception is registered, so that we know later on that an exception was raised.
- **At exit**:  If an exception was registered, the console log file is uploaded to S3.
- **On upload**: A log is send to Sentry (monitoring tool used by platform team) that triggers an alert on Slack

**Slack message** (channel=support-logs): 
- Includes the path to the logs
- Includes client id (not the client name yet...)
image


src/super_gradients/common/pro_user/monitoring.py
```diff
@@ -0,0 +1,92 @@
...
+
+        sys.excepthook = register_exceptions(sys.excepthook)
+        atexit.register(exception_upload_handler, platform_client)
+    else:
+        logger.info("You disabled the upload of your logs to deci platform." "You can reactivate it by setting the env variable UPLOAD_LOGS=TRUE")
+
+
+def setup_user_env():
+    if _imported_deci_lab_failure is None:
+        setup_pro_user_monitoring()
```
Louis-Dupont, line 92: This means that if someone has the pro requirements we will by default try to upload to platform, unless an env variable is set.
This would apply to most deci members, so I guess everyone will need to set the env variable to skip it.
Are we fine with that ? Isnt there another way ?

src/super_gradients/common/pro_user/monitoring.py
```diff
@@ -0,0 +1,91 @@
...
+            logger.info("Exception was uploaded to deci platform!")
+        except Exception as e:  # We don't want the code to break at exit because of the client (whatever the reason might be)
+            logger.warning(f"Exception could not be uploaded to platform with exception: {e}")
+
+
+def setup_pro_user_monitoring():
+    """Setup the pro user environment for error logging and monitoring"""
+    upload_console_logs = os.getenv("UPLOAD_LOGS", "TRUE") == "TRUE"
+    if upload_console_logs:
+        logger.info("You installed pro user libraries, so we assume that you have a pro user account.")
```
ofrimasad, line 71: let's change this phrasing:
"deci-lab-client package detected. activating automatic log uploading"
"If you do not have a deci-lab-client credentials or you wish to disable this feature, please set the env variable UPLOAD_LOGS=FALSE"

src/super_gradients/common/pro_user/monitoring.py
```diff
@@ -0,0 +1,91 @@
...
+
+        logger.info("Connecting to the deci platform ...")
+        platform_client = DeciPlatformClient()
+        platform_client.login(token=os.getenv("DECI_PLATFORM_TOKEN"))
+        logger.info("Connection to the deci platform successful!")
+
+        sys.excepthook = register_exceptions(sys.excepthook)
+        atexit.register(exception_upload_handler, platform_client)
+    else:
+        logger.info("You disabled the upload of your logs to deci platform." "You can reactivate it by setting the env variable UPLOAD_LOGS=TRUE")
```
ofrimasad, line 86: here also:
"automatic log upload was disabled. to enable it please set the env variable UPLOAD_LOGS=TRUE

src/super_gradients/common/pro_user/monitoring.py
ofrimasad: can't we check if the user is logged in?

ofrimasad (commented):
few comments inline

src/super_gradients/common/pro_user/monitoring.py
```diff
@@ -0,0 +1,94 @@
+import os
```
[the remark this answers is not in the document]
Louis-Dupont: moved to commons/plugins/exception_monitoring.py

src/super_gradients/common/pro_user/monitoring.py
Louis-Dupont: There is a private function that we could use, and we could also ask platform team to make it public.
The main problem I see is that in that case if a pro user is not logged in yet (or deleted his token for any reason) then this feature would be disabled by default, and I guess we would rather avoid this situation.

src/super_gradients/common/pro_user/monitoring.py
Louis-Dupont: Changed in bd18acf6a499530dbf4f7c45e9b61e05ed77dd5c

src/super_gradients/common/pro_user/monitoring.py
Louis-Dupont: Changed in bd18acf6a499530dbf4f7c45e9b61e05ed77dd5c

ofrimasad (approved):
LGTM

PR #455, opened 2022-11-07
Bug/sg 301 calling train twice crashes

shaydeci:
- Device was different as we didn't call prep_model_for_train(). This is a result of the option we left to not pass a model to train() for backward compatibility which isn't relevant anymore (build_model()). Now it is called each time.
- There was als a problem with average_model's snapshots- we also need to call _init_monitored_items() and therefore I added a private flag to call it as well.
- Added unit test as well.


ofrimasad (approved):
LGTM

PR #457, opened 2022-11-08
Apply black to detection datasets and transforms

Louis-Dupont:
This PR introduces no functional change in the code, only "black" formatting.


ofrimasad (approved):
LGTM

PR #460, opened 2022-11-08
Feature/sg 195 bbox formats

BloodAxe:
As a part of SG-195 and also related to @Louis-Dupont's work on dataset targets format, this PR adds explicit Bbox formats specs to be used in both datasets & output adapter.

### DIFF
--- a/src/super_gradients/training/utils/bbox_formats/bbox_format.py
+++ b/src/super_gradients/training/utils/bbox_formats/bbox_format.py
@@ -1,0 +1,118 @@
+from abc import abstractmethod
+from typing import Union, Tuple
+
+import numpy as np
+import torch
+from torch import Tensor
+
+__all__ = ["BoundingBoxFormat", "convert_bboxes", "normalizedxyxy2xyxy_inplace", "normalizedxyxy2xyxy", "xyxy2normalizedxyxy_inplace", "xyxy2normalizedxyxy"]
+
+
+def normalizedxyxy2xyxy(bboxes: Union[Tensor, np.ndarray], image_shape: Tuple[int, int]) -> Union[Tensor, np.ndarray]:
+    """
+    Convert unit-normalized XYXY bboxes to XYXY bboxes in pixel units.
+    :param bboxes: BBoxes of shape (..., 4) in XYXY (unit-normalized) format
+    :param image_shape: Image shape (rows,cols)
+    :return: BBoxes of shape (..., 4) in XYXY (pixels) format
+    """
+    rows, cols = image_shape
+    if torch.is_tensor(bboxes):
+        scale = torch.tensor([cols, rows, cols, rows], dtype=bboxes.dtype, device=bboxes.device)
+        scale = scale.reshape([1] * (len(bboxes.size()) - 1) + [4])
+    elif isinstance(bboxes, np.ndarray):
+        scale = np.array([cols, rows, cols, rows], dtype=bboxes.dtype)
+        scale = scale.reshape([1] * (len(bboxes.shape) - 1) + [4])
+    else:
+        raise RuntimeError(f"Only Torch tensor or Numpy array is supported. Received bboxes of type {str(type(bboxes))}")
+
+    return bboxes * scale
+
+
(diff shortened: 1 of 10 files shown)


src/super_gradients/training/utils/bbox_formats/bbox_format.py
```diff
@@ -0,0 +1,118 @@
...
+        Convert XYXY boxes to target bboxes format
+        :param bboxes: Input bounding boxes [..., 4] in XYXY format
+        :param image_shape: Dimensions (rows, cols) of the original image to support
+                            normalized boxes or non top-left origin coordinate system.
+        :return: Converted bounding boxes [..., 4] in target format
+        """
+        raise NotImplementedError()
+
+
+def convert_bboxes(
```
Louis-Dupont, line 106: Is there a reason why you defined convert_bboxes as a function and not a method ?

It would look like this:
`def to(self, bboxes, image_shape, target_format):
    xyxy = self.to_xyxy(bboxes, image_shape, inplace)
    return target_format.from_xyxy(xyxy, image_shape, inplace)`

src/super_gradients/training/utils/bbox_formats/bbox_format.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Convert bboxes from source to target format
+    :param bboxes: Tensor of shape (..., 4) with input bounding boxes
+    :param image_shape: Tuple of (rows, cols) corresponding to image shape
+    :param source_format: Format of the source bounding boxes
+    :param target_format: Format of the output bounding boxes
+    :return: Tensor of shape (..., 4) with resulting bounding boxes
+    """
+    xyxy = source_format.to_xyxy(bboxes, image_shape, inplace)
+    return target_format.from_xyxy(xyxy, image_shape, inplace)
```
Louis-Dupont, line 118: Is there a reason why you defined convert_bboxes as a function and not a method of BoundingBoxFormat?

It would look like this:
`def to(self, bboxes, image_shape, target_format):
    xyxy = self.to_xyxy(bboxes, image_shape, inplace)
    return target_format.from_xyxy(xyxy, image_shape, inplace)`

BloodAxe: No particular reason. It just looked a bit better in the OutputAdapter:

```
class OutputAdapter(nn.Module):
  def __init__(self, ...):
    self.source_bbox_format = ...
    self.target_bbox_format = ...

  def forward(self, inputs):
    bboxes = self.get_bboxes(inputs)
    bboxes = convert_bboxes(bboxes, image_shape, self.source_bbox_format, self.target_bbox_format)
    ...
```

Now with the class method:
```
class OutputAdapter(nn.Module):
  def __init__(self, ...):
    self.source_bbox_format = ...
    self.target_bbox_format = ...

  def forward(self, inputs):
    bboxes = self.get_bboxes(inputs)
    bboxes = self.source_bbox_format.to(bboxes, image_shape, self.target_bbox_format)
    ...
```

To me the first version looks a little nicer. There is something wrong with the `self.source_bbox_format.to` which I cannot explain, but I feel we should not additional methods there. However if you see the latter approach would be helpful in some use-case - I don't have any _real_ objections against it.

tests/unit_tests/bbox_formats_test.py
```diff
@@ -0,0 +1,118 @@
...
+    NormalizedXYWHCoordinateFormat,
+    XYWHCoordinateFormat,
+    YXYXCoordinateFormat,
+    XYXYCoordinateFormat,
+    NormalizedCXCYWHCoordinateFormat,
+    convert_bboxes,
+)
+
+
+class BBoxFormatsTest(unittest.TestCase):
```
shaydeci, line 18: Nice tests.
I think it'd be also a good idea to take some fixed boxes conversion cases and test them as well, as I'm worried that these might bot cover the correctness of the conversion mechanism itself.

shaydeci (changes requested):
Nice.
See my comment inline regarding adding tests for some pre-defined conversion cases.

BloodAxe:
@shaydeci agree on pre-defined bboxes to check for regression. Added them in the recent commit.

shaydeci (approved):
LGTM

PR #466, opened 2022-11-12
Fix: pass extra args to subprocesses in ddp

Louis-Dupont:
Currently, we don't pass the extra (non-hydra) args to the subprocesses. e.g. if you run `python my_script.py --config-name=config x=0 --y=1 ` then the subprocesses will get `python my_script.py --config-name=config x=0`. 
This is done so because hydra does not support extra args, so we drop it and currently we lose this information after @hydra.main

Solution: When "poping" the extra non-hydra args, we store them in a variable and which can later on be used when launching DDP subprocesses.

Depends on: https://github.com/Deci-AI/super-gradients/pull/464 (black formatting)


shaydeci (approved):
LGTM

PR #467, opened 2022-11-12
Fix: get_ckpt_local_path not working when no default "checkpoints" folder

Louis-Dupont:
For some reason it looks like "get_ckpt_local_path" is not using environment_config.PKG_CHECKPOINTS_DIR which is the right way to refer to checkpoints dir. This can lead to a bug when there is no default checkpoint dir.

Solution: use "get_checkpoints_dir_path" which itself refers to environment_config.PKG_CHECKPOINTS_DIR.

Depends on: https://github.com/Deci-AI/super-gradients/pull/464 (black formatting)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ ________________________________________________________________________________
 # Load model with pretrained weights
 model = models.get("yolox_s", pretrained_weights="coco")
 ```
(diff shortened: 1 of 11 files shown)


shaydeci (approved):
LGTM

PR #471, opened 2022-11-13
Feature/sg 415 add registry callback and transform

Louis-Dupont:
Documentation is missing but I think this is a broader task:
- How to implement your own callback / Transform ?


ofrimasad (approved):
LGTM

PR #472, opened 2022-11-13
fix os.putenv('WANDB_BASE_URL') to os.environ["WANDB_BASE_URL"]

ofrimasad:
only line 79 is a real change. all the rest is black-formatting


shaydeci (approved):
LGTM

PR #474, opened 2022-11-13
Mute env_sanity_check for subprocess in DDP

Louis-Dupont:
Previous behavior:
All the processes (launching process & ddp nodes) run the check.

New behavior:
When running DDP, only the launching process will run env_sanity_check. None of the subprocesses (including the rank 0) will run the check.


src/super_gradients/common/environment/environment_config.py
```diff
@@ -12,4 +13,7 @@
 DDP_LOCAL_RANK = int(os.getenv("LOCAL_RANK", default=-1))
 EXTRA_ARGS = []
 
+# This needs to be run at the beginning because later on we drop --local_rank
+IS_TORCH_DISTRIBUTED_LAUNCH = any("--local_rank" in arg for arg in sys.argv)
```
ofrimasad, line 17: is that different from is)distributed() ? if so, that is confusing

ofrimasad (commented):
not sure i understand the difference between IS_TORCH_DISTRIBUTED_LAUNCH and is_distributed()

src/super_gradients/common/environment/environment_config.py
Louis-Dupont: - is_distributed() is True if the current process is a DDP node.
- IS_TORCH_DISTRIBUTED_LAUNCH is True if the process was launched with torch.distributed.launch

To summarize, when running `python train_from_recipe`:

> Launcher process: is_distributed() = False
> node 0: is_distributed() = True | IS_TORCH_DISTRIBUTED_LAUNCH = False
> other nodes: is_distributed() = True | IS_TORCH_DISTRIBUTED_LAUNCH = False

When running `python -m torch.distributed.launch ...`:
> node 0: is_distributed() = True | IS_TORCH_DISTRIBUTED_LAUNCH = True
> other nodes: is_distributed() = True | IS_TORCH_DISTRIBUTED_LAUNCH = True

ofrimasad: just an idea for other implementation:
when calling `LaunchConfig()` set `run_id="sg_initiated"`
you can then check for the run Id within the process (I think) and if you find that `run_id="sg_initiated"` than don't run the test.


If you decide to keep the current implementation, please make the two more differentiable:
have two functions: `is_distributed()`, `is_distributed_launch()`
place them next to each other and add clear documentation: 
"True if the current process is a DDP node."
"True if the process was launched with torch.distributed.launch"

Louis-Dupont: Sounds great, I pushed the change here a136e3dfad51a9cec068fba2ef98a2f97b19ca14

ofrimasad, line 17: why do we need this now?

ofrimasad (commented):
just a small cleanup

src/super_gradients/common/environment/environment_config.py
Louis-Dupont: I took it out in f434737b7551d1747cd3522d72f5dc1e2a03e6a4

ofrimasad (approved):
LGTM

PR #482, opened 2022-11-14
Update upload callback to platform

Louis-Dupont:


src/super_gradients/common/sg_loggers/deci_platform_sg_logger.py
```diff
@@ -26,17 +26,9 @@ def __init__(self, **kwargs):
...
-            raise ValueError(
-                'The environment variable "DECI_PLATFORM_TOKEN" is required in order to use '
-                "DeciPlatformSGLogger. Please set it with your own credentials "
-                "(available in https://console.deci.ai/settings)"
-            )
-
         super().__init__(**kwargs)
         self.platform_client = DeciPlatformClient()
-        self.platform_client.login(token=auth_token)
+        self.platform_client.login(token=os.getenv("DECI_PLATFORM_TOKEN"))
```
ofrimasad, line 31: but what if DECI_PLATFORM_TOKEN is not set?

ofrimasad (commented):
I don't think that is the logic we discussed. 
with this implementation, if DECI_PLATFORM_TOKEN env variable is not set, the client will not work.
we said:
- if the client is already logged in (I.e log file exists in the system) - pass no thing 
- if not, try to look for DECI_PLATFORM_TOKEN

src/super_gradients/common/sg_loggers/deci_platform_sg_logger.py
Louis-Dupont: If DECI_PLATFORM_TOKEN is not set, then the client will ignore directly look for a token stored locally.

Basically, the question is do we want to check env variable or local token first?  My motivation for checking env variable first are:
- This allows us to connect with a different token easily, by just setting the env variable. In most situations it is not useful, but I see a few edge cases where this can help. Otherwise, you need to delete the local token manually first.
- The client code is coded in a way that currently only supports this flow. The workaround would be to use private methods to get the local token, or to ask to the platform team to implement this env variable check within the client
- I don't see a downside of doing so, but maybe I am not thinking of something?

What's your thought?

ofrimasad (approved):
LGTM

PR #492, opened 2022-11-15
Add ignore option for dataloder workers on windows

Louis-Dupont:
I added a decorator for making sure a function does not run on dataloder workers.
This is needed to protect functions called when importing super_gradients on windows, because it seems that dataloader workers on windows re-import all the packages (see https://pytorch.org/docs/stable/data.html#multi-process-data-loading)

Especially:
image

Still need to be tested on windows

### DIFF
--- a/src/super_gradients/common/auto_logging/console_logging.py
+++ b/src/super_gradients/common/auto_logging/console_logging.py
@@ -6,7 +6,7 @@ from io import StringIO
 import atexit
 from threading import Lock
 
-from super_gradients.common.environment.env_helpers import multi_process_safe, is_main_process
+from super_gradients.common.environment.env_helpers import multi_process_safe, is_main_process, dataloader_worker_safe
 
 
 class BufferWriter:
@@ -114,6 +114,7 @@ class ConsoleSink:
         self._setup()
         atexit.register(self._flush)  # Flush at the end of the process
 
+    @dataloader_worker_safe
     @multi_process_safe
     def _setup(self):
         """On instantiation, setup the default sink file."""
@@ -136,6 +137,7 @@ class ConsoleSink:
                 f.write("============================================================\n")
         self.stdout.write(f"The console stream is logged into {self.filename}\n")
 
+    @dataloader_worker_safe
     @multi_process_safe
     def _set_location(self, filename: str):
         """Copy and redirect the sink file into another location."""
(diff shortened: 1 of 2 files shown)


BloodAxe:
Would be also great to annotate `env_sanity_check()` call in `super_gradients/__init__.py` with this decorator.

Louis-Dupont:
I added decorator for env_sanity_check here: 3455140ceb70533e35c3dc98e036b60d2f69e24b
Does it work ?

BloodAxe:
@Louis-Dupont unfortunately this check does not work. I imagine it's because we do env_check at the import time, while torch worker info may not be available yet. 

A small snippet to prove my statement:
```
import os
import torch
from torch.utils.data import Dataset, DataLoader

print("torch.utils.data.get_worker_info() and PID at import level", torch.utils.data.get_worker_info(), os.getpid())

class TestDataset(Dataset):
    def __init__(self, n):
        self.n = n

    def __len__(self):
        return self.n

    def __getitem__(self, item):
        return torch.utils.data.get_worker_info(), os.getpid()


if __name__ == "__main__":
    dataset = TestDataset(4 * 4)
    loader = DataLoader(dataset, batch_size=4, num_workers=4, shuffle=True, collate_fn=list)

    print("torch.utils.data.get_worker_info() and PID at main", torch.utils.data.get_worker_info(), os.getpid())

    for batch in loader:
        print("torch.utils.data.get_worker_info() and PID from batch")
        print(batch)
        print()
```


```
torch.utils.data.get_worker_info() and PID at import level None 19388
torch.utils.data.get_worker_info() and PID at main None 19388
torch.utils.data.get_worker_info() and PID at import level None 21448
torch.utils.data.get_worker_info() and PID at import level None 18888
torch.utils.data.get_worker_info() and PID at import level None 20528
torch.utils.data.get_worker_info() and PID at import level None 18052
torch.utils.data.get_worker_info() and PID from batch
[(WorkerInfo(id=0, num_workers=4, seed=7188228039362830794, dataset=<__main__.TestDataset object at 0x0000024F708EA460>), 18052), (WorkerInfo(id=0, num_workers=4, seed=7188228039362830794, dataset=<__main__.TestDataset object at 0x0000024F708EA460>), 18052), (WorkerInfo(id=0, num_workers=4, seed=7188228039362830794, dataset=<__main__.TestDataset object at 0x0000024F708EA460>), 18052), (WorkerInfo(id=0, num_workers=4, seed=7188228039362830794, dataset=<__main__.TestDataset object at 0x0000024F708EA460>), 18052)]
```

Louis-Dupont:
I have a dirty solution.
Basically, it would be to not log the processes that are subprocesses of python subprocesses when using DDP (like in the picture)
image
Without DDP, we can just not log the subprocesses.

And this only for windows, because mac/linux work all right.
I would like to talk about it, to see if there is a better way.

PR #493, opened 2022-11-16
Crash tip

Louis-Dupont:
When starting we get this message: 
image

ERRORS: 

Wrong format when creating a object with factory:
image

DDP not setup: 
image

Relies on https://github.com/Deci-AI/super-gradients/pull/496 (We might want to handle exceptions from within this decorator)


src/super_gradients/common/crash_handler/crash_tips.py
```diff
@@ -0,0 +1,209 @@
...
+            :return:                Tip
+        """
+        explanation = indent_string(cls._get_explanation(exc_type, exc_value, exc_traceback), indent_size=4)
+        solution = indent_string(cls._get_solution(exc_type, exc_value, exc_traceback), indent_size=4)
+
+        msg = (
+            "- What went wrong:\n"
+            f"{explanation}\n"
+            "- How to solve it:\n"
+            f"{solution}\n"
```
Louis-Dupont, line 79: Question:
Is it better to force the split of "Problem explanation" / "Solution" with 2 functions _get_explanation, _get_solution.
OR is it better to give the freedom to organise the tip in any way ?

src/super_gradients/common/crash_handler/crash_tips.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @classmethod
+    def _get_solution(cls, exc_type: type, exc_value: Exception, exc_traceback: TracebackType) -> str:
+        msg = (
+            "Please run at the beginning of your script:\n"
+            f">>> {fmt_txt('from super_gradients.training.utils.distributed_training_utils import setup_gpu_mode', color='green')}\n"
+            f">>> {fmt_txt('from super_gradients.common.data_types.enum import MultiGPUMode', color='green')}\n"
+            f">>> {fmt_txt('setup_gpu_mode(gpu_mode=MultiGPUMode.DISTRIBUTED_DATA_PARALLEL, num_gpus=...)', color='green')}"
+        )
+        return msg
```
Louis-Dupont, line 197: Question:
This tip aims at fixing an error which is predictable when the user params are wrong.
Do we want to add a check and raising a custom error within SG, or to leave it as is and add a crash tip at the end ?

src/super_gradients/common/crash_handler/crash_tips.py
```diff
@@ -0,0 +1,209 @@
...
+        except Exception:
+            # It is important that the crash tip does not crash itself, because it is called atexit!
+            # Otherwise, the user would get a crash on top of another crash and this would be extremly confusing
+            return None
+
+
+class TorchCudaMissingTip(CrashTip):
+    @classmethod
+    def is_relevant(cls, exc_type: type, exc_value: Exception, exc_traceback: TracebackType) -> bool:
+        pattern = "symbol cublasLtHSHMatmulAlgoInit version"
```
BloodAxe, line 118: This error may actually indicate CUDA libraries version conflict (When Torchvision & Torch are installed for different CUDA versions) and the absence of CUDA support in PyTorch.

src/super_gradients/common/crash_handler/crash_tips.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+# /!\ Only the CrashTips classes listed below will be used !! /!\
+ALL_CRASH_TIPS: List[Type[CrashTip]] = [TorchCudaMissingTip, RecipeFactoryFormatTip, DDPNotInitializedTip]
+
+
+def get_relevant_crash_tip_message(exc_type: type, exc_value: Exception, exc_traceback: TracebackType) -> Union[None, str]:
+    """Get a CrashTip class if relevant for input exception"""
+    for crash_tip in ALL_CRASH_TIPS:
+        if crash_tip.is_relevant(exc_type, exc_value, exc_traceback):
```
BloodAxe, line 207: What if we have more than one relevant tip? Maybe somehow account for such case and show multiple explanations/solutions.

BloodAxe:
That's a great looking tips IMO.
I'm a bit concerned tho on the way we setup this crash handler.
Suppose if I have my own unhandled exception handler. It would be conflicting with proposed solution.

Currently it is registering itself at import level as unhandled exception handler. What if we use try/catch instead and use a decorator:
```
@sg.with_crash_tips
def main(...):
   Trainer....
```

A ```with_crash_tips``` would be wrapping underlying function with try/catch. While achieving same goal as proposed solution it would not modify global exception handler and give user more granular control over enabling/disabling them in THEIR code.

src/super_gradients/common/crash_handler/crash_tips.py
Louis-Dupont: Good point, I think we should have 1 class per error, but with >=1 "tip"., each "tip" including 1 couple. of explanation/soluton.

src/super_gradients/common/crash_handler/crash_tips.py
Louis-Dupont: Thanks, I'll check that

Louis-Dupont:
> That's a great looking tips IMO. I'm a bit concerned tho on the way we setup this crash handler. Suppose if I have my own unhandled exception handler. It would be conflicting with proposed solution.
> 
> Currently it is registering itself at import level as unhandled exception handler. What if we use try/catch instead and use a decorator:
> 
> ```
> @sg.with_crash_tips
> def main(...):
>    Trainer....
> ```
> 
> A `with_crash_tips` would be wrapping underlying function with try/catch. While achieving same goal as proposed solution it would not modify global exception handler and give user more granular control over enabling/disabling them in THEIR code.

I agree with you on the

Louis-Dupont:
> That's a great looking tips IMO. I'm a bit concerned tho on the way we setup this crash handler. Suppose if I have my own unhandled exception handler. It would be conflicting with proposed solution.
> 
> Currently it is registering itself at import level as unhandled exception handler. What if we use try/catch instead and use a decorator:
> 
> ```
> @sg.with_crash_tips
> def main(...):
>    Trainer....
> ```
> 
> A `with_crash_tips` would be wrapping underlying function with try/catch. While achieving same goal as proposed solution it would not modify global exception handler and give user more granular control over enabling/disabling them in THEIR code.

That's a good point but this also applies to https://github.com/Deci-AI/super-gradients/pull/454. We need to discuss it

src/super_gradients/common/crash_handler/crash_tips.py
ofrimasad: yes. it should be `_get_explanation() -> str` and `_get_solutions() -> List(str)`

src/super_gradients/common/crash_handler/crash_tips.py
ofrimasad: If you can catch it and raise a good error, then yes. but make sure you are covering all cases, not just the main one

src/super_gradients/common/crash_handler/crash_tips.py
Louis-Dupont: In 8f1cc5d28b908456ea4accb8e65e4a66a267abbc I merged explanation and solution into _get_tips() -> List[str]

src/super_gradients/common/crash_handler/crash_tips.py
Louis-Dupont: I added option to show multiple explanations/solutions here: 8f1cc5d28b908456ea4accb8e65e4a66a267abbc
Basically, explanation and solution were merged, and now _get_tips() -> List[str] needs to be implemented for each CrashTip class

src/super_gradients/common/crash_handler/crash_tips.py
Louis-Dupont: I improved the tip accordingle: a1c3a16425b8fbaf66b3b65883751d3ec1390e8a

Louis-Dupont:
Are we good with setting it up by default, with option to disable it using an env variable ?
image

src/super_gradients/common/crash_handler/crash_tips_setup.py
```diff
@@ -0,0 +1,24 @@
...
+
+def crash_tip_handler():
+    """Display a crash tip if an error was raised"""
+    crash_tip_message = ExceptionInfo.get_crash_tip_message()
+    if crash_tip_message:
+        print(crash_tip_message)
+
+
+def setup_crash_tips() -> bool:
+    if os.getenv("CRASH_HANDLER", "TRUE") != "False":
```
ofrimasad, line 18: you are checking if the string value is "False" but in your message you  say "set your environment variable to CRASH_HANDLER=FALSE" 
will this work?

ofrimasad (commented):
one last question

src/super_gradients/common/crash_handler/crash_tips_setup.py
Louis-Dupont: No it's a mistake, I fixed it

ofrimasad (approved):
LGTM

PR #495, opened 2022-11-17
Feature/sg 416 albumentations plugin for classification

shaydeci:


src/super_gradients/training/transforms/all_transforms.py
```diff
@@ -118,3 +120,6 @@
     Transforms.RandomAutocontrast: RandomAutocontrast,
     Transforms.RandomEqualize: RandomEqualize,
 }
+
+ALBUMENTATIONS_TRANSFORMS = {name: cls for name, cls in inspect.getmembers(importlib.import_module("albumentations"), inspect.isclass)}
```
BloodAxe, line 124: inspect.isclass & issubclass(A.BasicTransform) ?

src/super_gradients/training/transforms/all_transforms.py
```diff
@@ -118,3 +120,6 @@
...
+ALBUMENTATIONS_COMP_TRANSFORMS = {name: cls for name, cls in inspect.getmembers(importlib.import_module("albumentations.core.composition"), inspect.isclass)}
```
BloodAxe, line 125: inspect.isclass & issubclass(A.BasicCompose) ?

requirements.txt
```diff
@@ -30,4 +30,5 @@ packaging>=20.4
 # not directly required, pinned by Snyk to avoid a vulnerability
 wheel>=0.38.0
 # not directly required, pinned by Snyk to avoid a vulnerability
-pygments>=2.7.4 
+pygments>=2.7.4
+albumentations>=1.3.0
```
BloodAxe, line 34: Do we want to add every augmentation plugin as a hard requirement? 
Maybe try to import albumentations in we encounter "Albumentations" in the transforms, and fail nicely if albumentations is missing? E.g we initialize the Albumentations factory on first call.

src/super_gradients/training/transforms/pipeline_adaptors.py
```diff
@@ -0,0 +1,37 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def post_transforms_processing(self, sample):
+        raise NotImplementedError
+
+
+class AlbumentationsAdaptor(TransformsPipelineAdaptorBase):
+    def __init__(self, composed_transforms: Callable):
+        super(AlbumentationsAdaptor, self).__init__(composed_transforms)
```
BloodAxe, line 25: I think this is where we should instantiate A.Compose as in addition to transforms this class allow to specify bbox_params which would be necessary for detection task.

```
transform = A.Compose(
    [A.RandomSizedBBoxSafeCrop(width=448, height=336, erosion_rate=0.2)],
    bbox_params=A.BboxParams(format='coco', min_area=0.5, label_fields=['category_ids']),
)
```

tests/integration_tests/albumentations_test.py
```diff
@@ -0,0 +1,61 @@
...
+from albumentations import Compose, HorizontalFlip, InvertImg
+
+
+class AlbumentationsIntegrationTest(unittest.TestCase):
+    def _apply_aug(self, img_no_aug):
+        pipe = Compose(transforms=[HorizontalFlip(p=1.0), InvertImg(p=1.0)])
+        img_no_aug_transformed = pipe(image=np.array(img_no_aug))["image"]
+        return img_no_aug_transformed
+
+    def test_cifar10_albumentations_integration(self):
```
BloodAxe, line 15: Let's add some tests for segmentation & detection (even if they fails) to sake of completeness. 
I may help with making them pass next week :)

src/super_gradients/training/transforms/pipeline_adaptors.py
shaydeci: I actually disagree, since this would limit the users to use Compose, whilst actually they can actually pass something else that does not need to be wrapped (see my use of ALBUMENTATIONS_COMP_TRANSFORMS).

src/super_gradients/training/transforms/all_transforms.py
shaydeci: Good point, added in 7ec67fb008f8880c4b75348f10d865cfa5ede9b1

src/super_gradients/training/transforms/all_transforms.py
shaydeci: Good point, added in https://github.com/Deci-AI/super-gradients/commit/7ec67fb008f8880c4b75348f10d865cfa5ede9b1

src/super_gradients/training/transforms/pipeline_adaptors.py
BloodAxe: A top-level Compose class is not only a container for augmentations, it handles way more on top of that. For detection, users will have to pass bbox_params to Compose to specify bounding box format that comes in. How this would look in YAML config for that case?

requirements.txt
shaydeci: sure, modified in 23a5d578de548870d10e193d7c5bd79e8cbf82c5

src/super_gradients/training/transforms/pipeline_adaptors.py
shaydeci: In exactly the same way we pass any other parameter to a transform in the YAML file, we just need to extend the support for A.BboxParams inside the factory.

shaydeci: transform:
  Albumentations:
    Compose:
      transforms:
        - RandomSizedBBoxSafeCrop: ...
      bbox_params:
        BboxParams: ...

tests/integration_tests/albumentations_test.py
shaydeci: I don't see what's the point if they are not supported at the moment. I think its best practice that everyone test their own code and not something planned for someone else to do in the future.

BloodAxe: Well, I respectfuly disagree. We know as a matter of fact that we will it for segmentation and detection tasks. So, initial implementation of the plugin should support it straight away on the level of API design.

tests/integration_tests/albumentations_test.py
```diff
@@ -0,0 +1,70 @@
...
+import numpy as np
+
+from super_gradients.training.datasets import Cifar10, Cifar100, ImageNetDataset
+
+try:
+    from albumentations import Compose, HorizontalFlip, InvertImg
+
+    _imported_albumentations_failure = None
+except (ImportError, NameError, ModuleNotFoundError) as import_err:
+    _imported_albumentations_failure = import_err
```
BloodAxe, line 12: What is the purpose of this try/catch and why it is thrown in _apply_aug down the road?

src/super_gradients/training/transforms/all_transforms.py
```diff
@@ -118,3 +124,25 @@
...
+    imported_albumentations_failure = import_err
+
+if imported_albumentations_failure is None:
+    ALBUMENTATIONS_TRANSFORMS = {
+        name: cls for name, cls in inspect.getmembers(importlib.import_module("albumentations"), inspect.isclass) if issubclass(cls, BasicTransform)
+    }
+    ALBUMENTATIONS_COMP_TRANSFORMS = {
+        name: cls
+        for name, cls in inspect.getmembers(importlib.import_module("albumentations.core.composition"), inspect.isclass)
+        if issubclass(cls, BasicTransform)
```
BloodAxe, line 144: BasicCompose instead of BasicTransform here?

tests/integration_tests/albumentations_test.py
shaydeci: fixed in https://github.com/Deci-AI/super-gradients/pull/495/commits/cd65908b3f3e57eebad13b15176eaa80ede1fe50

src/super_gradients/training/transforms/all_transforms.py
shaydeci: fixed in 77563f1e7432552e70ce693d1003454f0af515d9

ofrimasad (approved):
LGTM

ofrimasad (approved):
LGTM

PR #501, opened 2022-11-21
Feature/sg 423 epoch summary improvement

Louis-Dupont:
Currently, we only show the metric to watch and the losses in the epoch summary.

This PR aims at showing all the training/validation metrics.
To do so, we had to add an attribute to the SG metrics to know if "greater is better". Metrics implemented outside of SG that don't have such an attribute will still be displayed but there won't be any color to know if the result is better or worse.


shaydeci (approved):
LGTM

PR #503, opened 2022-11-21
Hotfix/ write "=" in epoch summary when metric/loss did not change

Louis-Dupont:
The previous implementation only handled
- increase/improvement
- decrease/worsening.

In this PR, I adapted the code to also handle:
- Same (value did'nt change compared to previous epoch/best epoch)
- None (No information, like for instance when working with a metric for which we don't know if greater is better)


shaydeci (approved):
LGTM

PR #507, opened 2022-11-21
Fix evaluate_from_config doc and add logs

Louis-Dupont:
I wrote a mistake in the previous fix, saying that best checkpoint is the default when it is the latest.

On top of that:
- I removed `core_utils.get_param(cfg.training_hyperparams, "ckpt_name", "ckpt_best.pth")` because we should always have `cfg.training_hyperparams.ckpt_name` 
- I added a small log, to let the user know what checkpoint is being evaluated


shaydeci (approved):
LGTM

PR #508, opened 2022-11-22
Automatic head replacement for lab models

shaydeci:
- Modified get_architecture method to return pretrained_weights_path rather then modifying the pretrained_weights variable, then it can be used the same fashion as local models with pretrained weights in our S3 bucket.


ofrimasad (commented):
one comment regarding the version

ofrimasad (approved):
LGTM

PR #511, opened 2022-11-23
Feature/sg 384 system logger

Louis-Dupont:
Monitor and write to tensorboard the system statistics, such as CPU usage, GPU, ...


src/super_gradients/common/environment/monitoring/data_models.py
```diff
@@ -0,0 +1,55 @@
...
+    sampling_fn: Callable
+    aggregate_fn: Callable
+    _samples: List = dataclasses.field(default_factory=list)
+
+    def sample(self):
+        self._samples.append(self.sampling_fn())
+
+    def aggregate(self) -> float:
+        value = self.aggregate_fn(self._samples)
+        self._samples = []
```
ofrimasad, line 27: please move this into a separate function `reset()`.
It took me a while to understand you are resetting on every `aggregate()` call

src/super_gradients/common/environment/monitoring/monitoring.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _sample(self):
+        """Sample the stat_aggregators, i.e. get the current value of each of them."""
+        for stat_aggregator in self.stat_aggregators:
+            stat_aggregator.sample()
+
+    def _aggregate_and_write(self):
+        """Aggregate and write the results."""
+        for stat_aggregator in self.stat_aggregators:
+            self.tensorboard_writer.add_scalar(tag=stat_aggregator.name, scalar_value=stat_aggregator.aggregate(), global_step=self.write_count)
```
ofrimasad, line 77: call the stat_aggregator.reset() here to make it more understandable

src/super_gradients/common/environment/monitoring/monitoring.py
```diff
@@ -0,0 +1,84 @@
...
+        while True:
+            for _ in range(self.n_samples_per_aggregate):
+                self._sample()
+                time.sleep(self.sample_interval)
+            self._aggregate_and_write()
+
+    def _sample(self):
+        """Sample the stat_aggregators, i.e. get the current value of each of them."""
+        for stat_aggregator in self.stat_aggregators:
+            stat_aggregator.sample()
```
ofrimasad, line 72: please wrap stat_aggregator.sample() with try-except. you never know which weird thing the Nvidia code can do and this will crash all of your `while True` loop

src/super_gradients/common/environment/monitoring/monitoring.py
```diff
@@ -0,0 +1,84 @@
...
+
+class SystemMonitor:
+    """Monitor and write to tensorboard the system statistics, such as CPU usage, GPU, ...
+
+    :param tensorboard_writer:  Tensorboard object that will be used to save the statistics
+    :param extra_gpu_stats:     Set to True to get extra gpu statistics, such as gpu temperature, power usage, ...
+                                Default set to False, because this reduces the tensorboard readability.
+    """
+
+    def __init__(self, tensorboard_writer: SummaryWriter, extra_gpu_stats: bool = False):
```
ofrimasad, line 19: to make it cleaner, please add a function `close()` that breaks the `while True`
```
while self.running:
   ...

def close():
    self.running = False
```

ofrimasad: you can call close from the sg_logger of course

src/super_gradients/common/environment/monitoring/network.py
```diff
@@ -0,0 +1,14 @@
+import psutil
+from super_gradients.common.environment.monitoring.utils import bytes_to_megabytes
+
+initial_net_io_counters = psutil.net_io_counters()
+
+
+def get_network_sent_mb() -> float:
+    """Number of MegaBytes sent since import"""
+    return bytes_to_megabytes(psutil.net_io_counters().bytes_sent - initial_net_io_counters.bytes_sent)
```
ofrimasad, line 9: here also, instead of aggregation from import, show the value from the last call.

src/super_gradients/common/environment/monitoring/network.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def get_network_recv_mb() -> float:
+    """Number of MegaBytes received since import"""
+    return bytes_to_megabytes(psutil.net_io_counters().bytes_recv - initial_net_io_counters.bytes_recv)
```
ofrimasad, line 14: here also, instead of aggregation from import, show the value from the last call.

src/super_gradients/common/environment/monitoring/network.py
ofrimasad: perhaps in the case of network and disk, it will be better to show numbers not between 100 samples (with only 0.1 sec between them), but instead to show numbers between one minute to the previous one (or over x seconds). this will require to add a `reset()`  function of course

ofrimasad (commented):
looks very nice. some comments inline

src/super_gradients/common/environment/monitoring/data_models.py
Louis-Dupont: Added in 4be886f65b51c41b1b20074c04d5d19b8a87ac1a

src/super_gradients/common/environment/monitoring/monitoring.py
Louis-Dupont: Right, changed in 4be886f65b51c41b1b20074c04d5d19b8a87ac1a

src/super_gradients/common/environment/monitoring/disk.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def get_disk_usage_percent() -> float:
+    """Disk memory used in percent."""
+    return psutil.disk_usage("/").percent
+
+
+def get_io_read_mb() -> float:
+    """Number of MegaBytes read since import"""
+    return bytes_to_megabytes(psutil.disk_io_counters().read_bytes - initial_net_io_counters.read_bytes)
```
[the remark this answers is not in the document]
Louis-Dupont: I changed it in 4f0ffa9124fd052d1aaa90e38677ac4268ba1ecf (I still have some comments to fix)

src/super_gradients/common/environment/monitoring/disk.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def get_io_write_mb() -> float:
+    """Number of MegaBytes written since import"""
+    return bytes_to_megabytes(psutil.disk_io_counters().write_bytes - initial_net_io_counters.write_bytes)
```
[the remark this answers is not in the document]
Louis-Dupont: Same, 4f0ffa9124fd052d1aaa90e38677ac4268ba1ecf

src/super_gradients/common/environment/monitoring/monitoring.py
Louis-Dupont: Done in 4be886f65b51c41b1b20074c04d5d19b8a87ac1a.
The logic is that if for any reason a "sampling" step fails, the metric will still try in the next step, and the aggregation will be done only on valid samples.
If no sample when aggregating, the scalar will just not be uploaded this time

src/super_gradients/common/environment/monitoring/network.py
Louis-Dupont: I changed it in https://github.com/Deci-AI/super-gradients/commit/4f0ffa9124fd052d1aaa90e38677ac4268ba1ecf (I still have some comments to fix)
I chose to do it "per second" to have a constant metric that does not depend on how often we aggregate

src/super_gradients/common/environment/monitoring/network.py
Louis-Dupont: I changed it in https://github.com/Deci-AI/super-gradients/commit/4f0ffa9124fd052d1aaa90e38677ac4268ba1ecf (I still have some comments to fix)
I chose to do it "per second" to have a constant metric that does not depend on how often we aggregate

src/super_gradients/common/environment/monitoring/gpu/gpu.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def count_gpus() -> int:
+    """Count how many GPUS NVDIA detects."""
+    return pynvml.nvmlDeviceGetCount()
+
+
+def get_device_memory_usage_percent(gpu_index: int) -> float:
+    """GPU memory utilization in percent of a given GPU."""
+    handle = pynvml.nvmlDeviceGetHandleByIndex(gpu_index)
```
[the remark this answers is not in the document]
Louis-Dupont: I added a cache to only call it once per index: c69e3e48d54855667d5104eecc2ed0da1e1ff4a0
But this doest work well if Nvidia makes the handlers stale after some time. I'm letting a training run for some time to check, and I'll also see if I can find some doc about this.

src/super_gradients/common/environment/monitoring/monitoring.py
Louis-Dupont: Added in 00c6dfdfbf3f9325fe515e07dc790ef86c720cfe

ofrimasad (approved):
LGTM. very nice work!

PR #514, opened 2022-11-26
Feature/380 function to upload models to platform

Louis-Dupont:
The need for a function that uploads to the platform model (not as part of the callback) was raised. 

The main reason for it is that uploads sometimes fail and therefor a stand-alone platform might help.

Need to mimic the function in the callback external to the callback.

---

QUESTION
Do we prefer this:
```
from ... import DeciClient
client = DeciClient()
client.upload_model(model=..., model_meta_data=..., optimization_request_form=...)
```

or this:
```
from ... import upload_model
upload_model(model=..., model_meta_data=..., optimization_request_form=...)
```


ofrimasad (approved):
LGTM after adding parameter types

PR #516, opened 2022-11-27
Remove imports from factory.__init__

Louis-Dupont:
This tends to create circular import loops because most of these factories themselves import modules from training, while some of the training modules import themselves factories.
Not importing it in the __init__ prevents all these factories to be instantiated together, creating these hard to debug import loops


shaydeci (approved):
LGTM

PR #517, opened 2022-11-27
Rename gpu_mode to multi_gpu and setup_gpu_mode to setup_device

Louis-Dupont:
This was raised by Dan Bochman.
There is confusion caused by mismatch of namings in config and code. Specifically, setup_gpu_mode function has  gpu_model argument, which has it’s counterpart called multi_gpu in YAML files. 

setup_gpu_mode was also renamed to setup_device, because in a following PR we will also include device type (cuda/cpu) so it won't just be about "gpu"


src/super_gradients/training/utils/distributed_training_utils.py
```diff
@@ -171,23 +171,32 @@ def wait_for_the_master(local_rank: int):
 
 
 def setup_gpu_mode(gpu_mode: MultiGPUMode = MultiGPUMode.OFF, num_gpus: int = None):
+    """[DEPRECATED in favor of setup_device] If required, launch ddp subprocesses.
```
shaydeci, line 174: Add @deprecated decorator please instead of modifying the method content..

shaydeci (changes requested):
Minor inline comment.

src/super_gradients/training/utils/distributed_training_utils.py
Louis-Dupont: I added it but dont ge the deprecated message. Do you think it's because super_gradients doesnt have a __version__ ?

shaydeci (dismissed):
LGTM

PR #519, opened 2022-11-28
Prevent system monitoring to write after calling .close()

Louis-Dupont:
This might lead to a situation where the TB is already closed and yet the system monitoring would try to write on TB


ofrimasad (approved):
LGTM

PR #523, opened 2022-11-28
Add DDP doc

Louis-Dupont:


README.md
```diff
@@ -291,22 +291,48 @@ Recipes support out of the box every model, metric or loss that is implemented i
...
-# Define the objects
+# Call the trainer
+Trainer(multi_gpu=MultiGPUMode.DISTRIBUTED_DATA_PARALLEL, expriment_name=...)
+
+# Everything you do below will run on 4 gpus
+
+...
+
+Trainer.train(...)
+
```
shaydeci, line 329: Maybe add a comment on how to run this (i.e that this can be simply ran as any other python .py file for torch>=XXX and using torch.distributed.launch .. otherwise)?

shaydeci (changes requested):
Single minor comment inline.

README.md
Louis-Dupont: I added : 
Please note that if you work with `torch<1.9.0` (deprecated), you will have to launch your training with either 
`torch.distributed.launch` or `torchrun`, in which case `nproc_per_node` will overwrite the value  set with `gpu_mode`:
```bash
python -m torch.distributed.launch --nproc_per_node=4 main.py
```
```bash
torchrun --nproc_per_node=4 main.py
```

shaydeci (approved):
LGTM

PR #525, opened 2022-11-28
Feature/sg 404 ssd reproduce

BloodAxe:
To be merged after https://github.com/Deci-AI/super-gradients/pull/524


src/super_gradients/recipes/coco2017_ssd_lite_mobilenet_v2.yaml
```diff
@@ -1,6 +1,6 @@
 # SSD MobileNetV2 Detection training on CoCo2017 Dataset:
 # Trained in 320x320 mAP@0.5@0.95 (COCO API, confidence 0.001, IoU threshold 0.6, test on 320x320 images) ~20.52
-# Checkpoint path: https://deci-pretrained-models.s3.amazonaws.com/ssd_lite_mobilenet_v2_coco_res320_new_coco_filtered_affine_scale_5_15_no_mosaic/ckpt_best.pth
+# Checkpoint path: https://deci-pretrained-models.s3.amazonaws.com/ssd_lite_mobilenet_v2/coco2017/2022-11-28/average_model.pth
 # (trained with stride_16_plus_big)
 # Hardware: 8 NVIDIA RTX 3090
```
shaydeci, line 5: Didn't we train with 4 devices ? Can you update the comments in the YAML file accordingly ?

tests/integration_tests/pretrained_models_test.py
```diff
@@ -102,7 +102,7 @@ def setUp(self) -> None:
         }
 
         self.coco_pretrained_maps = {
-            "ssd_lite_mobilenet_v2": 0.2052,
+            "ssd_lite_mobilenet_v2": 0.2027,
```
shaydeci, line 105: Did we check the test itself ?
We use here "crowd" targets, so mAP should be slightly higher then achieved in training.

shaydeci (changes requested):
2 minor comments inline.

tests/integration_tests/pretrained_models_test.py
BloodAxe: How are we supposed to run the integration tests? Found that locally I'm unable to run these due to absence of dataset dirs. SSH to tzag and run them from cmd line is the only way?

src/super_gradients/recipes/coco2017_ssd_lite_mobilenet_v2.yaml
BloodAxe: Updated, thanks

tests/integration_tests/pretrained_models_test.py
BloodAxe: Ran via cmd line. It actually reached 20.41 which is reflected in test & description

shaydeci (dismissed):
LGTM

shaydeci (approved):

PR #527, opened 2022-11-29
Feature/al 441 ptq detection

spsancti:
Adds Identities to shortcut connections


src/super_gradients/training/utils/quantization/export.py
```diff
@@ -34,7 +34,17 @@ def export_quantized_module_to_onnx(model: torch.nn.Module, onnx_filename: str,
...
+
+    if train:
+        training_mode = TrainingMode.TRAINING
+        model.train()
+    else:
+        training_mode = TrainingMode.EVAL
+        model.eval()
+
+    torch.onnx.export(
+        model, dummy_input, onnx_filename, verbose=False, opset_version=13, enable_onnx_checker=False, do_constant_folding=True, training=training_mode
```
ofrimasad, line 46: why do we need the training=training_mode? 
after we call prep_model_for_conversion the model can train anymore...

ofrimasad (commented):
looks very nice. 
one comment / question

src/super_gradients/training/utils/quantization/export.py
spsancti: For debug purposes we sometimes need an onnx file without optimizations
Moved prep model to eval mode only

ofrimasad (approved):
LGTM

PR #531, opened 2022-12-01
Feature/sg 431 check classes

BloodAxe:
Fix https://deci.atlassian.net/browse/SG-431


src/super_gradients/training/datasets/detection_datasets/coco_detection.py
```diff
@@ -46,14 +48,27 @@ def __init__(
         kwargs["all_classes_list"] = kwargs.get("all_classes_list") or COCO_DETECTION_CLASSES_LIST
         super().__init__(*args, **kwargs)
 
+        if len(self.original_classes) != len(self.all_classes_list):
```
ofrimasad, line 51: don't we want to check this before calling `super().__init__(...)`

ofrimasad (approved):
LGTM... just a minor question/comment

src/super_gradients/training/datasets/detection_datasets/coco_detection.py
BloodAxe: We don't, as we rely on value `self.original_classes` which is initialized from `_setup_data_source` which is called from base c'tor. 
The dataset initialization is quite puzzled and could be refactored a bit to make it less tangled though.

ofrimasad (approved):
LGTM

PR #532, opened 2022-12-01
SG-448- dataset_params and arch_params logging

shaydeci:
- Support for (optional) additional configs to be logged in training's sg_logger.
- Full recipe config logged when launching train_from_config
- Small bug fix in base_sg_logger (all configs were added under the same tag=hyper_parameters).


ofrimasad (approved):
LGTM

PR #534, opened 2022-12-03
SGLogger fix - add warning and multiprocess_safe system_monitoring

Louis-Dupont:
- Make system_monitoring @multiprocess_safe.
- Allow system_monitoring to be passed to all the loggers, but raise a warning and ignore it when using wandb/clearml


ofrimasad (approved):
LGTM

PR #535, opened 2022-12-04
Bug/sg 000 torch version fix

shaydeci:


ofrimasad (approved):
looks ok...
let's refactor this as soon as we can

PR #537, opened 2022-12-04
Quantization infra mods for different calibrators and learnable amax

spsancti:


src/super_gradients/training/utils/quantization/selective_quantization_utils.py
```diff
@@ -192,7 +204,12 @@ def _instantiate_quantized_from_float(self, float_module, metadata, preserve_sta
 
         # COPY STATE DICT IF NEEDED
         if preserve_state_dict:
-            q_instance.load_state_dict(float_module.state_dict(), strict=True)
+            if self.default_learn_amax:  # quant state dict will now have additional parameters for clip and strict loading will fail
+                logger.warning(
+                    "Instantiating quant module in non-strict mode leaving Clip parameters non-initilaized. Use QuantizationCalibrator to initialize them. "
```
ofrimasad, line 209: I am not sure I understand that message... 
I see you set strict to False, so why do you need this?

ofrimasad (commented):
one comment (question)

src/super_gradients/training/utils/quantization/selective_quantization_utils.py
spsancti: TensorQuantizer.clip will have two additional parameters registered: clip_min and clip_max. They are learnable, so the state dict will include them, too. As they are not present in the original module's state_dict strict=True will fail with missing keys

spsancti: QuantizationCalibrator now has `module.init_learn_amax()` which will load computed amax to clip module as initialization value.

ofrimasad (approved):
LGTM

PR #538, opened 2022-12-04
fix  pplite-seg prep conversion

lkdci:
prevent prep_for_conversion from crashing when called twice and global pooling modules were already converted to AveragePool modules.

fixed raising this error:
```
AttributeError: 'AvgPool2d' object has no attribute 'output_size' when I try to export to model.prep_model_for_conversion(input_size=input_size)
```


ofrimasad (approved):
LGTM

PR #541, opened 2022-12-05
Add version_base to hydra to remove warning

Louis-Dupont:
- Add version_base to hydra to remove warning
- Add Crashtip to help user in case he is using wrong version and gets an error
- Add Crashtip to CI test


shaydeci (approved):
LGTM

PR #543, opened 2022-12-05
replace head for UNet module

lkdci:
Implemented the replace head for UNetCustom module.
Extracted duplicated code of SegmentationHead module to common.py file


shaydeci (approved):
LGTM

PR #544, opened 2022-12-05
Feature/sg 456 centralize ddp setup

Louis-Dupont:
Motivation:
- Setup DDP in a single place
- Setup DDP sooner to reduce risk of errors (when using an object that requires DDP before DDP is sertup)
- Improving the logic and clarity of how DDP is setup


src/super_gradients/common/environment/environment_config.py
```diff
@@ -1,15 +1,38 @@
...
+    if os.getenv("LOCAL_RANK") is not None:
+        return int(os.getenv("LOCAL_RANK"))
+    else:
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--local_rank", default=-1)
+        args, _ = parser.parse_known_args()
+        return args.local_rank
+
+
+_DDP_REQUESTED_RANK = _get_requested_rank()
```
BloodAxe, line 26: Maybe it make sense to call pop_arg("local_rank") in _get_requested_rank() instead of init_trainer?
It seems we call _get_requested_rank() only once and set _DDP_REQUESTED_RANK. We do this at import level, which would guarantee pop_arg("local_rank") to be called before hydra is about to start complaining on presence of `--local_rank`

src/super_gradients/training/utils/distributed_training_utils.py
```diff
@@ -171,30 +192,128 @@ def wait_for_the_master(local_rank: int):
...
+    if local_rank > 0:
+        mute_current_process()
+
+    logger.info("Distributed training starting...")
+    if not torch.distributed.is_initialized():
+        backend = "gloo" if os.name == "nt" else "nccl"
+        torch.distributed.init_process_group(backend=backend, init_method="env://")
+    torch.cuda.set_device(local_rank)
+
+    assert device_config.requested_rank == get_local_rank()
```
BloodAxe, line 306: It is recommended to use assert only inside unit tests and use exceptions for runtime checks elsewhere.  assert can be tuned off by interpreter settings while runtime checks can't.

src/super_gradients/common/environment/environment_config.py
```diff
@@ -1,15 +1,38 @@
...
+
 try:
     PKG_CHECKPOINTS_DIR = pkg_resources.resource_filename("checkpoints", "")
 except Exception:
     os.makedirs(os.path.join(os.getcwd(), "checkpoints"), exist_ok=True)
     PKG_CHECKPOINTS_DIR = os.path.join(os.getcwd(), "checkpoints")
 
 
-DDP_LOCAL_RANK = int(os.getenv("LOCAL_RANK", default=-1))
+def _get_requested_rank():
```
BloodAxe, line 16: May I suggest to rename the function and corespodnding variable name from `requested_rank` to `assigned_rank`? I feel it would have better clarity on what this functon returns.

src/super_gradients/common/environment/environment_config.py
Louis-Dupont: Mmhh the thing I'm afraid with poping on import is that a user might expect to find `--local_rank` when using `torch.distributed.launch` and if we pop it behind his back it might be very confusing for him. The advantage of using init_trainer is that we can document the function so that the user knows what happens when calling it, it wouldnt be hidden.

Maybe what we can do is as you say pop on import, and if we pop it add a log.info to let the user know.
In most cases we would'nt even log anything, because when launching with python, torchrun or torch.distributed.run the local rank is passed as env variable (LOCAL_RANK) and not as --local_rank arg

src/super_gradients/common/environment/env_helpers.py
```diff
@@ -142,7 +132,9 @@ def is_main_process():
     """
     if not is_distributed():  # If no DDP, or DDP launching process
         return True
-    elif is_rank_0() and not is_launched_using_sg():  # If DDP launched using torch.distributed.launch or torchrun, we need to run the check on rank 0
+    elif (
+        device_config.requested_rank == 0 and not is_launched_using_sg()
```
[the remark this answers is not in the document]
Louis-Dupont: It's a bit tricky. Basically:
- When launching with SG directly (`python train_from_recipe.py` for instance), we have a launcher job that will call `torch.distributed.launcher.api.elastic_launch` from within `setup_gpu`, which will launch all the DDP subprocesses. This launcher is our "main" process.
- When launching with torchrun/python -m torch.distributed.launch/run, the launcher is not based on SG code so we don't have any control over it (it will not run sanity_check, and we won't be able to save its logs). In that case, our "main" process is the rank0

shaydeci (changes requested):
Nice, 1 short comment inline.
Please also run the gpu tests before we merge, so we won't have surprises on release.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -117,9 +119,20 @@ def __init__(self, experiment_name: str, device: str = None, multi_gpu: Union[Mu
                                                 pkg_resources.resource_filename('checkpoints', "") exists and will be used.
 
         """
+
+        # This should later me removed
+        if device is not None or multi_gpu is not None:
```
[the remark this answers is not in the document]
Louis-Dupont: Humm, I chose to raise an error (raise KeyError below) to make sure that people don't miss the deprecated warning, since this is a major change.

shaydeci (approved):
LGTM

shaydeci (approved):

PR #546, opened 2022-12-06
Features/sg 409 check all params used

BloodAxe:


shaydeci (changes requested):
Looks really good!

My only concern, is for the DictConfigs we get back from hydra_instantiate, which we use when we launch recipe: in the unit tests we use OmegaConf.create() to mimic this scenario. Can you confirm their contents would be similar?\
This way or the other, I think it'd be wise to instantiate one of our yaml files (our create one for the sake of testing) and test this out.

BloodAxe:
@shaydeci I've added a test that checks instantiation of a model from YAML file (One of just a few that can be used as is without bells & whistles) which pass nicely. Checking all YAMLs in this way is a subject for another ticket: https://deci.atlassian.net/browse/SG-498

```
    def test_resnet18_cifar_arch_params(self):
        arch_params = get_arch_params("resnet18_cifar_arch_params")
        architecture_cls, arch_params, pretrained_weights_path, is_remote = get_architecture("resnet18", HpmStruct(**arch_params))

        with raise_if_unused_params(arch_params) as tracked_arch_params:
            _ = architecture_cls(arch_params=tracked_arch_params)

        with self.assertRaisesRegex(UnusedConfigParamException, "Detected unused parameters in configuration object that were not consumed by caller"):
            arch_params.override(me_is_not_used=True)
            with raise_if_unused_params(arch_params) as tracked_arch_params:
                _ = architecture_cls(arch_params=tracked_arch_params)

```

shaydeci (approved):
LGTM

PR #547, opened 2022-12-06
Feature/sg 458 support for any user dataloader using dataset registry

shaydeci:
- Added support for launching recipes when only user has only the dataset class (i.e torch.Dataset).
- Registry + factory added for datasets.
- Example: launching MNIST training using the Mnist dataset alone (no DataLoader implementation).


ofrimasad (approved):
LGTM

PR #548, opened 2022-12-07
Split and rename the modules from super_gradients.common.environment

Louis-Dupont:
Motivation:
The current separation creates many circular loops when I try to use them.

Solution proposed:
split the package in more modules, with a clear purpose


ofrimasad (approved):
LGTM

PR #549, opened 2022-12-07
Feature/infra 1481 call integration tests

ranrubin:
This will trigger the CircleCi pipeline tests that are being developed by @Louis-Dupont


.github/workflows/integration_tests_rc.yaml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Calling CircleCI job
+        shell: bash
+        run: |
+          curl --request POST \
+            --url https://circleci.com/api/v2/project/gh/Deci-AI/algo-integration-tests/pipeline \
+            --header 'Circle-Token: ${{ secrets.CIRCLE_CI_TOKEN }}' \
+            --header 'content-type: application/json' \
+            --data '{"parameters":{"sg_workflow_sg_version":"${{github.ref_name}}", "sg_workflow_install_from_scratch":false, "sg_workflow_run":true, "sg_workflow_rc":true }}'
```
shaydeci, line 18: Why don't we install from scratch for RC ?

shaydeci (changes requested):
1 question inline...

.github/workflows/integration_tests_rc.yaml
ranrubin: This is what @Louis-Dupont  wanted. 
Installing from scratch can take 15 minutes, and if it happens a few times a day that could be inconvenient.

Feel free to change it to true :)

shaydeci (approved):
LGTM

PR #552, opened 2022-12-07
add a __version__ variable

ofrimasad:


tests/verify_version.py
```diff
@@ -0,0 +1,12 @@
+import sys
+
+import super_gradients
+
+if __name__ == "__main__":
+
+    ci_version = sys.argv[1]
+    if ci_version == super_gradients.__version__:
```
shaydeci, line 8: This means, that before release we must edit the src/super_gradients/__init__.py manually, intended ?

shaydeci (changes requested):

shaydeci (approved):

PR #554, opened 2022-12-08
Make ConvBNReLU a subclass of ConvBNAct to keep backward compatibililty

BloodAxe:


ofrimasad (changes requested):
Why is that code in __init__.py ?
Can we move it to a separate code file?

BloodAxe:
> Why is that code in init.py ?
Good suggestions, thanks. It was added there while ConvBnRelu was a function, but since it's a subclass now it make sense to move to separate file. Done

ofrimasad (approved):
LGTM

PR #556, opened 2022-12-08
improve error msg, and improve test

Louis-Dupont:
Problem:
- self.data_dir and self.images_sub_directory combined in a very weird and bad way, potentially causing errors
- On top, this error message was not very clear

What I did:
- Fixing all the above
- Adding tempfile + download to allow the `integration_tests/detection_dataset_test` to work on any device


ofrimasad (approved):
LGTM

PR #558, opened 2022-12-08
Improve evaluate_from_recipe usability

Louis-Dupont:
Change introduced:
Now instead of evaluating training.ckpt_name, we first check if the checkpoint_path is provided. Only if it is not, we use training.ckpt_name.


ofrimasad (approved):
LGTM

PR #559, opened 2022-12-11
Feature/sg 468 detection transform support for any number of channels

shaydeci:
- Modified Mixup and HSV to work with any number of channels.
- Warning is displayed when applying HSV with additional channels.
- Added an integration test, that checks the support of our coco base recipe transforms (all detection transforms) compatibility with 6 channels.


src/super_gradients/training/transforms/transforms.py
```diff
@@ -567,7 +582,7 @@ def __call__(self, sample: dict):
             jit_factor = random.uniform(*self.mixup_scale)
 
             if len(img.shape) == 3:
-                cp_img = np.ones((self.input_dim[0], self.input_dim[1], 3), dtype=np.uint8) * 114
+                cp_img = np.ones((self.input_dim[0], self.input_dim[1], img.shape[2]), dtype=np.uint8) * 114
```
ofrimasad, line 585: if you added border_value, why is 114 hard codded here?

ofrimasad: what if if len(img.shape) == 2? (grayscale) I see you are checking that in line 607 but not here

src/super_gradients/training/transforms/transforms.py
```diff
@@ -699,8 +728,18 @@ def __init__(self, prob: float, hgain: float = 0.5, sgain: float = 0.5, vgain: f
         self.sgain = sgain
         self.vgain = vgain
         self.bgr_channels = bgr_channels
+        self._additional_channels_warned = False
 
     def __call__(self, sample: dict) -> dict:
+        if sample["image"].shape[2] > 3 and not self._additional_channels_warned:
```
ofrimasad, line 734: what if if len(img.shape) == 2? (grayscale) this will crash

ofrimasad (commented):
a few comments inline

src/super_gradients/training/transforms/transforms.py
shaydeci: fixed- raised exception when channels < 3 in latest.

shaydeci: Also added border value to all transforms where 114 was hard coded.

ofrimasad (approved):
LGTM

PR #560, opened 2022-12-11
Feature/sg 518 update evaluate from recipe to add output

Louis-Dupont:
Motivation:
I need to get the results of evaluate_from_recipe and currently we don't return anything.

Solution proposed:
return the same objects as train_from_recipe (i.e. model and result tuple)


shaydeci (approved):
LGTM

PR #561, opened 2022-12-12
Feature/sg 193 extend output formator

Louis-Dupont:
### Goals
Have a class that works like DetectionOutputAdapter but that supports np.arrays and use it on transforms.


src/super_gradients/training/utils/output_adapters/detection_adapter.py
```diff
@@ -67,6 +100,153 @@ def forward(self, x: Tensor) -> Tensor:
         return x
 
 
+def get_last_dim(tensor: Union[np.ndarray, Tensor]) -> int:
+    if isinstance(tensor, Tensor):
```
BloodAxe, line 104: Note torch tensor also supports .shape semantics

src/super_gradients/training/transforms/transforms.py
```diff
@@ -764,10 +770,27 @@ class DetectionTargetsFormatTransform(DetectionTransform):
         max_targets: int: max objects in single image, padding target to this size.
     """
 
+    _XYXY_LABEL = ConcatenatedTensorFormat(
```
shaydeci: Whats this doing here ?
If I understand correctly, you are doing this so that you can pass ConcatenatedTensorFormat as a default? Why not using the defaults (by import, by extending resolve_param so that it'd take in account defaults...) plenty of solutions here.

src/super_gradients/training/utils/output_adapters/detection_adapter.py
```diff
@@ -122,6 +165,8 @@ class DetectionOutputAdapter(nn.Module):
     >>>
     >>> yolox = nn.Sequential(yolox, output_adapter)
     >>>
+
+    Note: DetectionOutputAdapter has a common logic but different implementation to DetectionFormatAdapter in order to be serializable with TorchScript.
```
shaydeci, line 169: It took me quite a bit to understand the difference in use of those both.
Lets separate them into different pages and possibly rename one of them to avoid confusion...

src/super_gradients/training/transforms/transforms.py
```diff
@@ -777,66 +800,30 @@ def __init__(
...
-        label_first_in_input = self.input_format.value.split("_")[0] == "LABEL"
-        label_first_in_output = self.output_format.value.split("_")[0] == "LABEL"
-
-        input_xyxy_format = "XYXY" in self.input_format.value
-        output_xyxy_format = "XYXY" in self.output_format.value
-        convert2xyxy = not input_xyxy_format and output_xyxy_format
-        convert2cxcy = input_xyxy_format and not output_xyxy_format
-
-        image, targets, crowd_targets = sample["image"], sample["target"], sample.get("crowd_target")
+        self.convert_format = DetectionFormatAdapter(input_format=self.input_format, output_format=self.output_format, image_shape=image_shape)
```
shaydeci, line 789: Name is misleading (even if it looks nice in call()). Please rename.

tests/unit_tests/detection_targets_format_transform_test.py
```diff
@@ -12,110 +19,115 @@ def setUp(self) -> None:
...
-                                                    input_format=DetectionTargetsFormat.LABEL_XYXY,
-                                                    output_format=DetectionTargetsFormat.LABEL_NORMALIZED_XYXY)
         sample = {"image": self.image, "target": input}
+
+        transform = DetectionTargetsFormatTransform(
+            image_shape=self.image.shape[1:], max_targets=1, input_format=LABEL_XYXY, output_format=LABEL_NORMALIZED_XYXY
+        )
         t_output = transform(sample)["target"]
-        self.assertTrue(np.array_equal(output, t_output))
+        self.assertTrue(np.allclose(output, t_output, atol=1e-6))
```
shaydeci, line 38: Why the added tolerance? Results should be exact here...
This goes also for the rest (except, maybe when normalising, but then again it never failed yet because of that).

src/super_gradients/training/utils/output_adapters/formats_utils.py
```diff
@@ -0,0 +1,50 @@
+from typing import List, Callable, Union
+
+import numpy as np
+from torch import Tensor
+
+from super_gradients.training.utils.output_adapters.formats import ConcatenatedTensorFormat
+
```
shaydeci: If these aren't "private" _  methods please add PEP8 docs.

src/super_gradients/training/transforms/transforms.py
```diff
@@ -777,66 +800,30 @@ def __init__(
...
+            padded_targets[range(len(targets))[: self.max_targets]] = targets[: self.max_targets]
             padded_targets = np.ascontiguousarray(padded_targets, dtype=np.float32)
 
             return padded_targets
 
-        sample["target"] = _format_target(targets)
-        if crowd_targets is not None:
-            sample["crowd_target"] = _format_target(crowd_targets)
+        sample["target"] = _format_target(sample["target"])
+        if "crowd_target" in sample.keys():
```
shaydeci, line 811: Shouldn't crowd targets be handled differently now that we can extend detection targets?
Lets at least put a TODO to refactor crowd targets.

shaydeci (changes requested):
Some comments inline.
There are plenty of new classes here, their unique usage isn't very clear.
Please try and simplify some of them (like you suggested) with inheritance, and as a last resort multiple inheritance.
Another point to take in account is testing- the fact that the detection targets transforms test use your newly added components, does not imply that something might break in my implementation leaving yours unharmed. Its always best that we test newly added modules.

Nice work!

src/super_gradients/training/transforms/transforms.py
Louis-Dupont: Yeah this was actually removed and in the latest commit the default value is used

tests/unit_tests/detection_targets_format_transform_test.py
Louis-Dupont: There are some precision issues when normalizing/denormalizing:
image

src/super_gradients/training/transforms/transforms.py
Louis-Dupont: We could, but I think this would be an even bigger refacto, because it implies that the format of target/crowd_target is different either for the input, and/or for the output. In both cases, it means that the previous and/or following transforms should support target/crowrd_target with a different format, which is not the case.
But yeah I will write a ticket to think about this problematic

src/super_gradients/training/utils/output_adapters/detection_adapter.py
Louis-Dupont: I split it to 2 different modules:
format_converter.py (with class: ConcatenatedTensorFormatConverter)
output_format_adapter.py (with class: DetectionOutputFormatAdapter)

src/super_gradients/training/utils/output_adapters/formats_utils.py
Louis-Dupont: added

src/super_gradients/training/transforms/transforms.py
Louis-Dupont: Renamed like that: https://github.com/Deci-AI/super-gradients/blob/0f65b88cdfba6ceb791bb0927ea28d87859e855d/src/super_gradients/training/transforms/transforms.py#L787

shaydeci (approved):
LGTM

src/super_gradients/training/transforms/transforms.py
```diff
@@ -756,88 +761,59 @@ class DetectionTargetsFormatTransform(DetectionTransform):
...
+    @resolve_param("input_format", ConcatenatedTensorFormatFactory())
+    @resolve_param("output_format", ConcatenatedTensorFormatFactory())
     def __init__(
         self,
-        input_format: DetectionTargetsFormat = DetectionTargetsFormat.XYXY_LABEL,
-        output_format: DetectionTargetsFormat = DetectionTargetsFormat.LABEL_CXCYWH,
+        image_shape,
+        input_format: ConcatenatedTensorFormat = XYXY_LABEL,
+        output_format: ConcatenatedTensorFormat = LABEL_CXCYWH,
         min_bbox_edge_size: float = 1,
```
BloodAxe, line 779: I feed this parameter needs a documentation

src/super_gradients/training/transforms/transforms.py
```diff
@@ -756,88 +761,59 @@ class DetectionTargetsFormatTransform(DetectionTransform):
...
 
-            return padded_targets
+        targets = filter_on_bboxes(fn=_is_big_enough, tensor=targets, tensor_format=self.output_format)
+        return targets
 
-        sample["target"] = _format_target(targets)
-        if crowd_targets is not None:
-            sample["crowd_target"] = _format_target(crowd_targets)
-        return sample
+    def pad_targets(self, targets: np.ndarray) -> np.ndarray:
```
BloodAxe, line 811: This padding of targets to `max_targets` should be a part of collate function not a target transform, IMO.
Since this padding is specific to YoloX. 
Yes, this is not directly related to this PR and should not prevent from merging it, but worth keeping in mind. CC @ofrimasad

src/super_gradients/training/utils/tensor_formats/bbox_formats/cxcywh.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 import torch
 
-from super_gradients.training.utils.bbox_formats.bbox_format import (
+from super_gradients.training.utils.tensor_formats.bbox_formats.bbox_format import (
```
BloodAxe, line 7: What was the reasoning in moving 
`super_gradients.training.utils.bbox_formats.bbox_format` to from `super_gradients.training.utils.tensor_formats.bbox_formats` ?

src/super_gradients/training/utils/tensor_formats/default_formats.py
```diff
@@ -0,0 +1,103 @@
...
+from super_gradients.training.utils.tensor_formats.bbox_formats import (
+    XYXYCoordinateFormat,
+    XYWHCoordinateFormat,
+    CXCYWHCoordinateFormat,
+    NormalizedXYXYCoordinateFormat,
+    NormalizedXYWHCoordinateFormat,
+    NormalizedCXCYWHCoordinateFormat,
+)
+
+XYXY_LABEL = ConcatenatedTensorFormat(
```
BloodAxe, line 12: So what I don't like about current factory is that one has to make changes in three places to add new format (or optimizer, or model or loss):

1) Add new line to ConcatenatedTensorFormats: `MY_NEW_XYXY_LABEL_FORMAT = "MY_NEW_XYXY_LABEL_FORMAT"`
2) Define the format itself: `MY_NEW_XYXY_LABEL_FORMAT = ConcatenatedTensorFormat(...)`
3) Register it in `DEFAULT_CONCATENATED_TENSOR_FORMATS`

This is unnecessary self-repeating process which we can avoid by using decorators or function calls:

```
register(BBoxFormatFactory, "XYWH_LABEL", ConcatenatedTensorFormat(
    layout=(
        BoundingBoxesTensorSliceItem(name="bboxes", format=XYWHCoordinateFormat()),
        TensorSliceItem(length=1, name="labels"),
    )
))
```
By doing so we ditch DEFAULT_CONCATENATED_TENSOR_FORMATS and ConcatenatedTensorFormats classes and fill our factories in one place

src/super_gradients/training/utils/tensor_formats/bbox_formats/cxcywh.py
Louis-Dupont: I though that it made sense since they are all related to formats, and grouping it together would increase module cohesion. But to be honest I am not fully sure that moving bbox_formats in this module is good, because its also a standalone module for bbox format conversion.
So, should we just take bbox_formats out  of tensor_formats like it was?

src/super_gradients/training/utils/tensor_formats/default_formats.py
Louis-Dupont: I agree that a lot of unnecessary self-repeating, but it changes the current logic done for optimizer, model, loss, ...
If we change this, I think we should change it for all the factories to keep a homogenous approach.

Your proposition in the case of classes, like a model for instance, would be working with a decorator, the same way we ask people with a custom models to do, right?

@register_model
class Resnet(...):
    ...

Louis-Dupont: Your proposition for classes would be working with a decorator, the same way we ask people with a custom models to do, right?

```
@register_model
class Resnet(...):
    ...
```

src/super_gradients/training/transforms/transforms.py
Louis-Dupont: IMO `filter_small_bboxes` should also be moved to another transform (who could expect DetectionTargetsFormatTransform to also filter boxes?).
I didnt do it here to not overload more the PR, but I created a ticket for that

src/super_gradients/training/utils/tensor_formats/bbox_formats/cxcywh.py
BloodAxe: In my opinion - yes. It is just unexpected to find `bbox_formats` under `tensor_formats`. 
Maybe rename tensor_formats to data_formats. Then import path would be more clear: utils -> data_formats -> bbox_formats.
Side note: I'd actually recommend to move it to `datasets` from `utils`:
`super_gradients.datasets.data_formats.bbox_formats` ❤️ ❤️ ❤️

src/super_gradients/training/utils/tensor_formats/default_formats.py
BloodAxe: Whether it is possible - yes. For cases where we cannot do this (as with example of tensor format layout) - through `register_WHATEVER(...)` call.
Agree this is a huge change and def. not part of this PR

src/super_gradients/training/utils/tensor_formats/bbox_formats/cxcywh.py
BloodAxe: @Louis-Dupont overall, LGTM. Feel free to merge without this change if not feeling adventurous enough to include proposed change to this PR :)

shaydeci (approved):

PR #562, opened 2022-12-12
Added QARepVGGBlock

spsancti:
QARepVGG (S4) block from 'Make RepVGG Greater Again: A Quantization-aware Approach' (https://arxiv.org/pdf/2212.01593.pdf)
It consists of three branches:

3x3: a branch of a 3x3 Convolution + BatchNorm
1x1: a branch of a 1x1 Convolution with bias
identity: a Residual branch
BatchNorm is applied after summation of all three branches.

In contrast to our implementation of RepVGGBlock, SE is applied AFTER NONLINEARITY in order to fuse Conv+Act in inference frameworks.

This module converts to Conv+Act in a PTQ-friendly way by calling QARepVGGBlock.fuse_block_residual_branches().
Has the same API as RepVGGBlock and is designed to be a plug-and-play replacement but is not compatible parameter-wise.
Has less trainable parameters than RepVGGBlock because it has only 2 BatchNorms instead of 3.


src/super_gradients/modules/qarepvgg_block.py
```diff
@@ -0,0 +1,305 @@
...
+            in_channels=in_channels,
+            out_channels=out_channels,
+            kernel_size=1,
+            stride=stride,
+            padding=0,
+            groups=groups,
+            bias=use_1x1_bias,
+        )
+
+        if use_residual_connection and out_channels == in_channels and stride == 1:
```
ofrimasad, line 126: please assert that out_channels == in_channels if use_residual_connection. 
this is a silent fail. if you use `use_residual_connection=True` and `out_channels != in_channels` or `stride != 1`
this will work unexpectedly

src/super_gradients/modules/qarepvgg_block.py
```diff
@@ -0,0 +1,305 @@
...
+            self.__delattr__("id_tensor")
+
+        self.qat_mode = True
+        self.deploy_mode = False
+
+    def fuse_block_residual_branches(self):
+        self.prepare_for_deploy()
+
+
+if __name__ == "__main__":
```
ofrimasad, line 284: please remove this local test.

ofrimasad (commented):
This PR breaks the logic we have built for quantization.
instead of having a block and registering a quantized block, we define here a block that knows how to be quantized. I don't like this and I am not sure exactly how that will work with the rest of the quantization code.
Instead, lets define a block without the quantization part and (no `if qat_mode`) and override some logics in a second block that is registered as his Quantized version

ofrimasad (changes requested):
waiting for fixes

src/super_gradients/modules/qarepvgg_block.py
```diff
@@ -0,0 +1,297 @@
...
+
+        if self.partially_fused:
+            return self.se(self.nonlinearity(self.post_bn(self.rbr_reparam(inputs))))
+
+        if self.identity is None:
+            id_out = 0.0
+        else:
+            id_out = self.identity(inputs)
+
+        return self.se(self.nonlinearity(self.post_bn(self.branch_3x3(inputs) + self.alpha * self.branch_1x1(inputs) + id_out)))
```
ofrimasad, line 183: can please break it into a few lines? 
later when it crashes for some reason, it will be very hard to know which one f these 5 layers caused the crash

ofrimasad (commented):
only one minor comment

ofrimasad (approved):
LGTM

PR #564, opened 2022-12-12
fix F1 Precision Recall to represent all the defined rage and not onl…

ofrimasad:
…y the minimum IoU in the range


shaydeci (approved):
LGTM

PR #567, opened 2022-12-13
Import AutoLoggerConfig and ConsoleSink separatly

Louis-Dupont:
Motivation:
- get_logger requires AutoLoggerConfig
- AutoLoggerConfig and ConsoleSink are imported together in the auto_logging package __init__
- ConsoleSink includes some DDP dependencies, which themselves require other imports
This often creates circular import loops that are hard to fix.

Solution:
Not importing ConsoleSink when importing AutoLoggerConfig


shaydeci (approved):
LGTM

PR #569, opened 2022-12-13
Fix Multigpu.OFF factory

Louis-Dupont:
PB:
hydra resolves "OFF" as a boolean, and we don't process booleans in the typefactory.


shaydeci (approved):
LGTM

PR #579, opened 2022-12-18
Feature/sg 000 container and integration tests

ofrimasad:


shaydeci (approved):
LGTM

.circleci/config.yml
```diff
@@ -227,6 +239,7 @@ workflows:
           requires:
             - deci-common/persist_version_info
             - deci-common/codeartifact_login
+      - integration_test
```
ranrubin, line 242: Don't we want it to run only after the build? otherwise, integration of what are we testing?

scripts/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM nvidia/cuda:11.3.1-devel-ubuntu20.04
+
+LABEL maintainer "DECI.AI <services@deci.ai>"
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN mkdir /SG
+
+RUN apt-get update && apt-get install -y python3-pip python-is-python3 pip libgl1 libglib2.0-0 git python3-distutils python3-typing-extensions \
+    && rm -rf /var/lib/apt/lists/*
+RUN pip install stringcase git+https://github.com/Deci-AI/super-gradients.git@stable --no-cache-dir && pip install torch==1.11.0+cu113 torchvision==0.12.0+cu113 --extra-index-url https://download.pytorch.org/whl/cu113 --no-cache-dir
```
ranrubin, line 10: pip install stringcase? 
also, why use stable tag?

.circleci/config.yml
ranrubin: If the integration test did not pass, do we still want to release?

ranrubin (changes requested):

.circleci/config.yml
ranrubin: @strelok899

scripts/Dockerfile
ranrubin: @strelok899

.circleci/config.yml
strelok899: it is after the build .
maybe you looking on old code

ranrubin: I fixed it

PR #587, opened 2022-12-28
Feature/sg 521 gpu tests

shaydeci:
Added-

- Recipe tests suite: currently contains tests running on saved checkpoints from the experiments ran in line 206 .circleci/config.yml . Main purpose is to test training generally works with DDP and the ran recipes metrics aren't stuck.
- Small fix in SgTrainer (batch size warning for drop last- when using BatchSampler it failed since in that case there's no "batch_size" attribute).
- Exit code for DDP exit set to 0 (was message before, which led to CI failing, and is not an informative exit code).
- Numpy requirement which led to failure.


.circleci/config.yml
```diff
@@ -180,6 +179,40 @@ jobs:
...
+      - checkout
+      - run:
+          name: install requirements and run recipe tests
+          command: |
+            << parameters.sg_new_env_python_version >> -m venv << parameters.sg_new_env_name >>
+            source << parameters.sg_new_env_name >>/bin/activate
+            python3.8 -m pip install --upgrade setuptools pip wheel
+            python3.8 -m pip install -r requirements.txt
+            python3.8 -m pip install git+https://github.com/Deci-AI/super-gradients.git@${CIRCLE_BRANCH}
+            python3.8 -m pip install torch==1.12.0+cu116 torchvision==0.13.0+cu116 torchaudio==0.12.0 --extra-index-url https://download.pytorch.org/whl/cu116
```
ofrimasad, line 205: hardcoded torch version?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -954,7 +954,13 @@ def forward(self, inputs, targets):
             training_params = dict()
         self.train_loader = train_loader or self.train_loader
         self.valid_loader = valid_loader or self.valid_loader
-        if len(self.train_loader.dataset) % self.train_loader.batch_size != 0 and not self.train_loader.drop_last:
+
+        if hasattr(self.train_loader, "batch_sampler") and self.train_loader.batch_sampler is not None:
```
ofrimasad, line 958: how is this change related?

ofrimasad (commented):
a few comments inline

.circleci/config.yml
shaydeci: gotta install some version (properly, with cuXXX etc). Might as well be the latest one we support.
Other suggestion ?

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: thats my and point here https://github.com/Deci-AI/super-gradients/pull/587#issue-1512612668 which I found only during running yolox recipe in the test (batch sampler is used -> batch_size=None, which lead this line to crash).
Could've taken it out and merge it before but it seemed such a small thing.

ofrimasad (approved):
LGTM

PR #588, opened 2022-12-28
fix ignore index for DiceCEEdgeLoss

lkdci:
Torch native cpp loss classes except only integer as ignore_index, the default is -100 for not using ignore.
While python based implementation use `None` as explicit indication for no using an ignore_index, see DiceLoss and torchmetrics.JaccardIndex class signatures.
This PR prevent passing an integer to `DiceLoss` that crash in the `to_one_hot` function due to an extra unnecessary ignore class, as follows:
```
  File "/home/lior.kadoch/PycharmProjects/super-gradients/src/super_gradients/training/losses/dice_loss.py", line 28, in _calc_numerator_denominator
    numerator = labels_one_hot * predict
RuntimeError: The size of tensor a (22) must match the size of tensor b (21) at non-singleton dimension 1
```


ofrimasad (approved):
LGTM

PR #589, opened 2022-12-28
add a decorator that saves the code to file

ofrimasad:


src/super_gradients/common/decorators/code_save_decorator.py
```diff
@@ -0,0 +1,29 @@
+import inspect
+
+import os
+
+saved_codes = {}
+
+
```
shaydeci, line 7: Can we add PEP8 docstrings to the below methods ?
Also, would be nice to see inside their docs an example usage with input/output as it might be a bit vague.

src/super_gradients/common/decorators/code_save_decorator.py
```diff
@@ -0,0 +1,29 @@
+import inspect
+
+import os
+
+saved_codes = {}
```
shaydeci, line 5: Not sure this is the best idea...

If we treat saved_codes as a global var then consecutive runs will result in weird loggings...

shaydeci (commented):
2 comments inline.

Apart from that, we might want to reconsider the design of treating the saved_code as a global variable. On one hand, this means that all of the code will be saved (even the one after calling train() which is desired in case, for example when we have some convert script), but there's another concern when launching consecutive trainings (not really a concern, but if we launch Resnet18 recipe then Resnet50 recipe- unnecessary code would be logged). Don't we want just the places we imported (in the sane fashion we do for register?)
Also, would be nice to have a simple unit test, maybe just a sanity that prints some code to see nothing is broken (if it is, it also mean wed break sg_logger and therefore training).

shaydeci (approved):
LGTM

PR #599, opened 2023-01-05
Feature/quantizable stdc

spsancti:


src/super_gradients/modules/quantization/quantized_stdc_blocks.py
```diff
@@ -0,0 +1,116 @@
...
+        for conv, quant in zip(self.conv_list[1:], self.quant_list):
+            x = conv(x)
+            out_list.append(quant(x))
+
+        out = torch.cat(out_list, dim=1)
+        return out
+
+
+@register_quantized_module(float_source=AttentionRefinementModule, action=QuantizedMetadata.ReplacementAction.REPLACE_AND_RECURE)
+class QuantAttentionRefinementModule(SGQuantMixin, AttentionRefinementModule):
```
BloodAxe, line 50: Some types has prefix Quantized (QuantizedMetadata) and some - Quant (Like current class).
I think for sake of consistency it would be great to name them all using same convention to avoid confusion why some starts with Quantized and some with Quant.

src/super_gradients/training/models/segmentation_models/stdc.py
```diff
@@ -35,13 +35,17 @@ def __init__(self, in_channels: int, out_channels: int, steps: int, stdc_downsam
         """
         super().__init__()
         assert steps in [2, 3, 4], f"only 2, 3, 4 steps number are supported, found: {steps}"
```
BloodAxe, line 37: This is not directly related to changes of this PR, but since you're already touching the c'tor - please replace assert statement with if / raise statement.

**Why: Assertions can be toggled on or off at will.**

https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/

> So never use an assert when another method of capturing and raising an error is possible. Always write your code with the assumption that the assertions could disappear at any minute. And if you’re unsure, you can run your program with python -O or by setting PYTHONOPTIMIZE=1 to test if it still works.

TLDR: assert in unit tests, if/raise in library code

src/super_gradients/training/models/segmentation_models/stdc.py
```diff
@@ -379,6 +379,10 @@ def forward(self, x):
             return feat8, feat16_up, feat16, feat32
         return feat8, feat16_up
 
+    def prep_for_conversion(self, input_size):
+        context_embedding_up_size = (input_size[-2] // 32, input_size[-1] // 32)
```
BloodAxe, line 383: Do we need to change that `input_size[-2] % 32 == 0`?

src/super_gradients/modules/quantization/quantized_stdc_blocks.py
spsancti: QuantizedMetadata is a separate entity which is a part of the quantizer infra
Quant* are custom quantizeable counterparts for regular module

BloodAxe (approved):

BloodAxe:
LGTM

PR #603, opened 2023-01-08
Support 1.13

Louis-Dupont:


shaydeci (approved):
LGTM

requirements.txt
```diff
@@ -1,4 +1,4 @@
-torch>=1.9.0,<=1.12
+torch>=1.9.0,<=1.13.1
```
ofrimasad, line 1: can we set it to 1.13.*?

ofrimasad (commented):
my comment inline. can we make

ofrimasad (changes requested):
can we set it to 1.13.*?

requirements.txt
Louis-Dupont: I tried with 1.13 and got a security issue, I guess because of the nightly hack.
But apparently I can set other 1.13.X if I write X explicitely: 1.13.5 didnt raise the security issue despite not being released yet.

Should we stick to the latest stable released version (1.13.1) or set the requirements to a higher version in advance to not have to modify the requirements.txt next time ?
I tend to prefer the first (1.13.1) but I don't mind that much

ofrimasad (dismissed):
LGTM

ofrimasad (approved):
LGTM

PR #616, opened 2023-01-11
Interpollation crash tip

Louis-Dupont:
Objectif: Give a hint when people don't write interpolation with full path starting from config file, which is required by hydra

Example: '${dataset_params.train_dataloader_params.batch_size}' (good) instead of '${train_dataloader_params.batch_size}' (bad).

Note: hydra doenst tell us in what config file this breaks so we cannot show the exact solution. Instead we give hints with a general example.

image


shaydeci (approved):
nice, LGTM

shaydeci (approved):

PR #617, opened 2023-01-12
Feature/sg 000 break inner train loop

shaydeci:
max_train_batches feature, to break out of data loop during training.
Useful for debugging, needed for upcoming batch size selector feature.


ofrimasad (approved):
LGTM

PR #618, opened 2023-01-12
Group env variables

Louis-Dupont:
### Motivation
We are working with env variables but without any documentation/way to know which one, and where it is used.
I believe it is bad practice which makes the code harder to maintain (risk of duplicating existing env variables), but also to use (if I work on a new environment and want to setup everything, what env variable do I need to set ?).


src/super_gradients/common/env_variables.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+# Infra
+DECI_PLATFORM_TOKEN = os.getenv("DECI_PLATFORM_TOKEN")
```
shaydeci, line 4: Lets switch this to a class as you suggested (in order to avoid problems when a user would set the env vars after import of any of these)

shaydeci (changes requested):
1 small change we discussed then LGTM.

src/super_gradients/common/env_variables.py
Louis-Dupont: Changed: 
https://github.com/Deci-AI/super-gradients/blob/121b61114d08a0e33a1f92f0bde684cf72951ec8/src/super_gradients/common/environment/env_variables.py#L1-L39

shaydeci (approved):
LGTM

PR #622, opened 2023-01-15
Hotfix/sg 000 transform backward compatibility

Louis-Dupont:
Note: still need to update the tests

### DIFF
--- a/src/super_gradients/training/transforms/transforms.py
+++ b/src/super_gradients/training/transforms/transforms.py
@@ -786,7 +786,7 @@ class DetectionTargetsFormatTransform(DetectionTransform):
             raise TypeError(
                 "DetectionTargetsFormat is not supported for input_format and output_format starting from super_gradients==3.0.7.\n"
                 "You can either:\n"
-                "\t - use builtin format among super_gradients.training.datasets.data_formats.default_formats.<FORMAT_NAME> (like XYXY_LABEL, CXCY_LABEL, ..)\n"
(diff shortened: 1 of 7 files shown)


src/super_gradients/training/transforms/transforms.py
```diff
@@ -775,20 +775,43 @@ class DetectionTargetsFormatTransform(DetectionTransform):
...
+        input_dim: Optional[tuple] = None,
         input_format: ConcatenatedTensorFormat = XYXY_LABEL,
         output_format: ConcatenatedTensorFormat = LABEL_CXCYWH,
         min_bbox_edge_size: float = 1,
         max_targets: int = 120,
     ):
         super(DetectionTargetsFormatTransform, self).__init__()
+        if isinstance(input_format, DetectionTargetsFormat) or isinstance(output_format, DetectionTargetsFormat):
+            raise TypeError(
+                "DetectionTargetsFormat is not supported for input_format and output_format starting from super_gradients==3.0.7.\n"
```
shaydeci, line 787: I don't get this...

shaydeci (changes requested):
one clarification inline.

src/super_gradients/training/transforms/transforms.py
Louis-Dupont: I raise an error if they still use DetectionTargetsFormat (the Enum that we used before). It's to make the error explicit for a few versions.
Without this the error that they get if they dont change their old recipe is hard to understand.

shaydeci (approved):
LGTM

PR #623, opened 2023-01-15
fix default value of max_train_batches, max_valid_batches

Louis-Dupont:
### Problem
This "None" breaks the code because it is interpreted as a string.
to reproduce the error, you can run with --config-name=coco2017_yolox
Is it a bug? or did I miss something?

### Fix proposal
Just remove the default value, which automatically sets it to None


shaydeci (approved):
LGTM

PR #625, opened 2023-01-16
Add metadata to average_model

Louis-Dupont:
All the fields that were stored for other checkpoints (best, latest) are now added to average.


shaydeci (approved):
LGTM

PR #626, opened 2023-01-16
Register crashtips automatically

Louis-Dupont:
Small refacto class.
I want that any class inheriting from CrashTip is automatically registered, without needing to add the class to a list manually, because this step can easily be forgotten.


shaydeci (approved):
LGTM

PR #628, opened 2023-01-16
Feature/sg 541 auto batch selection

shaydeci:
Auto batch size selection, works out of the box for DDP.

Implemented as a "pre_launch_callback"- new feature needed for this design- upon calling train_from_config, there's often the need to modify the recipe config (for example to scale lr according to the actually number of available gpus).

The callback takes the cfg as an input, and increasingly tries different batch sizes until reaching COOM (or predefined max_batch_size), then optionally adapts the lr.

Additionally, I added the option to not kill the DDP group (needed for the consecutive batch size trial runs).


src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
```diff
@@ -0,0 +1,153 @@
...
+
+        For batch_size in range(min_batch_size:max_batch_size:size_step):
+            if batch_size raises CUDA OUT OF MEMORY ERROR:
+                return batch_size-size_step
+        return batch_size
+
+    Example usage: Inside the main recipe .YAML file (for example super_gradients/recipes/cifar10_resnet.yaml),
+     add the following:
+
+    pre_launch_callbacks_list:
```
BloodAxe, line 45: That's great addition to SG, but I try to think of the use cases and I find that adding the `pre_launch_callbacks_list` to config is slightly not I as a user would like to do.
Let me elaborate what I mean.
Suppose I have a recipe tailored for 8xA100. Batch size is hard-coded in YAML, num gpus as well.
I have 4x3090, that has less GPU memory on board.

So I may override this via command-line args:

`python train.py --config-name=my_recipe batch_size=16 num_gpus=4` (Let's assume we had 64 for A100)

This is our baseline. Whatever we're about to propose as a replacement should be AT LEAST as simple and as useful what user can do via cmdline.

Ideally, if I as a user can write this:

`python train.py --config-name=my_recipe batch_size=auto num_gpus=all`

And under the hood, these `auto`/`all` should trigger some PreLaunchCallback to infer num GPUS and batch size. 

Additionally, it is just impossible to add an item to `pre_launch_callbacks_list` list via command-line interface. So we are forcing user to add these lines to config file which may not be the in his intent. 

Certainly, for recipes that are meant to be executed on various setups it make sense to include `AutoTrainBatchSizeSelectionCallback` in the file. I can see three use-cases where we can use this feature:
* AutoNAC / CI - For this use-case the proposed approach is totally ok.
* Colab Notebooks - There guys certainly not going to modify recipes. And at most they can type `Trainer(..., batch_size="auto")`
* Regular users trying to fit recipe to their HW. I think they can do both command-line override and config modification, however command-line override is def. faster.

shaydeci: Thanks for the feedback (:

I completely agree that setting this option via command line is a good idea.
However, notice that there are multiple settings in which we can run the batch size selector callback, and they are quite sensitive (in the sense that setting defaults here might cause you to run forever until finding the right batch size). I am not confident this is the desired behaviour. Perhaps it's a good idea to ask around and see what answers we get.

Addressing the usages:
- We both agree for external use (i.e when someone has structured their own recipe) this is already useful.
- Now regarding Colab notebooks:
note that for for the callback to work, we are launching Trainer.train_from_config(), this is because we must access to dataset_params and modify them accordingly, as well as the learning rate which is in training params. My point here is that for coded train launch- we cant really offer auto batch size selection (i.e by just calling Trainer.train()) because Trainer.train() expects instantiated dataloaders. Re-instantation of (possibly external) dataloaders with a different batch size can get messy, especially when using samplers. With that being said, its probably possible, so if you have a suggestion how to do it I'm all ears (: .
I do have the feeling it would add much more complexity for a case that might not bee so common.
Also note that users can still use Trainer.train_from_config assuming they have their own config yaml files in the notebook file system somewhere.

-Third point- regular users using our recipes- this point I actually find the most interesting. We surely can add a shortcut for it. My concern is that since we don't have a "recipe base" at the moment, this might lead users to have errors if they did not inherit from our base recipe config (since they won't have "batch_size" arg in the main config file).

Lets further discuss these points.

src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
```diff
@@ -0,0 +1,153 @@
...
+        tmp_cfg.training_hyperparams.max_train_batches = self.num_forward_passes
+        tmp_cfg.training_hyperparams.run_validation_freq = 2
+        tmp_cfg.training_hyperparams.silent_mode = True
+        tmp_cfg.training_hyperparams.save_model = False
+        tmp_cfg.training_hyperparams.max_epochs = 1
+        tmp_cfg.training_hyperparams.average_best_models = False
+        tmp_cfg.training_hyperparams.kil_ddp_pgroup_on_end = False
+        tmp_cfg.pre_launch_callbacks_list = []
+
+        while True:
```
ofrimasad, line 106: why not use `for batch_size in range(min_batch_size, min_batch_size*1000, size_step)` ?

ofrimasad (commented):
minor comments

src/super_gradients/training/params.py
```diff
@@ -68,6 +68,7 @@
     "ckpt_name": "ckpt_latest.pth",
     "resume_strict_load": False,
     "sync_bn": False,
+    "kil_ddp_pgroup_on_end": True,  # Whether to kill the DDP process group in the end of training.
```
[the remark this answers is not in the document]
shaydeci: fixed in https://github.com/Deci-AI/super-gradients/pull/628/commits/91015297d46cd9c07dbbda64cf39553d19fdbb28

src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
shaydeci: Why should we limit our users ? Its very reasonable IMO that someone would start with a very small min batch size and large steps that would possibly reach this limit. Also if we replace step_size users still might reach this limit. This is the reason I added the optional max_batch_size.

ofrimasad (approved):
LGTM

PR #629, opened 2023-01-17
Feature/sg 624 remove avg from model plots

Louis-Dupont:
Currently SG adds the average model at the end of the training graph.

For instance if we train on 300 epochs, at the end of the training SG adds the average model validation metric/loss to the graph, leading to 301 data points on validation plots (vs 300 on training plots).

This is also the case with EarlyStop callback: if trained on 40 epochs, the average model will be added and we will have 41 data points, and even worse: the last point will be displayed at index 300!

This is bad especially since this information is stored in a separate graph (of only 1 data point per training), so I guess we should just remove this behavior.

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -1392,9 +1392,6 @@ class Trainer:
         self.net.load_state_dict(keep_state_dict)
 
         if not self.ddp_silent_mode:
-            # Adding values to sg_logger
-            # looping over last titles which corresponds to validation (and average model) metrics.
-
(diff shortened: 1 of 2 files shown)


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1394,10 +1394,6 @@ def _validate_final_average_model(self, cleanup_snapshots_pkl_file=False):
         if not self.ddp_silent_mode:
             # Adding values to sg_logger
             # looping over last titles which corresponds to validation (and average model) metrics.
```
ofrimasad, line 1396: please remove the documentation line as well

Louis-Dupont: done

shaydeci (approved):
LGTM

PR #632, opened 2023-01-17
Fixed id_tensor registry, so reparametrization works when .cuda() is called

spsancti:
…called on the module


src/super_gradients/modules/qarepvgg_block.py
```diff
@@ -133,10 +133,16 @@ def __init__(
...
 
             input_dim = self.in_channels // self.groups
-            self.id_tensor = torch.zeros((self.in_channels, input_dim, 3, 3))
+            id_tensor = torch.zeros((self.in_channels, input_dim, 3, 3))
             for i in range(self.in_channels):
-                self.id_tensor[i, i % input_dim, 1, 1] = 1.0
-            self.id_tensor = self.id_tensor.to(dtype=self.branch_1x1.weight.dtype, device=self.branch_1x1.weight.device)
+                id_tensor[i, i % input_dim, 1, 1] = 1.0
+
+            self.id_tensor: Optional[torch.Tensor]
```
ofrimasad, line 140: What is that line needed for? 
looks like you are overwriting `self.id_tensor` when calling `self.register_buffer`

ofrimasad (commented):
minor comment

src/super_gradients/modules/qarepvgg_block.py
spsancti: So Pycharm and flake8 do not show error when they can't find the attribute

ofrimasad (approved):
LGTM

PR #634, opened 2023-01-17
Feature/sg 573 pose estimation

BloodAxe:


src/super_gradients/training/datasets/pose_estimation_datasets/coco_keypoints.py
```diff
@@ -0,0 +1,198 @@
...
+from torch.utils.data import default_collate, Dataset
+
+from super_gradients.common.abstractions.abstract_logger import get_logger
+from super_gradients.training.transforms.keypoint_transforms import KeypointsCompose, KeypointTransform
+
+logger = get_logger(__name__)
+
+
+class COCOKeypointsDataset(Dataset):
+    """
```
shaydeci, line 21: Additionally, please add setup instructions ןמ src/super_gradients/training/datasets/Dataset_Setup_Instructions.md

src/super_gradients/training/transforms/keypoint_transforms.py
```diff
@@ -0,0 +1,302 @@
+import random
+from abc import abstractmethod
+from typing import Tuple, List, Iterable
+
+import cv2
+import numpy as np
+from torchvision.transforms import functional as F
+
+__all__ = [
```
shaydeci, line 9: I think it would be a good idea to add tests with fixed key points expected results for the below transforms.

src/super_gradients/recipes/dataset_params/coco_pose_estimation_dataset_params.yaml
```diff
@@ -0,0 +1,84 @@
...
+        output_size: [ 640, 640 ]
+        max_rotation: 30
+        min_scale: 0.75
+        max_scale: 1.5
+        scale_type: 'short'
+        max_translate: 40
+        image_pad_value: [ 127, 127, 127 ]
+        mask_pad_value: 1
+
+    - KeypointsToTensor: {}
```
shaydeci, line 34: no need for : {}

src/super_gradients/recipes/dataset_params/coco_pose_estimation_dataset_params.yaml
```diff
@@ -0,0 +1,84 @@
...
+  transforms:
+    - KeypointsLongestMaxSize:
+        max_sizes: [ 640, 640 ]
+
+    - KeypointsPadIfNeeded:
+        output_size: [ 640, 640 ]
+        image_pad_value: [ 127, 127, 127 ]
+        mask_pad_value: 1
+
+    - KeypointsToTensor: {}
```
shaydeci, line 58: no need for ": {}"

tests/integration_tests/pose_estimation_dataset_test.py
```diff
@@ -0,0 +1,36 @@
...
+from hydra import initialize_config_dir, compose
+from hydra.core.global_hydra import GlobalHydra
+
+import super_gradients
+from super_gradients.common.environment.path_utils import normalize_path
+from super_gradients.training.dataloaders.dataloaders import _process_dataset_params
+from super_gradients.training.datasets.pose_estimation_datasets import COCOKeypointsDataset
+
+
+class PoseEstimationDatasetIntegrationTest(unittest.TestCase):
```
shaydeci, line 14: nice, can we also add datalaoders instantiation to tests/unit_tests/datalaoder_factory_test.py ?

src/super_gradients/training/transforms/keypoint_transforms.py
```diff
@@ -0,0 +1,302 @@
...
+            mask = np.ascontiguousarray(np.flipud(mask))
+
+            rows, cols = image.shape[:2]
+            joints = joints.copy()
+            joints[:, :, 1] = rows - joints[:, :, 1] - 1
+
+        return image, mask, joints
+
+
+class KeypointsLongestMaxSize(KeypointTransform):
```
shaydeci, line 108: name is a bit misleading I feel..

shaydeci (changes requested):
Nice!
Besides some little comments, my main suggestions, detailed inline are:
- Regarding the way we handle transforms.I know we've yet to come up with a descent solution...but maybes better to handle transforms that are already implemented and can be resolved (which by itself, might impose a problem if we are allowing our users to resolve for example SegmentationTransform, but lets leave that for now..) by stating that non KeyPointsTransform are applied only on image, then we can reduce the amount of transforms we would need to reimplement simply for packing and unpacking items into the sample. See my suggestion inline, and let me know what you think (:
- Pose estimation base dataset- so I think it would be a good idea to pull at least some of the class methods of the dataset which we can probably take use in, and form a base class. If you feel that for COCO everything is very specific, we can probably skip this part and as time progresses and we have more datasets, the common methods will be more apparent and we can form a base class. 

See my other comments inline.

src/super_gradients/training/transforms/keypoint_transforms.py
```diff
@@ -0,0 +1,302 @@
...
+        format_string += "\n)"
+        return format_string
+
+
+class KeypointsToTensor(KeypointTransform):
+    def __call__(self, image: np.ndarray, mask: np.ndarray, joints: np.ndarray):
+        return F.to_tensor(image), mask, joints
+
+
+class KeypointsNormalize(KeypointTransform):
```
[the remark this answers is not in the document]
BloodAxe: We can have a base class ImageOnlyTransform class for photometric augmentations. But checking whether a given augmentation is Image-only a compose class would then pass only image in it.

src/super_gradients/recipes/dataset_params/coco_pose_estimation_dataset_params.yaml
BloodAxe: Ah, I see.
If I type just `- KeypointsToTensor` it's instantiated correct, but If I use `- KeypointsToTensor: ` then it fails inside base factory `return self.type_dict[_type](**_params)` since `_params` is None.

Do you think it make sense to handle this as well like so:
`return self.type_dict[_type](**_params or {})` ?

src/super_gradients/training/transforms/keypoint_transforms.py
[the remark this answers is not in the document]
shaydeci: mmmm but then again we would want all of the torch vision transforms to be included here, so that kind of defeats the purpose unless we define the torch vision transforms type to be included in ImageOnlyTransform type.

src/super_gradients/training/datasets/pose_estimation_datasets/coco_keypoints.py
BloodAxe: Nice catch. Added

src/super_gradients/recipes/dataset_params/coco_pose_estimation_dataset_params.yaml
shaydeci: but shouldn't it fail anyways in that case since None is not unpackable ?

shaydeci: Think we can leave it as is for now...

src/super_gradients/training/transforms/keypoint_transforms.py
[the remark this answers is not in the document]
BloodAxe: I agree that suggested change is a movement in the right direction. And perhaps we can schedule making it somewhere in upcoming sprints. But I feel it would affect transforms for other tasks, thus diluting the scope of change of this particular PR. Therefore if you're not strongly disagree, let's move the requested change to another ticket.

src/super_gradients/training/datasets/pose_estimation_datasets/coco_keypoints.py
```diff
@@ -0,0 +1,198 @@
...
+from super_gradients.common.factories.transforms_factory import TransformsFactory
+from torch.utils.data import default_collate, Dataset
+
+from super_gradients.common.abstractions.abstract_logger import get_logger
+from super_gradients.training.transforms.keypoint_transforms import KeypointsCompose, KeypointTransform
+
+logger = get_logger(__name__)
+
+
+class COCOKeypointsDataset(Dataset):
```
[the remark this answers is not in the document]
BloodAxe: Done! Thanks for the suggestion)

src/super_gradients/training/transforms/keypoint_transforms.py
```diff
@@ -0,0 +1,302 @@
...
+        self.max_rotation = max_rotation
+        self.min_scale = min_scale
+        self.max_scale = max_scale
+        self.scale_type = scale_type
+        self.max_translate = max_translate
+        self.image_pad_value = tuple(image_pad_value) if isinstance(image_pad_value, Iterable) else int(image_pad_value)
+        self.mask_pad_value = mask_pad_value
+        self.p = p
+
+    def _get_affine_matrix(self, center, scale, output_size: Tuple[int, int], rot=0):
```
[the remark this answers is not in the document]
BloodAxe: It was not compatible, but I greatly simplified current version.

src/super_gradients/training/transforms/keypoint_transforms.py
BloodAxe: Done

PR #635, opened 2023-01-17
Skip CI on .md files - SG side

shaydeci:


ranrubin (approved):
looks good. 
I suggest using yaml injection (I don't remember the actual name), instead of coping the environment every job
```
CIRCLE_COMPARE_URL: << pipeline.project.git_url >>/compare/<< pipeline.git.base_revision >>..<<pipeline.git.revision>>
```

What I mean is doing a similar thing to what we do with the branch filters.

PR #637, opened 2023-01-18
DDRNet39 module and cityscapes results

lkdci:
Support DDRNet39 which is bigger than DDRNet23 and reaches higher miou.
acc. to the paper and the original implementation, ddrnet39 differences with ddrnet23:
* more blocks
* wider segmentation head
* repeated of stage4 including doubling the interchange between the semantic and spatial path.

Author didn't report validation results on cityscapes, though the relative difference between ddrnet23 and ddrnet39 is about 1%,  similar gap to our results on the validation set


ofrimasad (approved):
LGTM

shaydeci (changes requested):
One request before LGTM on my end:

Please add a sanity test to .circleci/config.yml recipe_sanity_tests_segmentation (look at the other commands of the other recipes there, follow the same pattern of launching for one epoch and limiting the number of forward passes)

lkdci:
@shaydeci  Added sanity test

shaydeci (approved):
LGTM

PR #641, opened 2023-01-20
Added new detection transforms that are used in PPYoloE

BloodAxe:
Adding new detection transforms. They are used in PPYoloE, but can be used elsewhere and to keep PPYolo PR reasonable small, these are moved to separate PR


shaydeci (approved):
LGTM

PR #642, opened 2023-01-20
Remove calls cv2.setNumThreads(0) when importing SG

BloodAxe:
I believe it was copy-pasted from Albu. Back in the day it was necessary to fix the deadlock issue in old torch release. This call is not necessary since pytorch 0.4 :)
So there is no reason to keep it. It is not a mistake, but it prevent OpenCV from using multiprocessing under the hood.


shaydeci:
We actually had a problem with this for our DetectionDatase implementation.
[@Louis](https://github.com/Deci-AI/super-gradients/issues?q=is%3Apr+is%3Aopen+author%3ALouis-Dupont) perhaps you remember what the issue was Bach then and can elaborate ?

BloodAxe:
Curious to learn what it was. Haven't experienced any issues with torch+opencv for literally years)

shaydeci:
Im closing this down as we raised some concerns removing the calls for detection datasets.

PR #645, opened 2023-01-23
Feature/sg 568 single epoch sanity test for all recipes

shaydeci:
- Updated recipe configs of crashing recipes.
- New jobs and workflows for each regression test by task.


src/super_gradients/recipes/coco2017_ssd_lite_mobilenet_v2.yaml
```diff
@@ -35,10 +35,10 @@ architecture: ssd_lite_mobilenet_v2
 
 data_loader_num_workers: 8
 
-experiment_suffix: res${dataset_params.train_image_size}
+experiment_suffix: res320
```
ofrimasad, line 38: are these changes intentional?

src/super_gradients/recipes/coco_segmentation_shelfnet_lw.yaml
```diff
@@ -26,7 +26,7 @@ val_dataloader: coco_segmentation_val
 
 checkpoint_params:
   strict_load: True
-  load_backbone: True
+  load_backbone: False
```
ofrimasad, line 29: is this intentional?

ofrimasad (commented):
just some minor questions.
IF we can shorten the path to the test script somehow, that would be great

src/super_gradients/recipes/coco2017_ssd_lite_mobilenet_v2.yaml
shaydeci: Yes, this interpolation is not up to date (we removed all of these dataset_params.train_image_size when we got rid of datasetinterface)

ofrimasad (approved):
LGTM

PR #646, opened 2023-01-23
UNet registers

lkdci:
Added registers for UNet Bakbone stages and Up fuse blocks, for allowing users to use the base class with their custom blocks.
Added backbone ConvStage and decoder UpSumBlock


ofrimasad (approved):
LGTM

PR #647, opened 2023-01-23
Feature/sg 573 Integrate new EMA decay schedules

BloodAxe:
This PR brings option to use additional EMA decay schedules:

* **constant** - Fixed EMA decay value (Former config: `exp_activation: False`)
* **exp** - EMA decay warmup using exp rule `decay * (1 - math.exp(-x * self.beta))` (Former config: `exp_activation: True`)
* **threshold** - New schedule used in PP-Yolo training `min(decay, (1 + step) / (10 + step))`

Examples:

```
ema: True
ema_params:
  decay_type: constant
  decay: 0.9999
```

```
ema: True
ema_params:
  decay: 0.9999
  decay_type: exp
  beta: 15
```

```
ema: True
ema_params:
  decay_type: threshold
  decay: 0.9999
```

Note: We still keep backward compatibility with old params and emit a deprecation warning showing what should be updated:

```
    def test_train_with_old_ema_params(self):
        self._init_model()
        self._train({"decay": 0.9999, "exp_activation": True, "beta": 10})
```
Emits the following warning message:

```
[2023-01-24 10:33:37] WARNING - sg_trainer.py - Parameter `exp_activation` is deprecated for EMA model. Please update your config to use decay_type: str (constant|exp|threshold) instead:
ema: True
ema_params: 
  decay: 0.9999
  decay_type: exp
  beta: 15

ema: True
ema_params: 
  decay: 0.9999
  decay_type: constant

ema: True
ema_params: 
  decay: 0.9999
  decay_type: threshold
In the next major release of SG this warning will become an error.
```


src/super_gradients/training/kd_trainer/kd_trainer.py
```diff
@@ -255,17 +256,60 @@ def _get_hyper_param_config(self):
...
-        :param exp_activation:
+    def _instantiate_ema_model(self, decay_type: str = None, decay: float = None, **kwargs) -> KDModelEMA:
+        """Instantiate ema model for standard SgModule.
+        :param decay_type: (str) The decay climb schedule. See EMA_DECAY_FUNCTIONS for more details.
+        :param decay: The maximum decay value. As the training process advances, the decay will climb towards this value
+                      according to decay_type schedule. See EMA_DECAY_FUNCTIONS for more details.
+        :param kwargs: Additional parameters for the decay function. See EMA_DECAY_FUNCTIONS for more details.
         """
-        return KDModelEMA(self.net, decay, beta, exp_activation)
+        if decay is None:
```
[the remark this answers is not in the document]
ofrimasad: please also see my comment on the ModelEMA.__init__() function as well

src/super_gradients/training/utils/ema.py
```diff
@@ -30,7 +30,7 @@ class ModelEMA:
     GPU assignment and distributed training wrappers.
     """
 
-    def __init__(self, model, decay: float = 0.9999, beta: float = 15, exp_activation: bool = True):
+    def __init__(self, model, decay: float, decay_function: IDecayFunction):
```
ofrimasad, line 33: I think the proper way to handle this new signature is by using the `deprecate`  library (Important: note that this is "deprecate" without a d at the end. there are two python modules with close names).
you can see an example in the dataset_utils.

to use it, you will add a decorator on the __init__ function:
```
@deprecated(target=new_init, deprecated_in="3.0.6", remove_in="4.0.0", args_mapping=...,  args_extra=...)
def __init__(...)
```
in `args_mapping` and `args_extra` you can resolve all of the cases of new vs old arguments.
this library issues a formatted message with all of the information the user will need. 

we may need to figure out exactly how to use it in this unique case. since we don't want the user to start using another init function (we just want him to change the arguments). so perhaps instead of using the decorator, we can use `_raise_warn_arguments()` that issues the message, without defining a new target function

src/super_gradients/training/utils/ema_decay_schedules.py
```diff
@@ -0,0 +1,38 @@
+import math
+from abc import abstractmethod
+
+__all__ = ["IDecayFunction", "ConstantDecay", "ThresholdDecay", "ExpDecay", "EMA_DECAY_FUNCTIONS"]
+
+
+class IDecayFunction:
+    @abstractmethod
+    def __call__(self, decay: float, step: int, total_steps: int):
```
ofrimasad, line 9: please add a documentation here, in case someone want to write his own decay function

ofrimasad (commented):
looks great. 
I think we should use `deprecate` library for this deprecation. see my comments inline

src/super_gradients/training/utils/ema.py
BloodAxe: If we don't use `@deprecated` decorator, than what's the point of using `deprecate` at all in this case? 
Side note: This library has as few as 46 starts on GitHub. This does not necessary means it is bad library, rather than it's support time can be questioned.

ofrimasad (approved):
LGTM

PR #653, opened 2023-01-25
Fix grad clipping stage + upscale gradients before clipping

shaydeci:
- Clipping was done each batch- moved to being done once before gradient step.
- Unscale gradients before preforming gradient step (note that unsealing is preformed in optimizer.step if not explicitly done, so only need to unscale them before clipping, more info https://pytorch.org/docs/stable/notes/amp_examples.html?highlight=gradient+clipping)


ofrimasad (approved):
LGTM

PR #654, opened 2023-01-25
Feature/sg 631 add register for collate samplers

Louis-Dupont:


src/super_gradients/common/object_names.py
```diff
@@ -282,3 +282,9 @@ class ConcatenatedTensorFormats:
     LABEL_NORMALIZED_XYXY = "LABEL_NORMALIZED_XYXY"
     LABEL_NORMALIZED_XYWH = "LABEL_NORMALIZED_XYWH"
     LABEL_NORMALIZED_CXCYWH = "LABEL_NORMALIZED_CXCYWH"
+
+
+class CollateFunctions:
+    DEFAULT_COLLATE_FN = "DEFAULT_COLLATE_FN"
```
shaydeci, line 288: You missed some collate fns (mixup, mutiscale...cmd+shift+f collate and add the missing ones please).

src/super_gradients/training/utils/collate_fn/all_collate_fn.py
```diff
@@ -0,0 +1,10 @@
+from super_gradients.common.object_names import CollateFunctions
+from torch.utils.data._utils.collate import default_collate
+from super_gradients.training.utils.collate_fn.detection import DetectionCollateFN, CrowdDetectionCollateFN
+
+
+COLLATE_FUNCTIONS = {
+    CollateFunctions.DEFAULT_COLLATE_FN: default_collate,
+    CollateFunctions.DETECTION_COLLATE_FN: DetectionCollateFN(),
```
shaydeci, line 8: This might be problematic...you will always use the same object when mapping them like this.

src/super_gradients/training/utils/collate_fn/detection.py
```diff
@@ -0,0 +1,35 @@
+from typing import Tuple, Dict
+
+import torch
+from torch.utils.data._utils.collate import default_collate
+
+
+class DetectionCollateFN:
```
shaydeci, line 7: If you already decided to concentrate the different collate fns here- make sure to add the missing ones (discussed in earlier comment).

shaydeci (changes requested):
Just 2 minor issues:
- Missed some collate functions we have in our code.
- Also addressed another instantiation issue inline.

Also- where is the samplers factory ?

PR #659, opened 2023-01-26
Added qat_from_recipe script

spsancti:


src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,220 @@
...
+        logger.warning(f"Recipe requests multi_gpu={cfg.multi_gpu} and num_gpus={cfg.num_gpus}. Changing to multi_gpu=OFF and num_gpus=1")
+        cfg.multi_gpu = "OFF"
+        cfg.num_gpus = 1
+
+    # no augmentations
+    cfg.dataset_params.train_dataset_params.transforms = cfg.dataset_params.val_dataset_params.transforms
+
+
+@hydra.main(config_path=pkg_resources.resource_filename("super_gradients.recipes", ""), version_base="1.2")
+def main(cfg: DictConfig) -> None:
```
NatanBagrov, line 135: Please refactor this into 2 methods: one has Hydra's decorator, instantiation of the cfg, and then calls the other function that accepts an instantiated cfg.

```python
def qat_from_config(cfg):
   if "quantization_params" not in cfg:
        raise ValueError("Your recipe does not have quantization_params. Add them to use QAT.")
    ...

@hydra(...)
def main(cfg):
   cfg = hydra.utils.instantiate(cfg)
   return qat_from_config(cfg)
```
This way we can reuse the function that accepts a "ready" cfg.

NatanBagrov (commented):
Thanks a lot!
See comment inline

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,239 @@
...
+from super_gradients.training.utils.distributed_training_utils import setup_device
+from super_gradients.training.utils.module_utils import fuse_repvgg_blocks_residual_branches
+from super_gradients.training.utils.quantization.calibrator import QuantizationCalibrator
+from super_gradients.training.utils.quantization.export import export_quantized_module_to_onnx
+from super_gradients.training.utils.quantization.selective_quantization_utils import SelectiveQuantizer
+
+logger = get_logger(__name__)
+
+
+def quantize_and_calibrate(
```
shaydeci, line 40: Lets move this to a more appropriate place (see summary).
Also, please add docstring.

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,239 @@
...
+        num_calib_batches=num_calib_batches,
+        percentile=percentile,
+    )
+    model.to(device=device)  # we can have _amax buffers scattered over different devices
+
+    model.train()
+    return model
+
+
+def modify_training_params_for_qat(cfg):
```
shaydeci, line 89: Please re-implement as a pre_launch_callback (see summary).

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,239 @@
...
+        logger.warning(f"Recipe requests multi_gpu={cfg.multi_gpu} and num_gpus={cfg.num_gpus}. Changing to multi_gpu=OFF and num_gpus=1")
+        cfg.multi_gpu = "OFF"
+        cfg.num_gpus = 1
+
+    # no augmentations
+    if "transforms" in cfg.dataset_params.val_dataset_params:
+        cfg.dataset_params.train_dataset_params.transforms = cfg.dataset_params.val_dataset_params.transforms
+
+
+def qat_from_config(cfg):
```
shaydeci, line 146: See my point regarding inheriting from SgTrainer and modifying train_from_config.

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    if "quantization_params" not in cfg:
+        raise ValueError("Your recipe does not have quantization_params. Add them to use QAT.")
+
+    if "checkpoint_path" not in cfg.checkpoint_params:
+        raise ValueError("Starting checkpoint is a must for QAT finetuning.")
+
+    if cfg.quantization_params.modify_recipe_for_qat.enable:
+        modify_training_params_for_qat(cfg=cfg)
+
+    setup_device(
```
shaydeci, line 156: Also, please extract the mutual code from here to the ancestor SgTrainer (see previous comment).

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
```diff
@@ -0,0 +1,21 @@
+modify_recipe_for_qat:
```
shaydeci, line 1: See summary regarding extracting theses recommended practices parameters as a pre_launch_callback.
This should look pretty much the same actually, but just at the top of the config (i.e the inheriting recipe from the original).

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,239 @@
...
+
+    quantize_and_calibrate(
+        model,
+        train_dataloader,
+        method_w=cfg.quantization_params.selective_quantizer_params.method_w,
+        method_i=cfg.quantization_params.selective_quantizer_params.method_i,
+        per_channel=cfg.quantization_params.selective_quantizer_params.per_channel,
+        learn_amax=cfg.quantization_params.selective_quantizer_params.learn_amax,
+        skip_modules=cfg.quantization_params.selective_quantizer_params.skip_modules,
+        num_calib_batches=cfg.quantization_params.calib_params.num_calib_batches or (512 // cfg.dataset_params.train_dataloader_params.batch_size) or 1,
```
shaydeci, line 192: what's this 512 voodoo ?

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    logger.info("Performing validation of PTQ model...")
+
+    trainer = Trainer(experiment_name=cfg.experiment_name, ckpt_root_dir=get_param(cfg, "ckpt_root_dir", default_val=None))
+    val_results_tuple = trainer.test(model=model, test_loader=val_dataloader, test_metrics_list=cfg.training_hyperparams.valid_metrics_list)
+    valid_metrics_dict = get_metrics_dict(val_results_tuple, trainer.test_metrics, trainer.loss_logging_items_names)
+    results = ["Validate Results"]
+    results += [f"   - {metric:10}: {value}" for metric, value in valid_metrics_dict.items()]
+    logger.info("\n".join(results))
+
+    if cfg.training_hyperparams.max_epochs != 0:
```
shaydeci, line 207: Im not sure I'm in favour of diffrentiating between PTQ and QAT on the basis of cfg.training_hyperparams.max_epochs ....
Can we make this more explicit ?

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
shaydeci: Can we also separate this into calibrate and quantise ?

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+selective_quantizer_params:
+  method_w: "max"           # calibrator type for weights, acceptable types are ["max", "histogram"]
+  method_i: "histogram"     # calibrator type for inputs acceptable types are ["max", "histogram"]
+  per_channel: True         # per-channel quantization of weights, activations stay per-tensor by default
+  learn_amax: False         # enable learnable amax in all TensorQuantizers using straight-through estimator
+  skip_modules: null        # optional list of module names (strings) to skip from quantization
+
+calib_params:
+  calib_method: "percentile"  # calibration method for all "histogram" calibrators, acceptable types are ["percentile", "entropy", mse"], "max" calibrators are not affected
```
shaydeci, line 18: Its not so clear why there is method_w, method_i, learn_amax alongside with calib_params

shaydeci (changes requested):
It is very important that we reuse essential parts of the code for QAT/PTQ. It seems here that much of the logic resides in functions inside the qat_from_recipe.py, which is not a clear part of SG. 
Also, we want to keep these examples as straightforward as possible so that when someone from CFDLE (or any user actually) needs to run something similar but with some tweaks - they would do it seamlessly.

My design suggestions:
- Introduce a new class: QATrainer (or find a better name), inheriting from Trainer. This class should override Trainer's train_from_config, which should look pretty similar to what you implemented inside qat_from_recipe.
- Regarding the quantisation_params: we can implement a pre_launch_callback that would do exactly that. Then for some recipes, for instance, "imagenet_resnet50" we can simply inherit from it and just add this pre_launch_callback + quantisation parameters.
This would also allow pretty straightforward adjustment of existing recipes.

I tried to further explain my suggestions in the comments. There are also a few other ones regarding smaller changes.

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
spsancti: pytorch-quantization recommendation

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
spsancti: selective quantizer only adds calibrators for weights and inputs, which can be "max" and "histogram" but they don't know anything about calibration yet
Calibrator needs method names, which are ["max"] for "max" calibrator and ["mse", "entropy", "percentile"] for "histogram" one
Maybe, it's better to name it like "calib_i" and "calib_w" instead of "method_*" to make it less confusing

shaydeci: But then doesn't it mean calib_method can be inferred from method_i and method_w ?

Is it useful to use different calibrators for inputs and weights ?

shaydeci (changes requested):

src/super_gradients/examples/qat_from_recipe_example/qat_from_recipe.py
spsancti: There are many small changes in flow, let's leave it for another day

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
spsancti: calib_method is only relevant for "histogram" calibrators
"max" calibrators have only one option, which is "max" 

With selective quantizer we insert calibrators, which can only be of two types. With QuantizationCalibrator we call inserted calibrators, and for the histogram calibrator we have three different options ["mse", "entropy", "percentile"]

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
```diff
@@ -0,0 +1,13 @@
...
+  calibrator_w: "max"        # calibrator type for weights, acceptable types are ["max", "histogram"]
+  calibrator_i: "histogram"  # calibrator type for inputs acceptable types are ["max", "histogram"]
+  per_channel: True          # per-channel quantization of weights, activations stay per-tensor by default
+  learn_amax: False          # enable learnable amax in all TensorQuantizers using straight-through estimator
+  skip_modules: null         # optional list of module names (strings) to skip from quantization
+
+calib_params:
+  histogram_calib_method: "percentile"  # calibration method for all "histogram" calibrators, acceptable types are ["percentile", "entropy", mse"], "max" calibrators always use "max"
+  percentile: 99.99                     # percentile for all histogram calibrators with method "percentile", other calibrators are not affected
+  num_calib_batches: null               # number of batches to use for calibration, if null, 512 / batch_size will be used
```
shaydeci, line 12: I think this will be resolved to str("null"), just leave empty for None.

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
```diff
@@ -0,0 +1,13 @@
+ptq_only: False
+selective_quantizer_params:
+  calibrator_w: "max"        # calibrator type for weights, acceptable types are ["max", "histogram"]
+  calibrator_i: "histogram"  # calibrator type for inputs acceptable types are ["max", "histogram"]
+  per_channel: True          # per-channel quantization of weights, activations stay per-tensor by default
+  learn_amax: False          # enable learnable amax in all TensorQuantizers using straight-through estimator
+  skip_modules: null         # optional list of module names (strings) to skip from quantization
```
shaydeci, line 7: I think this will be resolved to str("null"), just leave empty for None.

src/super_gradients/recipes/quantization_params/default_quantization_params.yaml
```diff
@@ -0,0 +1,13 @@
+ptq_only: False
```
shaydeci, line 1: Add comment explaining this controls whether QAT is launched or not.

shaydeci (changes requested):
Looks great!
Just a few last minor comments inline.

src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
```diff
@@ -151,3 +152,105 @@ def _clear_model_gpu_mem(cls, model):
...
+            logger.warning(f"New dataset_params.val_dataloader_params.batch_size: {cfg.dataset_params.val_dataloader_params.batch_size}")
+
+        cfg.training_hyperparams.max_epochs //= self.max_epochs_divisor
+        logger.warning(f"New number of epochs: {cfg.training_hyperparams.max_epochs}")
+
+        cfg.training_hyperparams.initial_lr *= self.lr_decay_factor
+        if cfg.training_hyperparams.warmup_initial_lr is not None:
+            cfg.training_hyperparams.warmup_initial_lr *= self.lr_decay_factor
+        else:
+            cfg.training_hyperparams.warmup_initial_lr = cfg.training_hyperparams.initial_lr * 0.01
```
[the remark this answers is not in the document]
spsancti: No, it's indeed 0.01 and fixed, lr_decay factor can go way down to 1e-4 which is infeasible here

shaydeci (approved):
LGTM

PR #661, opened 2023-01-29
documentation on using configuration files

ofrimasad:


documentation/source/configuration_files.md
```diff
@@ -0,0 +1,248 @@
...
+```yaml
+...
+architecture: MyNet
+    num_classes: 8
+...
+
+```
+
+## Recipes library structure
+The `super_gradients/recipes` include the following subdirectories:
```
shaydeci, line 241: I think should mention the other parameters that are at the top level (num gpu, multi_gpu..)- otherwise users might crash their code.

documentation/source/configuration_files.md
```diff
@@ -0,0 +1,248 @@
...
+    - ToTensor
+    - Normalize:
+        mean: ${dataset_params.img_mean}
+        std: ${dataset_params.img_std}
+```
+
+Configuration file can also help you track the exact settings used for each one of your experiments, tweak and tune these settings, and share them with others.
+Concentrating all of these configuration parameters in one place, gives you great visibility and control of your experiments. 
+
+## How to use configuration files
```
shaydeci, line 78: There are nice features we've worked on that give the configuration files more uses than just training. I think resume_experiment and evaluate_experiment should be mentioned as well.

documentation/source/configuration_files.md
```diff
@@ -0,0 +1,248 @@
...
+```commandline
+python train_from_recipe.py --config-name=cifar10_resnet training_hyperparams.initial_lr=0.02 experiment_name=test_lr_002
+```
+In the example above, the same script we launched earlier is used, but this time it will run with a different experiment name and a different 
+initial learning-rate. This feature is extremely usefully when experimenting with different hyper-parameters.
+Note that the arguments are referenced without the `--` prefix and that each parameter is referenced with its full path in the 
+configuration tree, concatenated with a `.`.
+
+## Resolvers 
+Resolvers are converting the strings from the YAML file into Python objects or values. The most basic resolvers are the Hydra native resolvers.
```
shaydeci, line 153: Not sure why users would need to know about this...

shaydeci (changes requested):
Some suggestions inline.

documentation/source/configuration_files.md
```diff
@@ -0,0 +1,248 @@
...
+        size: 256
+    - CenterCrop:
+        size: 224
+    - ToTensor
+    - Normalize:
+        mean: ${dataset_params.img_mean}
+        std: ${dataset_params.img_std}
+```
+
+Configuration file can also help you track the exact settings used for each one of your experiments, tweak and tune these settings, and share them with others.
```
shaydeci, line 75: Maybe mention the .hydra directory inside checkpoints/experiment_name that will hold the final composed configuration.

shaydeci (changes requested):

documentation/source/configuration_files.md
```diff
@@ -0,0 +1,288 @@
...
+architecture: 
+    MyNet:
+      num_classes: 8
+...
+
+```
+
+## Required Hyper-Parameters
+Most parameters can be defined by default when including `default_train_params` in you `defaults`.
+However, the following hyper-parameters are required to launch a training run:
```
shaydeci, line 261: also multi_gpu and num_gpus in case distributed training support is wanted.

shaydeci (changes requested):
Just one last thing inline...

documentation/source/configuration_files.md
ofrimasad: fixed

ofrimasad:
fixed

ofrimasad (commented):
addressed

documentation/source/configuration_files.md
shaydeci: where ?

shaydeci (changes requested):

shaydeci (approved):
LGTM

PR #664, opened 2023-01-29
Add log readme

Louis-Dupont:


shaydeci (changes requested):
I feel like this is missing quite a lot of what we are aiming for...

All we do here is flash some parameters to integrate WandB and Clearml, when I think the appropriate thing to do here is to explain about SG loggers, detailing how one can access their methods (phase callbacks).

Also, there's quite a lot to say on what values we log by default, what values we print etc.
Don't forget about Tensorboards too.

I think there's quite a lot to detail about regarding training with configs- there's the .hydra directory, the "recipe_config" which is logged when using train_from_config (and if we are already talking about additional logs- additional_logs that one can pass through train() is not mentioned here either).

documentation/source/Logs.md
```diff
@@ -0,0 +1,73 @@
+# SuperGradient Logging system
+
+## Console logging
+For better debugging and understanding of past runs, SuperGradients gathers all the print statements and logs into a 
+local file, providing you the convenience to review console outputs of any experiment at any time.
+
+**Where is it saved?**
+- Upon importing SuperGradients, console outputs and logs will be stored in `~/sg_logs/console.log`.
+- When instantiating the super_gradients.Trainer, all console outputs and logs will be redirected to the experiment folder `/<path>/<to>/<experiments>/<experiment_name>/sg_logs_<date>.log`.
```
ofrimasad, line 9: I think it is saved in console_<date>.txt

ofrimasad (commented):
Please note that the SGLoggers should be on a different document (Monitoring). copy them to another document. 
missing in this file are the different types of logs and what is in each one of them. 
- console_<date>.txt
- log_<date>.txt
- sg_logs_<date>.txt 
Also, how to set a different log level
Crash tips and how to turn them off

documentation/source/logs.md
```diff
@@ -0,0 +1,102 @@
...
+For better debugging and understanding of past runs, SuperGradients gathers all the print statements and logs into a 
+local file, providing you the convenience to review console outputs of any experiment at any time.
+
+**What does it include?** All the prints and logs that were displayed on the console, but not the filtered logs.
+
+**Where is it saved?**
+- Upon importing SuperGradients, console outputs and logs will be stored in `~/sg_logs/console.log`.
+- When instantiating the super_gradients.Trainer, all console outputs and logs will be redirected to the experiment folder `<ckpt_root_dir>/<experiment_name>/console_<date>.txt`.
+
+**How to set log level?** You can filter the logs displayed on the console by setting `CONSOLE_LOG_LEVEL=<LOG-LEVEL> # DEBUG/INFO/WARNING/ERROR`
```
ofrimasad, line 38: maybe states "setting *environment variable* ..."

ofrimasad (changes requested):
Beautiful documentation!
just remove the code that was already pushed in the second PR please.
see my minor comment inline

documentation/source/logs.md
Louis-Dupont: changed

ofrimasad (approved):
LGTM

shaydeci (approved):
LGTM

PR #667, opened 2023-01-30
Feature/sg 646 rename logs

Louis-Dupont:
### 1. Rename
`log_<date>` -> `experiment_logs_<date>` (This includes only experiment related logs: config and training/validation results per epoch)
`sg_logs_<date>` -> `logs_<date>` (This actually also includes logs that are not from SG)

### 2. Upload all the logfiles to platform


ofrimasad (approved):
Looks great!

PR #672, opened 2023-01-31
Hotfix/sg 000 temporary mute upload print

Louis-Dupont:
Currently, we get many prints to say that a file is being uploaded. Yesterday we chose to upload many logs, but in some cases it leads to this message being printed like 8 times, and it looks bad...
image

The platform team has a ticket to fix this, but if we want to release before they fix+release a version, we can move stdout to logging.debug.
This might be necessary if we want to release quickly, otherwise, the (pro) user would get all these prints...

Note: relies on https://github.com/Deci-AI/super-gradients/pull/671


shaydeci (approved):
LGTM

PR #682, opened 2023-02-02
Feature/sg 614 ema docs

BloodAxe:


shaydeci (changes requested):
Very nice!!!

I only have one suggestion- can you please add some short explanation regarding where the EMA net is saved (i.e ema_net inside the state_dict, no need to go deep into details since we are also adding a checkpoints.md file) and how it can be easily loaded for testing (models.get will take by default the ema net if it exists using the load_ema_as net arg).

BloodAxe:
@shaydeci have to figure out by myself how this thing works :) but thanks for the suggestion.

shaydeci (approved):
LGTM

PR #689, opened 2023-02-05
Feature/sg 587 warn if unused on models

Louis-Dupont:
Issue faced:

image


tests/unit_tests/get_model_test.py
```diff
@@ -0,0 +1,25 @@
+import io
+import unittest
+
+from super_gradients.training import models
+from super_gradients.training.utils.config_utils import IS_UNUSED_MESSAGE_INTRO
+
+
+class GetModelTest(unittest.TestCase):
```
shaydeci, line 8: Please add to suite.

shaydeci, line 8: Also maybe a better name ?

src/super_gradients/training/models/model_factory.py
```diff
@@ -169,27 +170,29 @@ def get(
 
     NOTE: Passing pretrained_weights and checkpoint_path is ill-defined and will raise an error.
     """
-    checkpoint_num_classes = checkpoint_num_classes or num_classes
+    arch_params = arch_params if arch_params else {}
+    with warn_if_unused_params(arch_params) as arch_params:
```
shaydeci, line 174: It would be great if we start reducing the amount of warning duplicates when running in DDP.
Can we do it here explicitly? or even better inside warn_if_unused_params ?

shaydeci (changes requested):
Inline.

src/super_gradients/training/models/model_factory.py
Louis-Dupont: DDP will display it only once because the loggers are muted for rank>1.
Also models.get is usually called after destup_device, so the launcher will not raise any warning.
The conclusion is that with/without DDP, it will only print once

tests/unit_tests/get_model_test.py
Louis-Dupont: I actually moved it to integration test folder, because this is what it actually is

Louis-Dupont: Changed to WarnIfUnusedIntegrationTest

PR #692, opened 2023-02-06
created external model example MD file with images

eran-deci:


documentation/assets/Example: training an external model.md
```diff
@@ -0,0 +1,753 @@
...
+tensor's elements to be integers. Therefore, we will need to modify it a bit. For more details about implementing a 
+custom metric using torchmetrics, see [here](https://torchmetrics.readthedocs.io/en/stable/pages/implement.html).
+
+```
+class SoftIoU(JaccardIndex):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def update(self, preds: torch.Tensor, target: torch.Tensor):
+        target = torch.gt(target, 0.5).long()
```
shaydeci, line 528: This will take the mean of the ious for both background and the target 
(therefore the results are so promising so fast).

For binary segmentation it is more common to take the you of target only.

documentation/assets/Example: training an external model.md
```diff
@@ -0,0 +1,753 @@
...
+
+class CustomSegLoss(torch.nn.Module):
+    def __init__(self, bce_weight=1, iou_weight=1):
+        super(CustomSegLoss, self).__init__()
+
+        self.bce_weight = bce_weight
+        self.iou_weight = iou_weight
+
+        self.bce_loss = nn.BCELoss()
+        self.iou_func = JaccardIndex(num_classes=2)
```
shaydeci, line 483: from torchmetrics.classification import JaccardIndex
JaccardIndex.is_differentiable
Out[3]: False

Its a good idea to add iou component to the loss, but its not what is happening here. You are basically training just on BCE + some numbers as far as I can see.

documentation/assets/Example: training an external model.md
```diff
@@ -0,0 +1,753 @@
...
+
+        return self.bce_weight*bce_loss + self.iou_weight*iou_loss
+```
+
+Notice that here the BCE loss term is obtained simply by using PyTorch's `BCELoss()`. To compute the IoU score, or
+the Jaccard Index, we use torchmetric's `JaccardIndex`. To compute the IoU, the pixel values in both images should
+be binary, i.e., 0's and 1's. Since in the model's forward function a sigmoid function is already applied to the
+output, we only need to binarize both the predictions and the target tensors with a threshold of 0.5 (remember, we are 
+using **soft** masks). When measuring segmentation performance, higher IoU is better. Since IoU score is a number 
+in [0, 1], we simply compute `IoU loss = 1 - IoU` to make it a valid loss function for gradient-descent optimization.
```
shaydeci, line 498: If you decide to drop the ious component make sure to fix this as well.

shaydeci (changes requested):
Very nice tutorial!
I just have a few minor comments inline regarding some minor inaccuracies.

documentation/assets/Example: training an external model.md
shaydeci: Should be something like this:

class SoftIoU(JaccardIndex):
     def __init__(self, **kwargs):
         super().__init__(reduction="none", **kwargs)
     def update(self, preds: torch.Tensor, target: torch.Tensor):
         target = torch.gt(target, 0.5).long()
         super().update(preds, target)
     def compute(self):
          return super().compute()[1]

eran-deci: nice one! 
changed according to your suggestion

documentation/assets/Example: training an external model.md
eran-deci: Implemented a custom IoU loss, which is differentiable

documentation/assets/Example: training an external model.md
eran-deci: no need, kept the IoU

shaydeci (approved):
LGTM

PR #695, opened 2023-02-06
fix the way unet loads params + some recipe facory strings

Louis-Dupont:
## Issue
`get_arch_params` loads the `arch_params` directly from the .`../recipes/arch_params` folder

- This means that the interpolation is “**local**” (e.g. `backbone.width_list` works)

When running `train_from_config`, the whole recipe is loaded from `.../recipes` folder.

- This means that the interpolation is “**absolute**” (e.g. `arch_params.backbone.width_list` works)

This leads to an incompatibility of the interpolation format.

## Solution
Because `train_from_config` will always need to load the `arch_params` from `.../recipes` folder, I propose to change the `get_arch_params` logic to also load config from the `.../recipes` folder.


shaydeci (approved):
LGTM

PR #699, opened 2023-02-07
Add Troubleshooting md

Louis-Dupont:


documentation/source/troubleshooting.md
```diff
@@ -0,0 +1,62 @@
...
+- Decrease the number of batch accumulation steps (`training_hyperparams.batch_accumulate`)
+
+
+
+## CUDA error: device-side assert triggered
+
+You may encounter a generic CUDA error message that lacks information regarding the cause of the error:
+
+```
+RuntimeError: CUDA error: device-side assert triggered
```
shaydeci, line 45: There is actually a way to get the underlying CUDA errors/warnings that is a lot more informative...
I don't remember exactly what the env var is called, but I used it in the past. Will get back to you.

documentation/source/troubleshooting.md
```diff
@@ -0,0 +1,62 @@
...
+
+When [running on CPU](device.md) you won't have this issue of CUDA hiding the root cause of the error.
+
+**2. Set Environment Variable**
+
+You can add the following environment variable at the beginning of your code.
+
+```py
+import os
+os.environ["CUDA_LAUNCH_BLOCKING"] = "1"
```
shaydeci, line 60: Maybe it's better to tell our users to set the environment variable before launching in case the run crashes before its set inside the code.

shaydeci (changes requested):
Comments inline.
I will get back to you regarding the environment variable name (2nd comment).

documentation/source/troubleshooting.md
Louis-Dupont: isnt it CUDA_DEVICE_ASSERT=1?

documentation/source/troubleshooting.md
Louis-Dupont: Changed in d85742e8322e4bb6d059b63e012702dd45e12d61

documentation/source/troubleshooting.md
Louis-Dupont: I added CUDA_DEVICE_ASSERT in d85742e8322e4bb6d059b63e012702dd45e12d61, but I think that the NCCL_DEBUG=INFO  we talked about is for other issues concerning the communication between DDP nodes.

documentation/source/troubleshooting.md
```diff
@@ -0,0 +1,62 @@
...
+
+It is pretty common to run out of memory when using GPU. This is shown with following exception:
+
+```
+CUDA out of memory. Tried to allocate 20.00 MiB (GPU 0; 10.76 GiB total capacity; 4.29 GiB already allocated; 10.12 MiB free; 4.46 GiB reserved in total by PyTorch)
+```
+
+To reduce memory usage, try the following:
+- Decrease the batch size (`dataset_params.train_dataloader_params.batch_size` and `dataset_params.val_dataloader_params.batch_size`)
+- Decrease the number of batch accumulation steps (`training_hyperparams.batch_accumulate`)
```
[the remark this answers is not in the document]
Louis-Dupont: You're right, changed in d85742e

shaydeci (approved):
LGTM

PR #700, opened 2023-02-08
Fixes- checkpoint loading

shaydeci:
- Added default values for arguments in load_checkpoint_to_model so we can simplify its use for the users.
- Made _load_weights which is used by load_pretrained_weights to use adaptive_load_state_dict, which falls back to no_key_matching when load_state_dict fails (this is the problem @bloodaxe had with the PPYoloE weights)..


ofrimasad (approved):
LGTM

PR #702, opened 2023-02-08
Add "resume" option when resuming DeciPlatformClient 

Louis-Dupont:
This allows the user to upload a tensorboard related to any of the AUTONAC model from the platform.

The platform team will implement the possibility to also add custom models to the platform interface in the next sprint. This will allow the user to also have tensorboard of custom models.


shaydeci (approved):
LGTM

PR #705, opened 2023-02-11
Fix AccessCounterMixin 

Louis-Dupont:
2 issues appeared:
- AccessCounterDict does not support item assignment (__setitem__)
- AccessCounterMixin does not support deepcopy (which is called when deep copying a model)


src/super_gradients/training/utils/config_utils.py
```diff
@@ -63,9 +64,23 @@ def get_unused_params(self) -> Set[str]:
...
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            setattr(result, k, deepcopy(v, memo))
+        return result
+
 
 class AccessCounterDict(Mapping, AccessCounterMixin):
-    def __init__(self, config: Mapping, access_counter: Mapping[str, int] = None, prefix: str = ""):
+    def __init__(self, config: dict, access_counter: Mapping[str, int] = None, prefix: str = ""):
```
ofrimasad, line 83: why was that changed to `dict`?
we want to support other types as well (like `DictConfig` that does not inherit from dict)

ofrimasad (commented):
one comment

src/super_gradients/training/utils/config_utils.py
Louis-Dupont: `Mapping` doesn't necessarily support item assignment. I add DictConfig in the typing: 4afca0dea77e2ea98c6bd3ca6b51b38e2dc970bc

BloodAxe:
Fixed exception masking. LGTM

shaydeci (approved):
LGTM

PR #707, opened 2023-02-12
pixel shuffle in unet decoder

lkdci:
Support pixel shuffle as an upsample mode in UNet Decoder stages


src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
```diff
@@ -34,8 +35,16 @@ class UpFactorBlock(AbstractUpFuseBlock):
     Ignore Skip features, simply apply upsampling and ConvBNRelu layers.
     """
 
-    def __init__(self, in_channels: int, skip_channels: int, out_channels: int, up_factor: int, mode: str, num_repeats: int, **kwargs):
+    def __init__(self, in_channels: int, skip_channels: int, out_channels: int, up_factor: int, mode: str, num_repeats: int, fallback_mode: str, **kwargs):
```
BloodAxe, line 38: Two questions here:
1) Why not make/annotate `mode` & `fallback_mode` of `UpsampleMode` type?
2) Why do we prefer `fallback_mode` over raising an exception if requested mode is not compatible?

src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
```diff
@@ -72,12 +91,21 @@ class UpSumBlock(AbstractUpFuseBlock):
     Fuse features with concatenation and followed Convolutions.
     """
 
-    def __init__(self, in_channels: int, skip_channels: int, out_channels: int, up_factor: int, mode: str, num_repeats: int, **kwargs):
+    def __init__(self, in_channels: int, skip_channels: int, out_channels: int, up_factor: int, mode: str, num_repeats: int, fallback_mode: str, **kwargs):
         super().__init__(in_channels=in_channels, skip_channels=skip_channels, out_channels=out_channels)
+        if mode in [UpsampleMode.PIXEL_SHUFFLE.value, UpsampleMode.NN_PIXEL_SHUFFLE.value]:
```
BloodAxe, line 96: We have thee duplicate code blocks 

```
 if mode in [UpsampleMode.PIXEL_SHUFFLE.value, UpsampleMode.NN_PIXEL_SHUFFLE.value]:
            # Check if in_channels is divisible by (up_factor ** 2) for pixel shuffle, else fallback to fallback_mode.
            _in_ch = in_channels / (up_factor**2)
            if _in_ch % 1 == 0:
                in_channels = int(_in_ch)
            else:
                mode = fallback_mode
```

Please consider refactoring these into single function to reduce amount of [unnecessary] duplicated code

src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
```diff
@@ -34,8 +35,16 @@ class UpFactorBlock(AbstractUpFuseBlock):
     Ignore Skip features, simply apply upsampling and ConvBNRelu layers.
     """
 
-    def __init__(self, in_channels: int, skip_channels: int, out_channels: int, up_factor: int, mode: str, num_repeats: int, **kwargs):
+    def __init__(self, in_channels: int, skip_channels: int, out_channels: int, up_factor: int, mode: str, num_repeats: int, fallback_mode: str, **kwargs):
         super().__init__(in_channels=in_channels, skip_channels=0, out_channels=out_channels)
+        if mode in [UpsampleMode.PIXEL_SHUFFLE.value, UpsampleMode.NN_PIXEL_SHUFFLE.value]:
+            # Check if in_channels is divisible by (up_factor ** 2) for pixel shuffle, else fallback to fallback_mode.
```
BloodAxe, line 41: Usually if in_channels is not divisible by (up_factor ** 2) one can do 1x1 conv from in_channels  to make it divisible and then do pixel shuffle. Do we want to support this?

src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
lkdci: 1. I agree we can hint a type `Union[str, UpsampleMode]`.  Passing `str` should also be an option IMO, it simplify the yaml init, and it's common usage as torch reduction modes and upsample modes i.e
2. Let's take an example of a module with 4 decoder stages, in case only one of the stages isn't supported for pixel-shuffle, this module is still a valid network and the user might want to use this combination.
We can make this argument `Optional[str] = None`, in case is None, we can raise an axception

src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
lkdci: It isn't that simple, should we project to the closest floor or ceil? imagine for `input channels=2` and `scale_factor=4` does it make sense to project from 2 to 16 channels, to end with 1 channel? 
IMO the user should resolve those issues, and we shouldn't imply this. This is why I suggest a fallback mode that the user is aware of.

src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
lkdci: Changed hit type to `Union[UpsampleMode, str]` [bb14e68](https://github.com/Deci-AI/super-gradients/pull/707/commits/bb14e6877d8e24fb2d7de4d74b91d75b892b496f)

src/super_gradients/training/models/segmentation_models/unet/unet_decoder.py
lkdci: refactored to `validate_upsample_mode` method [bb14e68](https://github.com/Deci-AI/super-gradients/pull/707/commits/bb14e6877d8e24fb2d7de4d74b91d75b892b496f)

lkdci:
@BloodAxe Addressed, see my comments about how to resolve when pixel shuffle isn't supported.

BloodAxe (approved):
LGTM

PR #713, opened 2023-02-13
KD.md

shaydeci:


documentation/source/KD.md
```diff
@@ -0,0 +1,119 @@
...
+## Knowledge Distillation Training in SG: Key Components
+
+
+### [KDModule](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/training/models/kd_modules/kd_module.py)
+
+The most apparent difference in the flow of training using knowledge distillation is the fact that it requires two networks: the "teacher" and the "student".
+The relation between the two is also configurable - for example, we may decide that the teacher model should preprocess the inputs differently.
+For that matter, SG introduces a new `torch.nn.Module` that wraps both the student and the teacher models: `KDModule`.
+Uppon calling `KDTrainer.train()` the teacher and student models are passed, along the `kd_arch_params` to initialize a `KDModule` instance.
+Passing a `KDModule` instance explicitly to `KDTrainer.train()` through the `model` argument instead of student and teacher models is also possible, which gives the users the option to customize KD to their needs.
```
ofrimasad, line 53: it is not so clear what can we use the KDModule for. 
maybe add an example

documentation/source/KD.md
```diff
@@ -0,0 +1,119 @@
...
+
+### [KDOutput](https://github.com/Deci-AI/super-gradients/blob/12a4e53a96e8608409100b5ef83971157518434b/src/super_gradients/training/models/kd_modules/kd_module.py#L7)
+
+`KDOutput` defines the structure of the output of `KDModule`, and has two self-explanatory attributes: student_output, teacher_output.
+`KDTrainer` uses these attributes behind the scenes to perform the usual operations of normal training such as metrics calculations.
+This means that when customizing KD, its important for the custom `KDModule` to stick to this output format.
+
+### KD Losses
+
+Currently, [KDLogitsLoss](https://github.com/Deci-AI/super-gradients/blob/12a4e53a96e8608409100b5ef83971157518434b/src/super_gradients/training/losses/kd_losses.py#L15) is the only in SGs KD losses bank, but more are to come.
```
ofrimasad, line 63: typo: "the only in" => "the only one in"

ofrimasad (commented):
Some comments inline.
I feel like we may need some more examples here.

documentation/source/KD.md
shaydeci: So...currently, there is no use for it since we only apply KD directly to the student and teacher outputs.
This was my design proposal so that in the future, we can customise KD (for example, define a different KDModule that might extract additional intermediate layers) and inherit from it.

documentation/source/KD.md
shaydeci: fixed

documentation/source/KD.md
shaydeci: Added high-level customisation example.

ofrimasad (approved):
LGTM

PR #714, opened 2023-02-15
Update the way we load config file with DeciClient

Louis-Dupont:
## Goal
Homogenize how we load arch_params when working with local models and when downloading from platform.

### DIFF
--- a/src/super_gradients/common/environment/cfg_utils.py
+++ b/src/super_gradients/common/environment/cfg_utils.py
@@ -1,6 +1,8 @@
 import os
 from pathlib import Path
 from typing import List, Optional
+
+import hydra
 import pkg_resources
 
 from hydra import initialize_config_dir, compose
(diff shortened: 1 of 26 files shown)


src/super_gradients/common/environment/cfg_utils.py
```diff
@@ -57,13 +58,51 @@ def add_params_to_cfg(cfg: DictConfig, params: List[str]):
...
+
+
+def load_cfg(config_name, config_type: ConfigType, recipes_dir_path: Optional[str] = None) -> DictConfig:
     GlobalHydra.instance().clear()
-    sg_recipes_dir = pkg_resources.resource_filename("super_gradients.recipes", "")
-    dataset_config = os.path.join("arch_params", config_name)
-    with initialize_config_dir(config_dir=normalize_path(sg_recipes_dir), version_base="1.2"):
+
+    if recipes_dir_path is None:
+        recipes_dir_path = pkg_resources.resource_filename("super_gradients.recipes", "")
```
shaydeci, line 76: Is this always going to be the case? What if a user places their recipe somewhere else ?

My concern is raised since we are talking about DeciCient users.

shaydeci (changes requested):
Just one clarification inline.

src/super_gradients/common/plugins/deci_client.py
```diff
@@ -71,7 +69,9 @@ def _get_file(self, model_name: str, file_name: str) -> Optional[str]:
                 logger.debug(e.body)
             return None
 
-        return FilesDataInterface.download_temporary_file(file_url=download_link)
+        file_path = FilesDataInterface.download_temporary_file(file_url=download_link)
```
BloodAxe, line 72: Do we want to wrap this download in wait_for_master? To ensure we do not over-download and not cause race-condition when writing to the same file?

src/super_gradients/common/plugins/deci_client.py
```diff
@@ -207,12 +218,18 @@ def add_model(
...
+    """
+    src_dir_path = os.path.dirname(src_file_path)
+
+    dest_dir_path = os.path.join(src_dir_path, dest_dir_name)
+    dest_file_path = os.path.join(dest_dir_path, os.path.basename(src_file_path))
 
-    arch_params_file = Path(config_path)
-    with hydra.initialize_config_dir(config_dir=str(arch_params_file.parent), version_base=None):
-        return hydra.compose(config_name=arch_params_file.name)
+    os.makedirs(dest_dir_path, exist_ok=True)
```
BloodAxe, line 233: Same, maybe `wait_for_master` here?

src/super_gradients/common/plugins/deci_client.py
Louis-Dupont: That's a good question.
`FilesDataInterface.download_temporary_file(...)` downloads and saves to a local temp file. This temp file should be different for every node because it is generated by each process separately. So, in the end if I'm not wrong there should be no problem race-condition wise.
It still means that we download n times every file from the platform, but they are very lightweight (it's only yamls) so I'm not sure it is worth optimizing.

Does that make sens to you?

src/super_gradients/common/plugins/deci_client.py
Louis-Dupont: Because evert process has its own file (see comment above) there is no point to `wait_for_master`, right?

shaydeci (approved):
LGTM

PR #715, opened 2023-02-16
Feature/sg 678 centralize and improve how SG loads cfg from yaml

Louis-Dupont:
## Motivation
- Provide a set of utils functions to load a subconfig (arch_params, training_hyperparams, ...) quickly without needing to re-implement almost the same code every time.

## What I did
Implement these functions.
Replace custom code that did something similar, to use the newly created functions.

### Exception added
When the recipe is not found, we will raise an informative exception aiming at helping the user understand what is wrong:
image

The PR https://github.com/Deci-AI/super-gradients/pull/714 is based on this.


src/super_gradients/training/utils/hydra_utils.py
```diff
@@ -1,69 +0,0 @@
-import os
```
shaydeci, line 1: Why did you leave this file if its completely empty ?

shaydeci (changes requested):

shaydeci (approved):
LGTM

PR #724, opened 2023-02-20
add crowd to recipes

Louis-Dupont:
I don't really see a point to keep crowd=False, knowing that it doesnt increase by a lot the running time of validation, AND we report with crowd.


ofrimasad (approved):
LGTM

PR #728, opened 2023-02-20
UNet aux heads customization

lkdci:
Allow setting any encoder output as auxiliary head with no constraints for the heads to be consecutive (as currently supported)

i.e this example is now supported:
```
# for aux head upsample factor
aux_heads_factor: [ 2, 4, 8, 16, 32, 64 ]

# one can set the aux heads to output only stride 8 and 32, as follows (not supported before this PR):
use_aux_list: [False, False, True, False, True, False]

```


Louis-Dupont (approved):
LGTM

ofrimasad (approved):
LGTM

PR #733, opened 2023-02-22
Introduce transfer_weights and StrictLoad.KEY_MATCHING mode

BloodAxe:
This PR adds `StrictLoad.KEY_MATCHING` mode to do partial checkpoint load. This is different from `.load_state_dict(state_dict, strict=False)` in one but critical part - if the layer has matching name, but shape is different, we will just skip this layer and load what has matching shape, ignoring the rest. Vanilla `load_state_dict` would just raise an error.

See https://github.com/Deci-AI/super-gradients/blob/09c0c0d6ca966a84fc320566fa410fc233770bae/tests/unit_tests/load_checkpoint_test.py#L10 test to get the idea what it does.

The motivation for this change is to have an ability to do partial checkpoint load. In case of DEKR, the pretrained weights comping from another repo and NO_KEY_MATCHING load strategy cannot do the work. `NO_KEY_MATCHING` doing index-based loading and it puts very strict requirement on the order of layers stored in checkpoint (side note: IMO it is very, very, very wroong. We had chat with @shaydeci about this and he explained this is to support some old checkpoints. ).
With `KEY_MATCHING` we could use these pretrained checkpoints without issues.

### DIFF
--- a/src/super_gradients/training/utils/checkpoint_utils.py
+++ b/src/super_gradients/training/utils/checkpoint_utils.py
@@ -1,14 +1,17 @@
 import os
 import tempfile
 import pkg_resources
-
+import collections
 import torch
 
 from super_gradients.common.abstractions.abstract_logger import get_logger
 from super_gradients.common.data_interface.adnn_model_repository_data_interface import ADNNModelRepositoryDataInterfaces
 from super_gradients.common.decorators.explicit_params_validator import explicit_params_validation
 from super_gradients.training.pretrained_models import MODEL_URLS
+from super_gradients.common.data_types import StrictLoad
 
+from torch import nn, Tensor
+from typing import Union, Mapping
 
 try:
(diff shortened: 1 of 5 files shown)


src/super_gradients/training/utils/checkpoint_utils.py
```diff
@@ -172,6 +197,9 @@ def load_checkpoint_to_model(
     @param load_weights_only:
     @return:
     """
+    if isinstance(strict, str):
```
shaydeci, line 200: Thanks for being a good citizen (:

shaydeci (approved):
LGTM

src/super_gradients/training/utils/checkpoint_utils.py
BloodAxe: 🤗

PR #734, opened 2023-02-22
Fix the case when strict value could be just a boolean variable.

BloodAxe:
Somehow failing unit test did not marked build as red, so it slipped into master in #733 PR. 
This hotfix address the issue that strict sometimes can be just an ordinary bool


shaydeci (approved):
LGTM

PR #736, opened 2023-02-22
Feature/sg 691 resolve circular loop for loss

Louis-Dupont:
This fixes the issue of circular import on Losses.

This is a breaking change though because register_loss was moved from `registry.py` to `register_loss.py`. I am not sure if there is a simple way around that because:
- In order to avoid we need EVERY register to be accessible in `registery.py`
- In order to avoid the loop, we need to be able to call `register_loss(...)` before loading other registries dependencies.

The solution I would go for is to generalize the usage of this Registry class, and to decorate every model/metric/callback/.... using it. I didnt do it because it is out of scope and it would be a very huge (repetitive) PR, but we might want to consider it because it would allows us to not introduce any breaking change while fixing these circular loops.


src/super_gradients/common/registry/__init__.py
```diff
@@ -1,4 +1,4 @@
-from super_gradients.common.registry.registry import register_model, register_metric, register_loss, register_detection_module, register_lr_scheduler
-
-
-__all__ = ["register_model", "register_detection_module", "register_metric", "register_loss", "register_lr_scheduler"]
+# from super_gradients.common.registry.registry import register_model, register_metric, register_loss, register_detection_module, register_lr_scheduler
+#
+#
+# __all__ = ["register_model", "register_detection_module", "register_metric", "register_loss", "register_lr_scheduler"]
```
Louis-Dupont: This highly contributes to the circular loop... 
We might be able to keep it if we refactor EVERY register function the same way I propose in this PR (see description of PR)

src/super_gradients/common/registry/__init__.py
```diff
@@ -1,4 +0,0 @@
-from super_gradients.common.registry.registry import register_model, register_metric, register_loss, register_detection_module, register_lr_scheduler
-
-
-__all__ = ["register_model", "register_detection_module", "register_metric", "register_loss", "register_lr_scheduler"]
```
Louis-Dupont, line 4: This highly contributes to the circular loop...
We might be able to keep it if we refactor EVERY register function the same way I propose in this PR (see description of PR)

src/super_gradients/common/object_names.py
```diff
@@ -1,22 +1,3 @@
-class Losses:
```
ofrimasad, line 1: I don't understand why we need to remove this class...
it is not importing anything, so it cannot affect circular imports

ofrimasad (changes requested):
This change needs to be revised. The breaking changes are too significant.
Let's discuss alternative solutions.

PR #739, opened 2023-02-22
Feature/sg 576 add regression tests

BloodAxe:


src/super_gradients/training/metrics/pose_estimation_metrics.py
```diff
@@ -269,19 +269,19 @@ def compute(self) -> Dict[str, Union[float, torch.Tensor]]:
 
         predictions = self.predictions  # All gathered by this time
         if len(predictions) > 0:
-            preds_matched = torch.cat([x[0] for x in predictions], dim=0)
-            preds_to_ignore = torch.cat([x[1] for x in predictions], dim=0)
-            preds_scores = torch.cat([x[2] for x in predictions], dim=0)
+            preds_matched = torch.cat([x[0].cpu() for x in predictions], dim=0)
```
shaydeci, line 272: Why are we moving the results from the gpu? I don't recall we do this in other metrics and it seemed to work fine.

shaydeci (changes requested):
1 minor comment inline.

src/super_gradients/training/metrics/pose_estimation_metrics.py
BloodAxe: In detection metrics class this actually happening in `_sync_dist` method: https://github.com/Deci-AI/super-gradients/blob/8992343fdb77a602fe701be17fca6bc8b66c7d82/src/super_gradients/training/metrics/detection_metrics.py#L171

shaydeci (approved):
LGTM

PR #740, opened 2023-02-23
Feature/sg 691 register losses dynamically

Louis-Dupont:
This is the first PR of this type.
I will push similar PRs for every type of objects that we register, with exactly the same logic.

Note: this PR doesnt fully solve the circular import loop that Lior had. In order to fix it, we need to remove the dependencies that EVERY register functions has to the SG code (`ARCHITECTURES`, `CALLBACKS`, ...). 
That being said, this PR is NOT a breaking change, the code still works like before. It just doesnt solve the issue raised by lior yet, but it is a first step toward it.


src/super_gradients/training/losses/__init__.py
```diff
@@ -1,3 +1,6 @@
+from super_gradients.common.registry.registry import LOSSES
+from super_gradients.common.object_names import Losses
+
```
Louis-Dupont, line 3: Note: This will not lead to circular loop and add backward compatibility if someone is importing Losses or LOSSES from this module.

src/super_gradients/training/losses/dekr_loss.py
```diff
@@ -3,7 +3,11 @@
 import torch
 from torch import Tensor, nn
 
+from super_gradients.common.object_names import Losses
+from super_gradients.common.registry.registry import register_loss
 
+
+@register_loss(Losses.DEKR_LOSS)
```
BloodAxe, line 10: Do we want to make one step further and drop `Losses` entirely?

What are the advantages of having
```
@register_loss(Losses.DEKR_LOSS)
class DEKRLoss(nn.Module):
```
and not
```
@register_loss()
class DEKRLoss(nn.Module):
```

Implementation of `register_loss` can infer class name that it is wrapping, so there is no need to repeat ourselves and duplicate class name.

From what I can tell, we don't use this `Losses` class anywhere in SG codebase except for registration. So this should be safe to drop.

Louis-Dupont: That was my first thought, but actually keeping the names allows autocompletion because it is statically defined.
Without, we can always have a function like "list_losses()" that would help debugging, but only on runtime.

BloodAxe: What are the use cases where we need autocompletion? These do not work in YAML files, and we almost never use loss names in code, apart from unit tests.
Just trying to follow DRY principle here.

Louis-Dupont: Ok I got confused, it's true that these names are mainly only for Models and maybe Datasets because we can get the objects using `<module>.get("object_name")`
That being said, removing it would be a breaking change in case someone is using it in a python script.
Example:
```
from super_gradients.common.object_names import Losses
training_params = {
    "loss": Losses.R_SQUARED_LOSS,
    ...:...
}
```

What I think would be great is to keep it only for the objects that can be called with a `get(...)` function and to remove it for every other object, but to not do this in a patch release because of the breaking change it might introduce. On top of that I think it would be good to take this as an opportunity to finally match the object names to the name of the class it maps to (e.g. `"r_squared_loss"` -> `"RSquaredLoss"`)

BloodAxe (approved):
LGTM

PR #741, opened 2023-02-23
fix: Center detection affine

yurkovak:
Fix incorrect behavior of affine transform when applied to images of resolution bigger than the target one


BloodAxe:
LGTM

BloodAxe (approved):

PR #743, opened 2023-02-24
Feature/sg 684 benchmark ppyoloe

BloodAxe:
This is in draft as we need to update PPYolo results (retrain) after @yurkovak fix of affine transforms #741


src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
```diff
@@ -33,9 +33,11 @@ def prep_model_for_conversion(self, input_size: Union[tuple, list] = None, **kwa
         with convertible substitutes and remove all auxiliary or training related parts.
         :param input_size: [H,W]
         """
+        self.head.cache_anchors(input_size)
+
         for module in self.modules():
             if isinstance(module, RepVGGBlock):
-                module.prep_model_for_conversion(input_size)
+                module.fuse_block_residual_branches()
```
shaydeci, line 40: nice catch.

shaydeci (approved):
LGTM

PR #750, opened 2023-02-28
Feature/sg 680 lion optimizer

shaydeci:
Lion experiments produce similar results to AdamW for the PPYOLOE recipe.
I tested it on PPYOLOE  as it is the only recipe in SG currently using AdamW. There has been requests from the Algo team to integrate it regardless of these results.
Further assessment will be done on Segformer recipe as it also uses AdamW, and will soon be introduced to SG as well.

Experiments WandB: https://wandb.research.deci.ai/super-gradients/LionExperiments .


BloodAxe (approved):
LGTM

Louis-Dupont (approved):
LGTM

PR #751, opened 2023-02-28
input_size passed to prep_ model for conversion

shaydeci:
We do not pass explicitly input_size to prep_model_for_conversion upon calling models.convert().
I added to take the ONNX if none is passed through the prep_model_for_conversion_kwargs.


src/super_gradients/training/models/conversion.py
```diff
@@ -74,6 +75,10 @@ def convert_to_onnx(
         raise FileNotFoundError(f"Could not find destination directory {out_path} for the ONNX file.")
     torch_onnx_export_kwargs = torch_onnx_export_kwargs or dict()
     prep_model_for_conversion_kwargs = prep_model_for_conversion_kwargs or dict()
+
+    if get_param(prep_model_for_conversion_kwargs, "input_size") is None:
+        prep_model_for_conversion_kwargs["input_size"] = (1, *input_shape)
```
BloodAxe, line 80: Do we want to check that `input_size` (when it's present in prep_model_for_conversion_kwargs) equals the input_shape?
In other words, that prep_model_for_conversion gets the same spatial dimensions that we use of ONNX expor? Looks like a legit check for me

shaydeci: Sure, added a check in latest commit.

src/super_gradients/training/models/conversion.py
```diff
@@ -74,6 +75,16 @@ def convert_to_onnx(
...
```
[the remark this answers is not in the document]
shaydeci: sure. fixed.

BloodAxe (approved):
LGTM

Louis-Dupont (approved):
LGTM

PR #755, opened 2023-03-01
Feature/sg 672 add onnx simplifier

shaydeci:


requirements.txt
```diff
@@ -35,3 +35,4 @@ stringcase>=1.2.0
 numpy<=1.23
 rapidfuzz
 json-tricks==3.16.1
+onnx-simplifier
```
BloodAxe, line 38: Uhm, maybe we want to add more or less strict version pinning here?
There were instances (mostly in nodejs) when some 3rdparty package was compromised and due to loose version dependency other repos brought malware to customers.

src/super_gradients/training/models/conversion.py
```diff
@@ -80,6 +86,8 @@ def convert_to_onnx(
     complete_model = ConvertableCompletePipelineModel(model, pre_process, post_process, **prep_model_for_conversion_kwargs)
 
     torch.onnx.export(model=complete_model, args=onnx_input, f=out_path, **torch_onnx_export_kwargs)
+    if simplify:
```
BloodAxe, line 89: Function returns the `out_path` regardless of value of `simplify`.
As a user, if I pass `simplify=True`, then my expectation that result of function return value would be path to simplified model. 

IMHO, we should not generate `_simplified.onnx` at all and produce only one ONNX file. What do you think?

src/super_gradients/training/models/conversion.py
```diff
@@ -129,3 +137,15 @@ def convert_from_config(cfg: DictConfig) -> str:
...
+
+
+def onnx_simplify(onnx_path: str, onnx_sim_path: str):
+    """
+    onnx simplifier method, same as `python -m onnxsim onnx_path onnx_sim_path
+    :param onnx_path: path to onnx model
+    :param onnx_sim_path: path for output onnx simplified model
+    """
+    model_sim, check = simplify(model=onnx_path)
+    assert check, "Simplified ONNX model could not be validated"
```
BloodAxe, line 149: 🙏  assert in unit tests, if/raise in library code

Why: Assertions can be toggled on or off at will.

https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/

TLDR: 

> So never use an assert when another method of capturing and raising an error is possible. Always write your code with the assumption that the assertions could disappear at any minute. And if you’re unsure, you can run your program with python -O or by setting PYTHONOPTIMIZE=1 to test if it still works.

shaydeci: Completely right, this is something I always comment about as well, it slipped in :sweat_smile: .

src/super_gradients/training/models/conversion.py
shaydeci: Sure, fixed.

src/super_gradients/training/models/conversion.py
shaydeci: Anyway, fixed.

requirements.txt
shaydeci: Added requirement to be at least 0.3.6 which is what I tested, and an upper limit to be the upcoming major.

src/super_gradients/training/models/conversion.py
```diff
@@ -129,3 +136,16 @@ def convert_from_config(cfg: DictConfig) -> str:
...
     logger.info(f"Successfully exported model at {out_path}")
     return out_path
+
+
+def onnx_simplify(onnx_path: str, onnx_sim_path: str):
+    """
+    onnx simplifier method, same as `python -m onnxsim onnx_path onnx_sim_path
+    :param onnx_path: path to onnx model
+    :param onnx_sim_path: path for output onnx simplified model
+    """
```
Louis-Dupont, line 146: Minor comment: If we return onnx_sim_path we should add typing and :return:
But I'm not sure we need/want to return it because it's just the input params. So maybe we can just remove the return line to make it more clear ?

Louis-Dupont (approved):
LGTM, just left a small note

src/super_gradients/training/models/conversion.py
shaydeci: sure, removed in https://github.com/Deci-AI/super-gradients/pull/755/commits/94af6edeee6ca1705778884db817e11cab32ee40

PR #756, opened 2023-03-01
unet binary segmentation example

lkdci:
Binary segmentation training example of UNet model on the Supervisely person dataset.
Also added arch_params, for vanilla UNet implementation with bilinear upsampling.


src/super_gradients/training/models/segmentation_models/unet/unet.py
```diff
@@ -214,3 +214,15 @@ def __init__(self, arch_params: HpmStruct):
             aux_heads_params=get_param(arch_params, "aux_heads_params"),
             dropout=get_param(arch_params, "dropout", 0.0),
         )
+
+
+class UNet(UNetCustom):
+    """
+    implementation of:
+     "U-Net: Convolutional Networks for Biomedical Image Segmentation", https://arxiv.org/pdf/1505.04597.pdf
+    The upsample operation is done by using bilinear interpolation which is reported to show better results.
```
lkdci, line 223: If we want to support the original Conv transpose version, we'll have to implement a custom up block only for the vanilla UNet.
The original conv transpose use kernel=2 and stride=2 (not recommended since there is no spatial dependencies with adjacent pixels), and out_channels == in_channels // 2

src/super_gradients/recipes/dataset_params/supervisely_persons_dataset_params.yaml
```diff
@@ -1,5 +1,8 @@
+dataset_dir: /data/supervisely-persons
+batch_size: 8
+
 train_dataset_params:
-  root_dir: /data/supervisely-persons
+  root_dir: ${..dataset_dir}
```
Louis-Dupont, line 5: I see you are using ${..dataset_dir} instead of ${dataset_params.dataset_dir} which is the syntax we usually use.
To be honest, I wonder why we are not using this local syntax which seems to be more robust. Do you have any downside of using ${..dataset_dir}  in mind?

src/super_gradients/training/models/segmentation_models/unet/unet_encoder.py
```diff
@@ -122,44 +122,53 @@ def assert_divisible_channels(num_channels: int, steps: int):
...
-    QARepVGG stage with QARepVGGBlock as building block. If `anti_alias=True`, `AntiAliasDownsample` module is used for
-    downsampling.
-    """
-
-    def build_stage(self, in_channels: int, out_channels: int, stride: int, num_blocks: int, anti_alias: bool, **kwargs):
-        blocks = []
-        # Anti alias gaussian down-sampling
-        if anti_alias and stride == 2:
-            blocks.append(AntiAliasDownsample(in_channels, stride))
+        elif max_pool and stride == 2:
```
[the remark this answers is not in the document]
Louis-Dupont: I would go for that if possible. We can always add DownsampleMode and deprecate the current params to not introduce a breaking change.

Also with your current code if `anti_alias==True` and `max_pool==True` the code will continue with `AntiAliasDownsample` and not `MaxPool2d ` without any notice and I guess we don't want this to happen. We could add a warning/exception.

Louis-Dupont (commented):
I like the refacto, it looks clean!
Just check my comment on `ConvBaseStage`.

src/super_gradients/training/models/segmentation_models/unet/unet_encoder.py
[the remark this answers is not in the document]
lkdci: added `make_downsample_module` and `DownSampleMode` enum.

I left `anti_alias` arg for backward compatibility with a deprecation warning. An exception is raised if both `downsample_mode` and `anti_alias` are set by the user.

src/super_gradients/recipes/dataset_params/supervisely_persons_dataset_params.yaml
lkdci: not that I know of :)

lkdci:
Thanks for the review @Louis-Dupont , added `make_downsample_module` and `DownSampleMode` enum as suggested.

Louis-Dupont (approved):
LGTM

PR #758, opened 2023-03-06
Feature/sg 704 pose estimation base model

BloodAxe:
This draft introduce the concept of config-driven model architecture definition for pose estimation module. 
Currently there are two architectures implemented:
* [DDRNet39](https://github.com/Deci-AI/super-gradients/pull/758/files#diff-414cc5a5cb96c99fd9db0a1de3873ed0e517f79d349cafba125c534e7bc65a7a) backbone + DEKR-like head without deformable convolutions (no neck)
* [PP-Yolo-L](https://github.com/Deci-AI/super-gradients/pull/758/files#diff-25f884db6c6e9ab8b18a136b2e9b5e52e1d4ceeae07723beda1a38dfb80d43c4) (backbone + neck) + DEKR-like head without deformable convolutions

This is based on flexible detection architecture building blocks done by @yurkovak and extends this beyond detection task. 

What I do like
* Quite easy to play with model architecture through config

Things I don't quite like:
1. We have to have a new class for each model in order to refer them via `architecture`.
See https://github.com/Deci-AI/super-gradients/pull/758/files#diff-97ddc9c808eb2bf2f415eae735ec6e34eca97c0c6d120d59ec3335925019a0a4 and https://github.com/Deci-AI/super-gradients/pull/758/files#diff-3ac4e0d7467a2d6458dae1c4d01378cf2016268ef470564d3fa3b3ad220251cf. These are essentially same classes.

2. Our models can and should be refactored to have clear backbone + segmentation / classification / detection parts. This is to avoid necessity to write Frankenstein classes like this https://github.com/Deci-AI/super-gradients/pull/758/files#diff-1bca5825e01f5403bc3e4265fc67fd3db8c4cc1eef74acad058768d5837e6b87


src/super_gradients/training/models/segmentation_models/ddrnet_backbones.py
```diff
@@ -0,0 +1,73 @@
...
+
+import torch
+
+from super_gradients.training.models.segmentation_models.ddrnet import DDRNet39
+from super_gradients.training.utils import HpmStruct
+
+__all__ = ["DDRNet39Backbone"]
+
+
+class DDRNet39Backbone(DDRNet39):
```
shaydeci, line 11: Why not inherit from super_gradients.training.models.segmentation_models.ddrnet.BasicDDRBackBone ?
Im feeling a bit unease with the amount of duplicate backbone classes we have.

src/super_gradients/recipes/coco2017_pose_dekr_w32.yaml
```diff
@@ -67,6 +67,17 @@ dataset_params:
 
 training_hyperparams:
   resume: ${resume}
+
```
shaydeci, line 70: Please remove wandb stuff from recipe.

shaydeci (changes requested):
Nice! Some comments are inline.
I think another point that makes an argument for your earlier suggestion regarding taking out the inheriting classes for our models and leaving only the base class is that it would make backbone configuration much easier as well (i.e, instead of adding new classes for each backbone).
I do think this is a big change and now we have a lot on our plate, but we should definitely discuss this.

src/super_gradients/training/models/segmentation_models/ddrnet_backbones.py
BloodAxe: I actually didn't notice it exists 😅 because of quite massive inheritance chain there.
A BasicDDRBackBone has a bunch of it's own problems. We instantiate and put the whole backbone into arch_params o_O in order store it later in base class. 

```
class DDRNet23Slim(DDRNetCustom):
    def __init__(self, arch_params: HpmStruct):
        _arch_params = HpmStruct(**DEFAULT_DDRNET_23_SLIM_PARAMS)
        _arch_params.override(**arch_params.to_dict())
        # BUILD THE BACKBONE AND INSERT TO THE _arch_params
        backbone_layers, _arch_params.additional_layers = _arch_params.layers[:4], _arch_params.layers[4:]
        _arch_params.backbone = BasicDDRBackBone(  <---- HERE
            block=_arch_params.block,
            width=_arch_params.planes,
            layers=backbone_layers,
            input_channels=_arch_params.input_channels,
            layer3_repeats=_arch_params.layer3_repeats,
        )
        super().__init__(_arch_params)
```

src/super_gradients/training/models/segmentation_models/ddrnet_backbones.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    A somewhat frankenstein version of the DDRNet39 model that tries to be a feature extractor module.
+    """
+
+    def __init__(self, arch_params: HpmStruct):
+        super().__init__(arch_params)
+
+        # Delete everything that is not needed for feature extraction
```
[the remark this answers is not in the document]
BloodAxe: Me neither. I'd rather refactor DDRNet file to make it more friendly to using backbone only or construct end-2-end model. However it turns out to be yet another big effort, so it was not included in this PR. 

The proposed version of DDRNet39Backbone is a dirty hack to test the approach, see it works and then do a nice & clean variant after. @ofrimasad I'm really into refactoring DDRNet somewhere in the next sprints.

src/super_gradients/training/models/segmentation_models/ddrnet_backbones.py
shaydeci: Well, I don't seem to see the problem yet - we are talking about inheriting the BasicDDRBackBone and actually its your implementation that inherits from DDRNet itself. What am I missing ?

BloodAxe: Unfortunately it is not that easy to go with BasicDDRBackBone and here's why
1) BasicDDRBackBone  does not have forward() method implemented at all. So we have to add one. Not a big deal as we still have our own forward method anyway
2) BasicDDRBackBone  does not have `layer5` and `spp` layer (They are initialized in derived class). That is more serious as we will have to instantiate them. During construction of layer5 and spp layer we actually do forward() to get number of output channels
```
        self._backbone = backbone
        self._backbone.validate_backbone_attributes()
        out_chan_backbone = self._backbone.get_backbone_output_number_of_channels() <---- HERE
```

To me the right way to get this done - refactor DDRNet entirely to extract more clear backbone class and build DDRnetSegmentation / DDRNetClassification models on top of it. Without this all attempts to use ddrnet backbone would look ugly

shaydeci: mmm to that I agree.
This is a recurring issue, so perhaps we should have a convo about re-designing the base modules to prevent further discussions like this one (for example, we can introduce new base classes and have them inherit alongside with CustomizableDetector from some base class, something like CustomizableModel).
We should consider taking some time for refactoring before we integrate more models into SG. Lets merge this though, as it is clearly out of the scope here...

shaydeci (approved):
LGTM

PR #763, opened 2023-03-07
Added time factor to pre launch callback - batch size selector

shaydeci:
Added 2 different modes + prints detailing the time taking to perform the dataloader iterations (including forward passes).

### DIFF
--- a/src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
+++ b/src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
@@ -9,6 +9,8 @@ from super_gradients import is_distributed
 from super_gradients.common.abstractions.abstract_logger import get_logger
 from super_gradients.training import models
 from torch.distributed import barrier
+import cv2
+import numpy as np
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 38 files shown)


src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
```diff
@@ -104,39 +111,51 @@ def __call__(self, cfg: DictConfig) -> DictConfig:
         tmp_cfg.training_hyperparams.kill_ddp_pgroup_on_end = False
         tmp_cfg.pre_launch_callbacks_list = []
 
+        fastest_batch_time = np.inf
```
BloodAxe, line 114: I had to read couple time the logic to catch what is happening and when batch size is increased actually. I think it would be great to add to a docstring to this function a description of the logic and what are the exit criterions. 
Currently there are two return statements from the `while True` loop. Maybe it is possible to refactor the code to have exactly one return point?

src/super_gradients/training/pre_launch_callbacks/pre_launch_callbacks.py
```diff
@@ -104,39 +111,51 @@ def __call__(self, cfg: DictConfig) -> DictConfig:
...
                         raise e
                     else:
-                        logger.info(f"Ran out of memory for {curr_batch_size}, setting batch size to {curr_batch_size - self.size_step}.")
-                        self._adapt_lr_if_needed(cfg, found_batch_size=curr_batch_size - self.size_step)
-                        cfg.dataset_params.train_dataloader_params.batch_size = curr_batch_size - self.size_step
-                        self._clear_model_gpu_mem(model)
+                        selected_batch_size = curr_batch_size - self.size_step if self.mode == "largest" else fastest_batch_size
+                        msg = f"Ran out of memory for {curr_batch_size}, setting batch size to {selected_batch_size}."
+                        self._inject_selected_batch_size_to_config(cfg, model, msg, selected_batch_size)
                         return cfg
```
BloodAxe, line 137: Can we do `return self._inject_selected_batch_size_to_config(cfg, model, msg, selected_batch_size)` instead?

BloodAxe (approved):
LGTM

PR #765, opened 2023-03-08
Bug/sg 709 default collate older torch version bug

shaydeci:


src/super_gradients/training/utils/detection_utils.py
```diff
@@ -13,7 +15,6 @@
 import torchvision
 from omegaconf import ListConfig
 from torch import nn
-from torch.utils.data._utils.collate import default_collate
```
BloodAxe, line 16: Quick question:
Why don't we do import try/catch instead?

```
try:
   from torch.utils.data._utils.collate import default_collate
except ImportError:
   from torch.utils.data import default_collate
```

BloodAxe: I personally import from `from torch.utils.data.dataloader import default_collate` and haven't had to change that import for literally years :)

My question was more about whether we want to have our version of collate at all.
+ If fixes import problem
- We have to test it
- We have to support it (Meaning of there is update of torch we should check whether it is the same for every new release).

shaydeci: ahhh ok torch.utils.data.dataloader (:

You wrote down torch.utils.data which didn't work for me with 1.10, the other import seems to work fine so ill close this PR down.

PR #767, opened 2023-03-08
Feature/sg 700 fix params format - Part 1

Louis-Dupont:
This PR aims at improving the typing  and docstring to sphinx format.
I will divide this into 2-3 PRs to not make it too heavy.


ofrimasad (approved):
WOW! didn't even realised it was that much...
great work @Louis-Dupont

PR #769, opened 2023-03-09
Add Roboflow100

Louis-Dupont:
Basically, COCOLikeDetectionDataset is an abstraction of COCODetectionDataset, which allows us to easily work with any dataset with coco-like format (such as roboflow100).
I also added a few minor chagnes to the DetectionDataset


src/super_gradients/recipes/dataset_params/roboflow_detection_dataset_params.yaml
```diff
@@ -0,0 +1,95 @@
...
+  class_inclusion_list:
+  max_num_samples:
+  with_crowd: False
+
+train_dataloader_params:
+  shuffle: True
+  batch_size: 16
+  num_workers: 0
+  batch_sampler: True
+  sampler:
```
BloodAxe, line 56: No InfiniteSampler 🙏 🙏 🙏

PR #770, opened 2023-03-09
Feature/sg 737 register model in deci plat

Louis-Dupont:
# Description of the task
- If there is a training that begins with `deci_platform_sg_logger` -> we need to register it + add register architecture
- Upload all the local files (all relevant logs & checkpoints) for it (training tab 2.1. blocker to SG next version)

## Registration
There are 3 scenarios:

### 1.
The user sets `sg_logger_params.model_name`, regardless of how the model was loaded. Then, the value of `sg_logger_params.model_name` will be used to register the model in the platform.

### 2.
The user does not set `sg_logger_params.model_name`, but loads the model using `models.get("<model-name>")`. Then, the "<model-name>" will be used to register the model in the platform.

### 3.
The user does not set `sg_logger_params.model_name`, and the model was not loaded the model using `models.get("<model-name>"`. Then, an exception is raised:
```
ValueError: `model_name` is required to use `training_hyperparams.sg_logger="deci_platform_sg_logger"`.
Please set `training_hyperparams.sg_logger_params.model_name="<your-model-name>"`.
Note that specifying `model_name` is not required when the model was loaded using `models.get(...)`.
```

## Upload of local files
The requirement says: "*Upload all the local files (all relevant logs & checkpoints) for it (training tab 2.1. blocker to SG next version)*".
@ofrimasad  Isnt it what we already do when calling `sg_logger.upload()`?
https://github.com/Deci-AI/super-gradients/blob/9fd66386ef607e1ece4c09cd6a5884c9fd4a9a72/src/super_gradients/common/sg_loggers/deci_platform_sg_logger.py#L70-L85


src/super_gradients/common/plugins/deci_client.py
```diff
@@ -173,6 +173,14 @@ def register_experiment(self, name: str, model_name: str, resume: bool):
...
         :param model_name:  Name of the model architecture to connect the experiment to
         """
+        try:
+            self.lab_client.register_user_architecture(BodyRegisterUserArchitecture(architecture_name=model_name))
+        except ApiException as e:
+            if e.status == 422:
+                logger.debug(f"The model was already registered, or validation error: {e.body}")
+            else:
+                raise e
+
```
Louis-Dupont, line 183: Question: Is this how we want to handle the exceptions?
Do we want only debug? Do we want to cover more exceptions status? Any other important thing to think of?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1562,13 +1563,29 @@ def _initialize_sg_logger_objects(self, additional_configs_to_log: Dict = None):
...
+                if sg_logger_params.get("model_name") is None:
+                    # Use the model name used in `models.get(...)` if relevant
+                    sg_logger_params["model_name"] = get_model_registration_name(self.net.module)
+
+                if sg_logger_params["model_name"] is None:
+                    raise ValueError(
+                        f'`model_name` is required to use `training_hyperparams.sg_logger="{sg_logger}"`.\n'
+                        'Please set `training_hyperparams.sg_logger_params.model_name="<your-model-name>"`.\n'
+                        "Note that specifying `model_name` is not required when the model was loaded using `models.get(...)`."
+                    )
```
Louis-Dupont, line 1586: I don't like using "model_name" in get_callable_param_names(sg_logger_cls.__init__) but I feel like this is the least dirty solution we can come up with current design.
Another alternative would be to add *args, **kwargs to the SgLoggers but this means that it would silently ignore errors in the usage and/or typos. And adding a kwargs check increases code complexity. So, I think it's actually even more dirty.
If you have an opinion/idea on that I'm Interested

src/super_gradients/common/plugins/deci_client.py
ofrimasad: yes. this is ok

ofrimasad (approved):
LGTM

PR #771, opened 2023-03-09
Add entry point train_from_recipe

BloodAxe:
This PR add a bit more simple way to train from recipe when SG is pip installed to:
`python -m super_gradients.train_from_recipe --config-name=CONFIG_NAME_OF_YOUR_DREAM`

There is a bunch (100+) of examples using `python src/super_gradients/examples/train_from_recipe_example/train_from_recipe.py ....`. 
Do we want to update them all to use new way to launch training or maybe add it as an alternative command? @ofrimasad @shaydeci @Louis-Dupont WDYT?


shaydeci (approved):
LGTM.

IMO, lets keep the other entry point for now and add a ticket so that we do it more thoroughly (perhaps include the convert_recipe, evaluate and resume as well...this adds up to some more work)

PR #773, opened 2023-03-12
fix base_keypoints to support torch version lower than 1.11

ofrimasad:


BloodAxe:
@ofrimasad we merged this fix https://github.com/Deci-AI/super-gradients/pull/772/files on friday which is known to work even on 1.8 (We tested it intentionally on version even lower than in our requirements).
So bottom line - you probably wan to update your import to "from torch.utils.data.dataloader import default_collate"

PR #776, opened 2023-03-13
Minor fixes and changes on detection dataset

Louis-Dupont:
Mainly adding verbose option (this is usefull when loading roboflow100, otherwise you end up with 100 x num_device tqdm)
Also, fixing plot with new tensor type


ofrimasad (commented):
Minor comments.

ofrimasad (approved):
LGTM

PR #781, opened 2023-03-14
Improve test for KeypointsRandomAffineTransform

BloodAxe:
Instead of using randomly generated keypoints we use keypoints covering entire image.
This way test is deterministic w.r.t. to input


Louis-Dupont (approved):
LGTM

PR #786, opened 2023-03-16
Update regnet.py

eran-deci:
a small refactoring to the RegNet head, to allow conversion ONNX-->Keras via onnx2tflite which currently doesn't support the shape op. Also reduces a few nodes in the ONNX graph.


Louis-Dupont (approved):
LGTM

PR #791, opened 2023-03-20
Feature/sg 729 add roboflow100 testing script no first index

Louis-Dupont:
Introducing Roboflow100 (https://blog.roboflow.com/roboflow-100/) to SuperGradients, along with 2 base recipes to train YoloX or PPYoloE


src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -714,6 +718,30 @@ def apply_stage_change(self, context: PhaseContext):
...
+class RoboflowResultCallback(Callback):
+    """Append the training results to a csv file. Be aware that this does not fully overwrite the existing file, just appends."""
+
+    def __init__(self, dataset_name: str, output_path: Optional[str] = None):
+        """
+        :param dataset_name:    Name of the dataset that was used to train the model.
+        :param output_path:     Full path to the output csv file. By default, save at 'checkpoint_dir/results.csv'
+        """
+        self.dataset_name = dataset_name
+        self.output_path = output_path or os.path.join(get_project_checkpoints_dir_path(), "results.csv")
```
BloodAxe, line 731: The code should respect the case when `get_project_checkpoints_dir_path` returns None (We had this issue in Docker for out-of-source run)

BloodAxe (dismissed):

src/super_gradients/training/datasets/detection_datasets/coco_format_detection.py
```diff
@@ -68,7 +72,7 @@ def _setup_data_source(self) -> int:
         """
 
         self.coco = self._init_coco()
-        self.class_ids = sorted(self.coco.getCatIds())
+        self.class_ids = sorted(cls_id for cls_id in self.coco.getCatIds() if cls_id not in self.class_ids_to_ignore)
```
Louis-Dupont, line 75: Do we want this feature?
It is useful for roboflow because we want to drop index 0, but at the same time it is not common at all so I feel like we are kind of overfitting the base class to the need of the roboflow class.
But I also don't have a better solution, so maybe we should just keep this feature like that.

src/super_gradients/examples/evaluate_roboflow100_from_recipe/__init__.py
Louis-Dupont, line 1: Do we want to keep this example folder ?
From what I know @BloodAxe you are not using it and work directly through wandb.
It can be nice to have, but do we want to maintain?

src/super_gradients/training/utils/callbacks/callbacks.py
Louis-Dupont: Same as previous comment, do we want to keep this?
If we don't use/want to use it might be better to take it out.

src/super_gradients/training/datasets/detection_datasets/coco_format_detection.py
BloodAxe: Probably we can keep it here for now. Dropping some classes or merging classes (Are recently requested) could be helpful feature to have. So we may get back to this in future.

PR #804, opened 2023-03-28
Feature/sg 747 add preprocessing

Louis-Dupont:
The way the processing classes work is that they support processing of image and post processing of predictions, which is all that is required in inference.


src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,183 @@
...
+
+@dataclass
+class RescaleMetadata(ProcessingMetadata):
+    original_size: Tuple[int, int]
+    sy: float
+    sx: float
+
+
+@dataclass
+class DetectionPaddedRescaleMetadata(ProcessingMetadata):
```
BloodAxe, line 42: IMO I find DetectionPaddedRescale transform as a very confusing class - is it padding first, or then rescale or vice versa?
Since we already have `RescaleMetadata` and `DetectionPadToSizeMetadata` we can easily describe equivalent postprocessing transform as a composition of these two steps instead of this one. 
I think this would be more clear in terms of order of applied transformations and would not introduce new [unnecessary] entity.

Louis-Dupont: I took it out. I added `DetectionSidePadding` (because the padding of `DetectionPadToSizeMetadata` is center and not `DetectionPaddedRescale`) and also an option `keep_aspect_ratio` in `_Rescale` to preserve the ratio or not (`Rescale` used to it, while `DetectionSidePadding` keeps it).

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,186 @@
...
+@dataclass
+class DetectionPadToSizeMetadata(ProcessingMetadata):
+    shift_w: float
+    shift_h: float
+
+
+@dataclass
+class RescaleMetadata(ProcessingMetadata):
+    original_size: Tuple[int, int]
+    sy: float
```
BloodAxe, line 36: 1) Please add the docstring to this class that explains meaning of sx/sy parameters - are they represent original_size / preprocessed_size or vice-versa?
2) original_size is rows,cols or width,height? This should be also explicitly stated everywhere to avoid confusion of users. For sake of consistency I suggest to use term "shape" for rows,cols and "size" for width, height.

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,186 @@
...
+    :param pad_value:   Padding value for image
+    """
+
+    def __init__(self, output_size: Tuple[int, int], pad_value: int):
+        self.output_size = output_size
+        self.pad_value = pad_value
+
+    def preprocess_image(self, image: np.ndarray) -> Tuple[np.ndarray, DetectionPadToSizeMetadata]:
+        shift_h, shift_w, pad_h, pad_w = _get_center_padding_params(input_size=image.shape, output_size=self.output_size)
+        processed_image = _shift_image(image, pad_h, pad_w, self.pad_value)
```
BloodAxe, line 125: Looks like _shift_image does top-left and not center padding.
shift_h, shift_w are not used in _shift_image at all.

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,186 @@
...
+            scale_factor = min(sy, sx)
+            sy, sx = (scale_factor, scale_factor)
+            rescale_shape = (int(image.shape[0] * sx), int(image.shape[1] * sy))
+
+        rescaled_image = _rescale_image(image, target_shape=rescale_shape)
+
+        return rescaled_image, RescaleMetadata(original_size=image.shape[:2], sy=sy, sx=sx)
+
+
+class DetectionRescale(_Rescale):
```
BloodAxe, line 179: I think we also need LongestMaxSizeRescale than preserves aspect ratio and resize image that it's longest size is of desired size.

src/super_gradients/training/transforms/processing.py
Louis-Dupont: pad_h is actually a tuple that includes the shift (i.e. padding_top), but also the padding to add at the bottom of the image. Same logic applies to pad_w.
I updated the _shift_image function with a more explicit docstring
```
    :param pad_h:       Tuple of (padding_top, padding_bottom).
    :param pad_w:       Tuple of (padding_left, padding_right).
```
Is it enough, or you think it's still confusing? Any idea on how to improve in the second case?

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class Processing(ABC):
+    @abstractmethod
+    def preprocess_image(self, image: np.ndarray) -> Tuple[np.ndarray, Union[None, ProcessingMetadata]]:
+        """Processing an image, before feeding it to the network. Expected to be in (H, W, C) or (H, W)."""
+        pass
+
+    @abstractmethod
+    def postprocess_predictions(self, predictions: np.ndarray, metadata: Union[None, ProcessingMetadata]) -> np.ndarray:
```
BloodAxe, line 47: predictions is [N,4] or [N,4+]? what is the format of bboxes here?
I suggest to require exactly [N,4] of specific format. Since we apply spatial transformations only on box coordinates not on labels and scores there is no need to propagate this information.
So we probably can simplify the API by explicitly saying `predictions` -> `bboxes` has shape of [N,4] of xyxy format.

src/super_gradients/training/transforms/processing.py
Louis-Dupont: 1. I renamed to `scale_factor_h` and `scale_factor_w`. I believe this is now self-explanatory, do you agree or still prefer a docstring?
2. I agree, I actually forgot to do this cleanup thanks for noticing. Just to make sure, when you write (rows,cols), you means (n_rows, n_cols) right ? Which is equivalent to (H, W) and is basically how it is stored in np.ndarray.

src/super_gradients/training/transforms/processing.py
Louis-Dupont: I agree that this function doesn't need to know about the whole prediction, that's a good point.
The thing is that it moves the responsibility to the processing pipeline to handle this (to know how to get bbox, label, ...).
The current implementation of the pipeline is (almost) fully task agnostic which I believe is the right way of doing, and to do so it relies on the processing functions to handle the predictions themselves. It also relies on the postprocessing step to return targets in xyxy_score_label format.


The better solution I can think of is making predictions an instance of a class that would abstract away how to get the bbox, score, label. This would move away the responsibility of knowing how to get the target/score/label away from both pipeline and the processing function.
This abstraction can be useful in different contexts but are we sure we want it ? In particular, this would impact the functions that are in utils which are also being used in `transforms.py`, so it looks like it would be a big refacto with a risk of breaking change to handle.

src/super_gradients/training/transforms/utils.py
```diff
@@ -0,0 +1,132 @@
...
+        - shift_h:  Horizontal shift.
+        - shift_w:  Vertical shift.
+        - pad_h:    Horizontal padding.
+        - pad_w:    Vertical padding.
+    """
+    pad_h, pad_w = output_size[0] - input_size[0], output_size[1] - input_size[1]
+    shift_h, shift_w = pad_h // 2, pad_w // 2
+    pad_h = (shift_h, pad_h - shift_h)
+    pad_w = (shift_w, pad_w - shift_w)
+    return shift_h, shift_w, pad_h, pad_w
```
BloodAxe, line 49: maybe `pad_top, pad_left, pad_right, pad_bottom` instead `shift_h, shift_w, pad_h, pad_w` ? I feel this has better clarity to which side this offset is applied

Louis-Dupont: Changed in 0cb58e216df83afe13f6123edbee9f33d39dba05

tests/unit_tests/transforms_test.py
```diff
@@ -120,6 +131,96 @@ def test_detection_pad_to_size(self):
...
+    def test_rescale_image(self):
+        image = np.random.randint(0, 256, size=(640, 480, 3), dtype=np.uint8)
+        target_shape = (320, 240)
+        rescaled_image = _rescale_image(image, target_shape)
+
+        # Check if the rescaled image has the correct target shape
+        self.assertEqual(rescaled_image.shape[:2], target_shape)
+
+    def test_rescale_bboxes(self):
+        bboxes = np.array([[10, 20, 50, 60, 1], [30, 40, 80, 90, 2]], dtype=np.float32)
```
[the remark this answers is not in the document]
Louis-Dupont: 1a32cf2196e47fc0e80fd0fee08681e6428fe9d0

src/super_gradients/training/transforms/processing.py
BloodAxe: 💯 `DetectionPrediction` ❤️ ❤️ ❤️

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,193 @@
...
+        pad_top, pad_bot, pad_left, pad_right = _get_center_padding_params(input_shape=image.shape, output_shape=self.output_shape)
+        processed_image = _pad_image(image, (pad_top, pad_bot), (pad_left, pad_right), self.pad_value)
+
+        return processed_image, DetectionPadToSizeMetadata(pad_top=pad_top, pad_left=pad_left)
+
+    def postprocess_predictions(self, predictions: np.ndarray, metadata: DetectionPadToSizeMetadata) -> np.ndarray:
+        return _shift_bboxes(targets=predictions, shift_h=-metadata.pad_top, shift_w=-metadata.pad_left)
+
+
+class DetectionSidePadding(Processing):
```
BloodAxe, line 140: How is it different from DetectionCenterPadding? I assume it pad image from the right & bottom?
If so, there should be a docstring reflecting this.
Also perhaps we can simplify these two classes into this to avoid duplication?:
```
@dataclass
class PaddingParams:
    pad_top: int
    pad_left: int
    pad_bottom: int
    pad_right: int


class DetectionPadding(Processing):
    def __init__(self, output_shape: Tuple[int, int], pad_value: int):
        self.output_shape = output_shape
        self.pad_value = pad_value

    def preprocess_image(self, image: np.ndarray) -> Tuple[np.ndarray, DetectionPadToSizeMetadata]:
        padding_params = _get_center_padding_params(input_shape=image.shape, output_shape=self.output_shape)
        processed_image = _pad_image(image, padding_params, self.pad_value)

        return processed_image, DetectionPadToSizeMetadata(padding_params)

    def postprocess_predictions(self, predictions: np.ndarray, metadata: DetectionPadToSizeMetadata) -> np.ndarray:
        return _shift_bboxes(targets=predictions, padding_params)
    
    @abc.abstractmethod
    def _get_padding_params(self, input_shape) -> PaddingParams:
        pass


class DetectionCenterPadding(DetectionPadding):
    def __init__(self, output_shape: Tuple[int, int], pad_value: int):
        self.output_shape = output_shape
        self.pad_value = pad_value

    def _get_padding_params(self, input_shape):
        return _get_center_padding_params(input_shape=image.shape, output_shape=self.output_shape)


class DetectionBottomRightPadding(DetectionPadding):
    def __init__(self, output_shape: Tuple[int, int], pad_value: int):
        self.output_shape = output_shape
        self.pad_value = pad_value

    def _get_padding_params(self, input_shape):
        return _get_bottom_right_padding_params(input_shape=image.shape, output_shape=self.output_shape)

```

src/super_gradients/training/transforms/processing.py
Louis-Dupont: Something like this ? I just wrote it quickly, didnt test and didnt add it to the code
```

@dataclass
class DetectionPredictions(Prediction):
    bboxes: np.ndarray
    confidence: np.ndarray
    labels: np.ndarray

    bbox_format: str
    image_shape: Tuple[int, int]

    @property
    def bboxes_cxcywh(self):
        return self._get_bbox_as("cxcywh")

    @bboxes_cxcywh.setter
    def bboxes_cxcywh(self, bboxes: np.ndarray):
        self._set_bbox_from(bboxes=bboxes, input_bbox_format="cxcywh")

    @property
    def bboxes_xywh(self):
        return self._get_bbox_as("xywh")

    @bboxes_xywh.setter
    def bboxes_xywh(self, bboxes: np.ndarray):
        self._set_bbox_from(bboxes=bboxes, input_bbox_format="xywh")

    @property
    def bboxes_normalized_xyxy(self):
        return self._get_bbox_as("normalized_xyxy")

    @bboxes_normalized_xyxy.setter
    def bboxes_normalized_xyxy(self, bboxes: np.ndarray):
        self._set_bbox_from(bboxes=bboxes, input_bbox_format="normalized_xyxy")

    def _get_bbox_as(self, desired_bbox_format: str):
        return convert_bboxes(
            bboxes=self.bboxes,
            image_shape=self.image_shape,
            source_format=factory.get(self.bbox_format),
            target_format=factory.get(desired_bbox_format),
            inplace=False,
        )

    def _set_bbox_from(self, bboxes: np.ndarray, input_bbox_format: str):
        self.bboxes = convert_bboxes(
            bboxes=bboxes,
            image_shape=self.image_shape,
            source_format=factory.get(input_bbox_format),
            target_format=factory.get(self.bbox_format),
            inplace=False,
        )

    def _set_default_bbox_format(self, desired_format: str):
        self.bbox_format = desired_format
        self.bboxes = self._get_bbox_as(desired_format)

```

src/super_gradients/training/transforms/processing.py
Louis-Dupont: DOne something similar

https://github.com/Deci-AI/super-gradients/blob/3229c5447e8af8ffc44550fb7245670c8c2ef430/src/super_gradients/training/transforms/processing.py#L152-L204

src/super_gradients/training/transforms/processing.py
Louis-Dupont: @BloodAxe This is what I propose, what do you think ? b012d46148801cdbf2d835fec4e3f47abb4482aa

src/super_gradients/training/models/predictions.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def bboxes_xyxy(self):
+        return self._get_bbox_as("xyxy")
+
+    @bboxes_xyxy.setter
+    def bboxes_xyxy(self, bboxes: np.ndarray):
+        self._set_bbox_from(bboxes=bboxes, input_bbox_format="xyxy")
+
+    @property
+    def bboxes_cxcywh(self):
```
BloodAxe, line 49: A getter and setter for bboxes_cxcywh is not used anywhere and can (and should IMO) be removed.
You don't need this format for visualization and if someone needs boxes in format other than XYXY he/she can use convert_bboxes call outside the DetectionPrediction. IMO this class should not be responsible to give boxes to user in custom formats.

src/super_gradients/training/models/predictions.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, bboxes: np.ndarray, bbox_format: str, confidence: np.ndarray, labels: np.ndarray, image_shape: Tuple[int, int]):
+        """
+        :param bboxes:      BBoxes in the format specified by bbox_format
+        :param bbox_format: BBoxes format that can be a string ("xyxy", "cxywh", ...)
+        :param confidence:  Confidence scores for each bounding box
+        :param labels:      Labels for each bounding box
+        :param image_shape: Shape of the image the prediction is made on
+        """
+        self._bboxes = bboxes
+        self._bbox_format = bbox_format
```
BloodAxe, line 35: Why don't we directly convert and store boxes in XYXY format?
For me it sounds perfectly fine to drop the original predictions format from the model and narrow the scope only to XYXY.

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,202 @@
...
+    :param output_shape: (H, W)
+    """
+
+    def __init__(self, output_shape: Tuple[int, int]):
+        self.output_shape = output_shape
+
+    def preprocess_image(self, image: np.ndarray) -> Tuple[np.ndarray, RescaleMetadata]:
+
+        scale_factor = min(self.output_shape[0] / image.shape[0], self.output_shape[1] / image.shape[1])
+        rescale_shape = (int(image.shape[0] * scale_factor), int(image.shape[1] * scale_factor))
```
BloodAxe, line 187: There are extremely rare but possible cases when
int(scale * actual_size) != requested_longest_size.
This trick ensures this is not happening:
https://github.com/Deci-AI/super-gradients/blob/737a50080ab4b803b16ab91b328c0e437b162de7/src/super_gradients/training/transforms/keypoint_transforms.py#L224

src/super_gradients/training/models/predictions.py
Louis-Dupont: Sounds reasonable

src/super_gradients/training/transforms/processing.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class _LongestMaxSizeRescale(Processing, ABC):
+    """Resize image to given image dimensions WITH preserving aspect ratio.
+
+    :param output_shape: (H, W)
+    """
+
+    def __init__(self, output_shape: Tuple[int, int]):
+        self.output_shape = output_shape
```
BloodAxe, line 182: Why _LongestMaxSizeRescale needs output_shape of two values as argument?
The longest size is a single scalar, and having a need to pass two arguments is a bit confusing why one have to pass two values here.

src/super_gradients/training/transforms/processing.py
Louis-Dupont: Why not use round?
`rescale_shape = (round(image.shape[0] * scale_factor), round(image.shape[1] * scale_factor))`
or
`new_height, new_width = tuple(round(dim * scale) for dim in (height, width)) `

src/super_gradients/training/transforms/processing.py
Louis-Dupont: Not sure what you mean here.
We need both because they might be different.
Example:
```
image_shape = (320, 480)
output_shape = (960, 960) # Rescaled shape = (640, 960)
output_shape = (640, 960) # Rescaled shape = (640, 960)
output_shape = (640, 640) # Rescaled shape = (426, 640)
```
It also seems to be what you did with keypoints
https://github.com/Deci-AI/super-gradients/blob/90f076e91a29c17fbf3e4c763a70701318e27c4a/src/super_gradients/training/transforms/keypoint_transforms.py#L185-L200

No?

src/super_gradients/training/models/predictions.py
Louis-Dupont: 35717803ed9636d6a04d886010f078f532783c1b
I am not sure about bboxes naming.
I tend to prefer to be explicit about format (`bboxes_xyxy`) because I see people missing that we convert the bbox to xyxy inside the init.

src/super_gradients/training/transforms/processing.py
Louis-Dupont: 7b73edbd3afb3d74be34b5df533447efe289cc0a

src/super_gradients/training/models/predictions.py
Louis-Dupont: I replied about this in another comment

BloodAxe (approved):
LGTM

src/super_gradients/training/transforms/processing.py
BloodAxe: round would also work here i beleive

src/super_gradients/training/models/predictions.py
BloodAxe: 👍 for `bboxes_xyxy` name

PR #807, opened 2023-03-29
Feature/sg 747 add full pipeline with preprocessing

Louis-Dupont:
Includes code from https://github.com/Deci-AI/super-gradients/pull/804, so https://github.com/Deci-AI/super-gradients/pull/804 needs to be merged first


src/super_gradients/training/models/detection_models/yolo_base.py
```diff
@@ -408,6 +421,29 @@ def __init__(self, backbone: Type[nn.Module], arch_params: HpmStruct, initialize
...
             self._initialize_module()
 
+        self._image_processor = ComposeProcessing(
+            [
+                DetectionRescale((640, 640), keep_aspect_ratio=True),
+                DetectionSidePadding((640, 640), 114),
+                ImagePermute((2, 0, 1)),
+            ]
+        )
+        self._class_names = COCO_DETECTION_CLASSES_LIST
```
shaydeci, line 431: We should not hardcode default _class_names.

shaydeci (changes requested):
Writing down the main concern we have discussed together:
- Hard coding the default _image_processor and _class_names is bad practice, and might lead users who do not use these (which will probably happen - any time they do not train on coco). Lets add setters that will by the pretrained_weights str set these attributes accordingly. 
- Please also add a ticket for making it possible to set them from the recipe itself - so that when users complete their training they can use predict right away.

Dont forget to add the implementation for DeciYolo.

src/super_gradients/training/models/detection_models/yolo_base.py
Louis-Dupont: Changes this and _image_processor logic in f3db9559cff93e54e3a403dd2eae12d4bb88f908

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -0,0 +1,145 @@
...
+    @abstractmethod
+    def __call__(self, images: Union[ImageType, List[ImageType]]) -> Results:
+        """Apply the pipeline on images and return the result.
+
+        :param images:  Single image or a list of images of supported types.
+        :return         Results object containing the results of the prediction and the image.
+        """
+        pass
+
+    def _run(self, images: Union[ImageType, List[ImageType]]) -> Tuple[List[np.ndarray], List[Any]]:
```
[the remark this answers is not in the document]
Louis-Dupont: Changed in 8e837a86043ea32c89874ef0a8083a0d30767d3b

Louis-Dupont:
@shaydeci I also updated the `Results`  classes to work with the changes that were introduced in the previous PR, so you can check it out now src/super_gradients/training/models/results.py

src/super_gradients/training/models/model_factory.py
```diff
@@ -135,6 +136,9 @@ def instantiate_model(
                 net.replace_head(new_num_classes=num_classes_new_head)
                 arch_params.num_classes = num_classes_new_head
 
+            class_names, image_processor = get_pretrained_processing_params(model_name, pretrained_weights)
+            net.set_dataset_processing_params(class_names, image_processor)
```
Louis-Dupont, line 140: Not a very big fan of this, because it is hard coding the "processing_params" as being `class_names` and `image_processor`.
But at the same time this might be all we will ever need, I feel like creating an abstraction for this would be overkill.
What do you think ?

src/super_gradients/training/models/sg_module.py
```diff
@@ -62,3 +63,10 @@ class to implement.
         """
 
         raise NotImplementedError
+
+    def predict(self, images, *args, **kwargs) -> Result:
+        raise NotImplementedError(f"`predict` is not implemented for {self.__class__.__name__}.")
+
+    def set_dataset_processing_params(self, *args, **kwargs) -> None:
+        """Set the processing parameters for the dataset."""
+        pass
```
Louis-Dupont, line 72: @shaydeci 
Not a big fan of having `pass` as default.
BUT with the design we discussed we want to apply `set_dataset_processing_params` to every dataset (see `net. set_dataset_processing_params=...` in `models.get`)
Because of that, having an abstract class or Raising `NotImplementedError` would make all other models crash...

An alternative solution could be to simply implement a default `set_dataset_processing_params` behavior (with `class_names` and `image_processor`)
OR to raise `NotImplementedError` and do try/except, but it looks even worse imo

src/super_gradients/training/models/detection_models/yolo_base.py
```diff
@@ -408,6 +416,36 @@ def __init__(self, backbone: Type[nn.Module], arch_params: HpmStruct, initialize
...
+        return YoloPostPredictionCallback(conf=conf, iou=iou)
+
+    def set_dataset_processing_params(self, class_names: Optional[List[str]], image_processor: Optional[Processing]) -> None:
+        """Set the processing parameters for the dataset.
+
+        :param class_names:     (Optional) Names of the dataset the model was trained on.
+        :param image_processor: (Optional) Image processing objects to reproduce the dataset preprocessing used for training.
+        """
+        self._class_names = class_names or self._class_names
+        self._image_processor = image_processor or self._image_processor
```
Louis-Dupont, line 433: @shaydeci Also, not a big fan of having this duplicate (this is implemented in both yolobase and ppyoloe).
But they don't share a parent besides SgModule, so unless we implement this for SgModule or if we implement such an abstract  (which you and Ofri seemed to not want for now) we have to duplicate the code

PR #812, opened 2023-03-31
Fix YoloX loss to handle negative batch correctly

BloodAxe:
When no targets were present in batch, the loss failed to handle it correctly


shaydeci (approved):
LGTM

PR #815, opened 2023-04-02
Feature/sg 747 Adding utils to load/save videos

Louis-Dupont:
This is the first part of the feature to support videos


shaydeci (dismissed):
LGTM

src/super_gradients/training/utils/videos.py
```diff
@@ -0,0 +1,105 @@
...
+
+    for frame in frames:
+        if frame.ndim == 2:
+            frame = frame[:, :, np.newaxis]
+        video_writer.write(frame)
+
+    video_writer.release()
+
+
+def _validate_frames(frames: List[np.ndarray]) -> Tuple[float, float]:
```
BloodAxe, line 84: Why do we need this function? One can easily stack frames to [B,H,W,C] that would guarantee size constraint

shaydeci: That won't work if H,W are not the same...

src/super_gradients/training/utils/videos.py
```diff
@@ -0,0 +1,105 @@
...
+import numpy as np
+
+SUPPORTED_FORMATS = (".avi", ".mp4", ".mov", ".wmv", ".mkv")
+
+
+__all__ = ["load_video", "save_video"]
+
+
+def load_video(file_path: str, max_frames: Optional[int] = None) -> List[np.ndarray]:
+    """Open a video file and extract each frame into numpy array.
```
BloodAxe, line 14: OpenCV loads images as BGR. I believe this should be reflected in documentation if it's intended. Or BGR->RGB conversion should be done inside. Save for image saving. When saving
OpenCV assumes input image is in BGR order.

src/super_gradients/training/utils/videos.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _open_video(file_path: str) -> cv2.VideoCapture:
+    """Open a video file.
+
+    :param file_path:   Path to the video file
+    :return:            Opened video capture object
+    """
+    ext = os.path.splitext(file_path)[-1].lower()
+    if ext not in SUPPORTED_FORMATS:
```
BloodAxe, line 33: Well, supported formats is defined inside OpenCV library and why not we rely on `if not cap.isOpened()` only? This check looks unnecessary to me.

src/super_gradients/training/utils/videos.py
Louis-Dupont: It doesnt work if H,W are not the same, we could pad but I don't think this should be our responsability, this should be done before

src/super_gradients/training/utils/videos.py
Louis-Dupont: I wasnt aware of that, thanks for that info

src/super_gradients/training/utils/videos.py
Louis-Dupont: Fixed in d2be71775d822f2a304b9f73044118add2ecaefd

BloodAxe (approved):
LGTM

PR #816, opened 2023-04-03
Feature/sg 735 qat ptq for deciyolo

shaydeci:
PR For PTQ and QAT tutorial. I left placeholder for visualisation (need to update after we have "predict" implemented).

Also, I included small fix for pre_launch_callbacks which was not updated with our current registry process.


BloodAxe (approved):
LGTM

PR #817, opened 2023-04-03
Added tutorial on DetectionOutputAdapter

BloodAxe:


shaydeci (approved):
LGTM.


Minor suggestion, would be nice to have some example of actual coordinates (could be just me though...your call (: )

BloodAxe:
@shaydeci thanks for the suggestion, in fact while adding an example output a bug in OutputAdapter was discovered and fixed. You can't have too many unit tests.

shaydeci (approved):

PR #819, opened 2023-04-04
Feature/sg 736 deci yolo predict

Louis-Dupont:
Adding predict

### DIFF
--- a/src/super_gradients/examples/predict/detection_predict.py
+++ b/src/super_gradients/examples/predict/detection_predict.py
@@ -2,12 +2,12 @@ from super_gradients.common.object_names import Models
 from super_gradients.training import models
 
 # Note that currently only YoloX and PPYoloE are supported.
-model = models.get(Models.PP_YOLOE_S, pretrained_weights="coco")
+model = models.get(Models.DECIYOLO_L, pretrained_weights="coco")
 
 IMAGES = [
     "https://miro.medium.com/v2/resize:fit:500/0*w1s81z-Q72obhE_z",
-    "https://s.hs-data.com/bilder/spieler/gross/128069.jpg",
     "https://datasets-server.huggingface.co/assets/Chris1/cityscapes/--/Chris1--cityscapes/train/28/image/image.jpg",
+    "https://media.timeout.com/images/105921973/750/422/image.jpg",
 ]
-prediction = model.predict(IMAGES, iou=0.65, conf=0.5)
+prediction = model.predict(IMAGES, iou=0.65, conf=0.6)
 prediction.show()
--- a/src/super_gradients/examples/predict/detection_predict.py
+++ b/src/super_gradients/examples/predict/detection_predict.py
@@ -9,5 +9,5 @@ IMAGES = [
     "https://datasets-server.huggingface.co/assets/Chris1/cityscapes/--/Chris1--cityscapes/train/28/image/image.jpg",
     "https://media.timeout.com/images/105921973/750/422/image.jpg",
 ]
-prediction = model.predict(IMAGES, iou=0.65, conf=0.6)
+prediction = model.predict(IMAGES)
 prediction.show()
--- a/src/super_gradients/training/models/detection_models/customizable_detector.py
+++ b/src/super_gradients/training/models/detection_models/customizable_detector.py
@@ -5,9 +5,7 @@ A base for a detection network built according to the following scheme:
  * each module accepts in_channels and other parameters
  * each module defines out_channels property on construction
 """
-
-
-from typing import Union, Optional
+from typing import Union, Optional, List
 
 from torch import nn
 from omegaconf import DictConfig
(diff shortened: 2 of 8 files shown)


src/super_gradients/training/transforms/processing.py
```diff
@@ -210,6 +253,8 @@ def get_pretrained_processing_params(model_name: str, pretrained_weights: str) -
         return default_yolox_coco_processing_params()
     elif "ppyoloe" in model_name and pretrained_weights == "coco":
         return default_ppyoloe_coco_processing_params()
+    elif "deciyolo" in model_name and pretrained_weights == "coco":
```
shaydeci, line 256: There is a better way to do this probably....

src/super_gradients/examples/predict/detection_predict.py
```diff
@@ -2,12 +2,12 @@
...
+model = models.get(Models.DECIYOLO_L, pretrained_weights="coco")
 
 IMAGES = [
     "https://miro.medium.com/v2/resize:fit:500/0*w1s81z-Q72obhE_z",
-    "https://s.hs-data.com/bilder/spieler/gross/128069.jpg",
     "https://datasets-server.huggingface.co/assets/Chris1/cityscapes/--/Chris1--cityscapes/train/28/image/image.jpg",
+    "https://media.timeout.com/images/105921973/750/422/image.jpg",
 ]
-prediction = model.predict(IMAGES, iou=0.65, conf=0.5)
+prediction = model.predict(IMAGES, iou=0.65, conf=0.6)
```
shaydeci, line 12: This is problematic, we are really letting our users figure out for themselves what visually works out the best...

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -96,3 +101,30 @@ def replace_head(self, new_num_classes: Optional[int] = None, new_head: Optional
...
+    def set_dataset_processing_params(self, class_names: Optional[List[str]], image_processor: Optional[Processing]) -> None:
+        """Set the processing parameters for the dataset.
+
+        :param class_names:     (Optional) Names of the dataset the model was trained on.
+        :param image_processor: (Optional) Image processing objects to reproduce the dataset preprocessing used for training.
+        """
+        self._class_names = class_names or self._class_names
+        self._image_processor = image_processor or self._image_processor
+
+    def predict(self, images, iou: float = 0.65, conf: float = 0.01) -> DetectionResults:
```
shaydeci, line 118: Continuing my previous comment - predict is here to visualize results ad not to optimize mAP, hence the values should be selected accordingly.
This means that the default values should be related to the dataset on which the model was trained on.
It also looks way better IMO if all the users needs to do is call predict on their images.

src/super_gradients/training/transforms/processing.py
```diff
@@ -242,3 +287,20 @@ def default_ppyoloe_coco_processing_params() -> Tuple[List[str], Processing]:
     )
     class_names = COCO_DETECTION_CLASSES_LIST
     return class_names, image_processor
+
+
+def default_deciyolo_coco_processing_params() -> Tuple[List[str], Processing]:
+    """Processing parameters commonly used for training DeciYolo on COCO dataset."""
+    from super_gradients.training.datasets.datasets_conf import COCO_DETECTION_CLASSES_LIST
```
shaydeci, line 294: Is this because of circular imports ?

src/super_gradients/training/transforms/processing.py
```diff
@@ -242,3 +287,20 @@ def default_ppyoloe_coco_processing_params() -> Tuple[List[str], Processing]:
...
+        [
+            ReverseImageChannels(),
+            DetectionLongestMaxSizeRescale(output_shape=(636, 636)),
+            DetectionCenterPadding(output_shape=(640, 640), pad_value=114),
+            StandardizeImage(max_value=255.0),
+            ImagePermute(permutation=(2, 0, 1)),
+        ]
+    )
+    class_names = COCO_DETECTION_CLASSES_LIST
+    return class_names, image_processor
```
shaydeci, line 306: So this is actually what we had concerns over yesterday...
preprocessing parameters might vary. Class names might not even be relevant for future tasks for instance.
My suggestion - lets return a dictionary that can later on be passed to set_dataset_processing_params.
This way we can also pass the best iou and conf in terms of visuall sucess on COCO.

shaydeci (changes requested):

src/super_gradients/training/models/detection_models/customizable_detector.py
Louis-Dupont: Done here d974313215196cc3c4210f1c49682040b93285cf

src/super_gradients/examples/predict/detection_predict.py
Louis-Dupont: Default added in d974313215196cc3c4210f1c49682040b93285cf

src/super_gradients/training/transforms/processing.py
Louis-Dupont: It was but actually it's fine when moved outside of function, so I took it out: 3262e2ab3d32b7cbcd640330984bded8a673f1c3

src/super_gradients/training/transforms/processing.py
Louis-Dupont: We now return a dict d974313215196cc3c4210f1c49682040b93285cf

PR #821, opened 2023-04-04
Feature/sg 735 deci yolo qs

shaydeci:


documentation/source/DeciYoloQuickstart.md
```diff
@@ -0,0 +1,75 @@
...
+We will use the ```src/super_gradients/recipes/roboflow_deciyolo_s.yaml```configuration to train the small variant of our DeciYolo, DeciYolo S.
+
+So we navigate to our ```train_from_recipe``` script:
+```commandline
+cd <YOUR-LOCAL-PATH>/super_gradients/src/super_gradients/examples/train_from_recipe_example
+```
+
+Then to avoid collisions between our cloned and installed SG:
+```commandline
+export PYTHONPATH=<YOUR-LOCAL-PATH>/super_gradients/:<YOUR-LOCAL-PATH>/super_gradients/src
```
BloodAxe, line 67: Why do we need adding PYTHONPATH? We assume we are writing docs for stable version of SG,
so user should have SG==3.10 installed. And now can one use `python -m super_gradients.train_from_recipe ...` 
I just don't see why we are asking user to do `cd` & `export` stuff.

shaydeci: Completely true, fixed!

Louis-Dupont (approved):
LGTM

PR #828, opened 2023-04-10
Feature/sg 747 support predict video full pipeline deci yolo

Louis-Dupont:


src/super_gradients/training/utils/media/media.py
```diff
@@ -0,0 +1,37 @@
...
+    # Check if media_files is a path to a folder containing images
+    if isinstance(media_files, str) and os.path.isdir(media_files):
+        return MediaType.IMAGE_FOLDER
+
+    # Check if media_files is an image
+    if isinstance(media_files, ImageSource):
+        return MediaType.IMAGE
+
+    # If none of the above conditions are met, raise an exception
+    raise ValueError("Invalid input type. Input should be a video, the path of a folder containing images, an image or a list of images.")
```
Louis-Dupont, line 37: I added this if we want to have `predict` function that supports ANY input (video, image_folder and image/images).
But this not being used at the moment, so if we don't want such a `predict` we might want to remove this

PR #829, opened 2023-04-10
Feature/sg 747 support predict video full pipeline master

Louis-Dupont:
Adding support for image folder and video as part of the base `predict` function. 
Also adding support for `predict_webcam`.
This also introduced some refactoring on the Pipeline to support lazy batch processing.


src/super_gradients/training/utils/media/media.py
```diff
@@ -0,0 +1,37 @@
...
+    # Check if media_files is a path to a folder containing images
+    if isinstance(media_files, str) and os.path.isdir(media_files):
+        return MediaType.IMAGE_FOLDER
+
+    # Check if media_files is an image
+    if isinstance(media_files, ImageSource):
+        return MediaType.IMAGE
+
+    # If none of the above conditions are met, raise an exception
+    raise ValueError("Invalid input type. Input should be a video, the path of a folder containing images, an image or a list of images.")
```
Louis-Dupont, line 37: I added this if we want to have `predict` function that supports ANY input (`video`, `image_folder` and `image`/`images`).
But this not being used at the moment, so if we don't want such a `predict` we might want to remove this

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -96,3 +105,92 @@ def replace_head(self, new_num_classes: Optional[int] = None, new_head: Optional
...
+    def predict_video(
+        self,
+        video_path: str,
+        output_video_path: str = None,
+        batch_size: int = 32,
+        visualize: bool = False,
+        iou: float = 0.65,
+        conf: float = 0.01,
+    ):
+        """Predict on a folder of images.
```
shaydeci, line 185: wrong docs

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -96,3 +105,92 @@ def replace_head(self, new_num_classes: Optional[int] = None, new_head: Optional
...
+        """Set the processing parameters for the dataset.
+
+        :param class_names:     (Optional) Names of the dataset the model was trained on.
+        :param image_processor: (Optional) Image processing objects to reproduce the dataset preprocessing used for training.
+        :param iou:             (Optional) IoU threshold for the nms algorithm
+        :param conf:            (Optional) Below the confidence threshold, prediction are discarded
+        """
+        self._class_names = class_names or self._class_names
+        self._image_processor = image_processor or self._image_processor
+        self._default_nms_iou = iou or self._default_iou
```
shaydeci, line 129: I dont get why we need both actually...(_default_iou and _default_nms_iou)

src/super_gradients/training/utils/media/media.py
BloodAxe: I think we should also support URL links as well. To simplify use in colab notebooks where we can simply put link to S3 location with an image

BloodAxe:
Regarding processing video/image folder, I wonder whether we can extract common logic via generators:

```
def process_series(self, images: Iterable[np.ndarray]) -> Iterable[Result]:
   for image in images:
      yield self.process(image)
```

So processing logic for a collection of images is in one place, regardless what if it is an image folder or video.

So now even if we want to process and save entire LoTR movie we can do this without storing all frames in memory:
```
class VideoFileIterator:
  def __init__(self, input_path):
     self.cap = cv2.VideoCapture(input_path)

  def __next__(self):
     success, frame = self.cap.grab()
     if not success:
        raise StopIteration()
     return frame

def process_video(self, input_path, output_path):
    input = VideoFileIterator(input_path)
    output = VideoWriter(output_path)
    for result in process_series(input):
        output.write_frame(result)
```

src/super_gradients/examples/predict/detection_predict_video.py
```diff
@@ -0,0 +1,8 @@
+from super_gradients.common.object_names import Models
+from super_gradients.training import models
+
+# Note that currently only YoloX and PPYoloE are supported.
+model = models.get(Models.YOLOX_N, pretrained_weights="coco")
```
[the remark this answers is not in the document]
Louis-Dupont: I added images for `detection_predict_images` and `detection_predict_images_from_folder`, but I cannot do it for `predict_video` and `predict_webcam` for obvious reasons.
I think once we have something that doesnt only work for yolox, ppyoloe and deciyolo we can add some nice tutorial on that

src/super_gradients/training/models/detection_models/customizable_detector.py
Louis-Dupont: You're right this was actually an error, I kept `self. _default_nms_iou ` and `self._default_nms_conf`

src/super_gradients/training/transforms/processing.py
```diff
@@ -204,19 +248,10 @@ def postprocess_predictions(self, predictions: DetectionPrediction, metadata: Re
...
-def default_yolox_coco_processing_params() -> Tuple[List[str], Processing]:
-    """Processing parameters commonly used for training YoloX on COCO dataset."""
-    from super_gradients.training.datasets.datasets_conf import COCO_DETECTION_CLASSES_LIST
+def default_yolox_coco_processing_params() -> dict:
+    """Processing parameters commonly used for training YoloX on COCO dataset.
+    TODO: remove once we load it from the checkpoint
+    """
 
     image_processor = ComposeProcessing(
         [
```
Louis-Dupont, line 257: This part and the code below (within the processing.py module) were actually merge to deciyolo already.
The thing is that I believe that we should merge first to master because otherwise the deciyolo branch would have lots of code to review at once, so I chose to merge this branch to master first, and I took with me some required code that was already in deciyolo branch.

src/super_gradients/examples/predict/detection_predict_image_folder.py
```diff
@@ -0,0 +1,9 @@
+from super_gradients.common.object_names import Models
+from super_gradients.training import models
+
+# Note that currently only YoloX and PPYoloE are supported.
+model = models.get(Models.PP_YOLOE_S, pretrained_weights="coco")
+
+image_folder_path = "<path/to/image_folder>"
```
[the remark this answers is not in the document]
Louis-Dupont: https://github.com/Deci-AI/super-gradients/blob/37e3e2b08cf7451c425752c5cace43b055284fa3/src/super_gradients/examples/predict/detection_predict_image_folder.py#L7-L9

documentation/source/images/examples/countryside.jpg
BloodAxe, line 1: This image looks like it has wrong channel order (BGR not RGB). is it intentional?

src/super_gradients/training/transforms/processing.py
BloodAxe: Make sense to me

src/super_gradients/training/transforms/processing.py
```diff
@@ -240,5 +282,49 @@ def default_ppyoloe_coco_processing_params() -> Tuple[List[str], Processing]:
...
+
+
+def default_deciyolo_coco_processing_params() -> dict:
+    """Processing parameters commonly used for training DeciYolo on COCO dataset.
+    TODO: remove once we load it from the checkpoint
+    """
+
+    image_processor = ComposeProcessing(
+        [
+            ReverseImageChannels(),
```
shaydeci, line 302: Like we said- its the other way around (since we load with PIL its already RGB).
Remove ReverseImageChannels() from here and add it to the other detection preprocessing pipeplines please.

shaydeci (changes requested):
1 small comment then LGTM on my end.

src/super_gradients/training/transforms/processing.py
Louis-Dupont: Changed in 21b5534c31cf9079e67b6201a2081764ac156d8a

shaydeci (approved):
LGTM

PR #832, opened 2023-04-13
Add support to yolov5 format dataset

Louis-Dupont:
This does the job, but:

### 1. 
Do we want to add recipe ? If we don't, we cannot register dataloader (which when we think about it is a little annoying)

### 2.
What seems like a bad design. A cleaner design would be to have the format as an input parameter. 
With the design I just pushed, if we want to add explicit support to robolow and coco we would need to implement 1 new class for each that would inherit from `Yolov5Formated...`. 
Having the format as a parameter would 1. reduce the amount of code and 2. be nicer and easier to use. The downside is that it would require some more work for this one PR and we might not want that.

### DIFF
--- a/src/super_gradients/training/datasets/detection_datasets/__init__.py
+++ b/src/super_gradients/training/datasets/detection_datasets/__init__.py
@@ -2,6 +2,6 @@ from super_gradients.training.datasets.detection_datasets.coco_detection import 
 from super_gradients.training.datasets.detection_datasets.pascal_voc_detection import PascalVOCDetectionDataset
 from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataset
 from super_gradients.training.datasets.detection_datasets.roboflow import RoboflowDetectionDataset
-from super_gradients.training.datasets.detection_datasets.yolo_format_detection import YoloFormatDetectionDataset
+from super_gradients.training.datasets.detection_datasets.yolo_format_detection import YoloDarknetFormatDetectionDataset
 
-__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset", "YoloFormatDetectionDataset"]
+__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset", "YoloDarknetFormatDetectionDataset"]
--- a/src/super_gradients/training/datasets/detection_datasets/__init__.py
+++ b/src/super_gradients/training/datasets/detection_datasets/__init__.py
@@ -2,5 +2,6 @@ from super_gradients.training.datasets.detection_datasets.coco_detection import 
 from super_gradients.training.datasets.detection_datasets.pascal_voc_detection import PascalVOCDetectionDataset
 from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataset
 from super_gradients.training.datasets.detection_datasets.roboflow import RoboflowDetectionDataset
+from super_gradients.training.datasets.detection_datasets.yolo_format_detection import YoloV5FormattedDetectionDataset
 
-__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset"]
+__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset", "YoloV5FormattedDetectionDataset"]
--- a/src/super_gradients/training/datasets/detection_datasets/__init__.py
+++ b/src/super_gradients/training/datasets/detection_datasets/__init__.py
@@ -2,6 +2,6 @@ from super_gradients.training.datasets.detection_datasets.coco_detection import 
 from super_gradients.training.datasets.detection_datasets.pascal_voc_detection import PascalVOCDetectionDataset
 from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataset
 from super_gradients.training.datasets.detection_datasets.roboflow import RoboflowDetectionDataset
-from super_gradients.training.datasets.detection_datasets.yolo_format_detection import YoloV5FormattedDetectionDataset
+from super_gradients.training.datasets.detection_datasets.yolo_format_detection import YoloFormatDetectionDataset
 
-__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset", "YoloV5FormattedDetectionDataset"]
+__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset", "YoloFormatDetectionDataset"]
--- a/src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
+++ b/src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
@@ -1,0 +1,125 @@
+import os
+
+import imagesize
+import numpy as np
+from typing import List, Optional
+
+from super_gradients.common.abstractions.abstract_logger import get_logger
+from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataset
+from super_gradients.training.exceptions.dataset_exceptions import DatasetValidationException
+
+from super_gradients.training.datasets.data_formats import ConcatenatedTensorFormatConverter
+from super_gradients.training.datasets.data_formats.default_formats import XYXY_LABEL, LABEL_NORMALIZED_CXCYWH
+
+logger = get_logger(__name__)
+
+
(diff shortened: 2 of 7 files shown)


src/super_gradients/training/datasets/detection_datasets/__init__.py
```diff
@@ -2,5 +2,6 @@
 from super_gradients.training.datasets.detection_datasets.pascal_voc_detection import PascalVOCDetectionDataset
 from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataset
 from super_gradients.training.datasets.detection_datasets.roboflow import RoboflowDetectionDataset
+from super_gradients.training.datasets.detection_datasets.yolo_format_detection import YoloV5FormattedDetectionDataset
 
-__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset"]
+__all__ = ["COCODetectionDataset", "DetectionDataset", "PascalVOCDetectionDataset", "RoboflowDetectionDataset", "YoloV5FormattedDetectionDataset"]
```
ofrimasad: don't call it V5. call it YoloFormattedDetectionDataset

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+from super_gradients.training.datasets.detection_datasets.detection_dataset import DetectionDataset
+from super_gradients.training.exceptions.dataset_exceptions import DatasetValidationException
+
+from super_gradients.training.datasets.data_formats import ConcatenatedTensorFormatConverter
+from super_gradients.training.datasets.data_formats.default_formats import XYXY_LABEL, LABEL_NORMALIZED_CXCYWH
+
+logger = get_logger(__name__)
+
+
+class YoloV5FormattedDetectionDataset(DetectionDataset):
```
ofrimasad: fix name

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+    """Base dataset to load ANY dataset that is with a similar structure to the YoloV5 dataset.
```
ofrimasad, line 18: Yolo Darknet dataset.
add here the structure of the dataset (directories, files...)

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+        images_dir: str,
+        labels_dir: str,
+        classes: List[str],
+        class_ids_to_ignore: Optional[List[int]] = None,
+        *args,
+        **kwargs,
+    ):
+        """
+        :param data_dir:                Where the data is stored.
+        :param images_dir:              Name of the directory that includes all the images. Path relative to data_dir.
```
ofrimasad, line 37: Please call it images_dir_name and labels_dir_name to make it clearer

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+        :param labels_dir:              Name of the directory that includes all the labels. Path relative to data_dir.
+        :param classes:                 List of class names.
```
ofrimasad, line 39: please make `classes` optional if possible

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+        self.images_folder = os.path.join(self.data_dir, self.images_dir)
+        self.labels_folder = os.path.join(self.data_dir, self.labels_dir)
+
+        self.images_file_names = list(sorted(os.listdir(self.images_folder)))
+        self.labels_file_names = list(sorted(os.listdir(self.labels_folder)))
+
+        image_file_base_names = set(os.path.splitext(os.path.basename(image_file_name))[0] for image_file_name in self.images_file_names)
+        label_file_base_names = set(os.path.splitext(os.path.basename(label_file_name))[0] for label_file_name in self.labels_file_names)
+        if image_file_base_names != label_file_base_names:
+            raise DatasetValidationException(f"image folder {self.images_folder} and label folder {self.labels_folder} include files that don't match")
```
ofrimasad, line 70: Let's not crash on this. 
A large dataset might be slightly corrupted. 
instead, log an error with the names of the missing files

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+            - "resized_img_shape":  Resides image (height, width)
+            - "img_path":           Path to the associated image
+        """
+        image_path = os.path.join(self.images_folder, self.images_file_names[sample_id])
+        label_path = os.path.join(self.labels_folder, self.labels_file_names[sample_id])
+
+        image_width, image_height = imagesize.get(image_path)
+        image_shape = (image_height, image_width)
+
+        yolov5_format_target = parse_yolov5_label_file(label_path)
```
ofrimasad, line 90: please remove the "v5"

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,125 @@
...
+    """Parse a single label file in yolo v5 format.
+
+    :return: np.ndarray of shape (n_labels, 5) in yolo v5 format (LABEL_NORMALIZED_CXCYWH)
+    """
+    with open(label_file_path, "r") as f:
+        labels_txt = f.read()
+
+    labels_yolo_v5_format = []
+    for line in labels_txt.split("\n"):
+        label_id, cx, cw, w, h = line.split(" ")
```
ofrimasad, line 123: Please add a TODO here to support additional fields (it will require the user to pass a ConcatenatedTensorFormat)

ofrimasad (commented):
looks very nice.
My comments inline

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: Changed in 773b92834d6cdd9d3ec530e5da7069527b171aaa

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: Changed in 773b92834d6cdd9d3ec530e5da7069527b171aaa

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: removed

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: To not crash on this, we have to filter out labels/images that don't have a valid image/label file.
74a041e68abfb641b48cb0d09ff00a8041243905

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: done

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: The solution would be to read it from a yaml they get when downloading from ultralytics/roboflow, but as discussed we prefer to let the user set it manually

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -0,0 +1,190 @@
...
+            "target": target,
+            "initial_img_shape": image_shape,
+            "resized_img_shape": resized_img_shape,
+            "img_path": image_path,
+            "id": np.array([sample_id]),
+        }
+        return annotation
+
+
+def parse_yolo_label_file(label_file_path: str) -> np.ndarray:
```
ofrimasad, line 176: why is that not a class method? 
can we make it static and private?

ofrimasad (commented):
I generally agree with @BloodAxe, but let's take this out of the scope of this PR.
I have added a Jira ticket for this change: https://deci.atlassian.net/browse/SG-779

see my minor comment inline

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
Louis-Dupont: Sure
0aec05f9a5d50dceae72974033ea09073817d89b

ofrimasad (approved):
LGTM

PR #833, opened 2023-04-16
Fix yolo darknet dataset doc + param name

Louis-Dupont:
I added some doc for yolo/darknet detection dataset.
This made me realized that renaming `image_dir` to `image_dir_name` was a mistake because it can also be a path (e.g. `test/images` and this is not intuitive if names `image_dir_name`.
We can still change it quickly without breaking change since nobody is using it.


ofrimasad (approved):
LGTM

PR #835, opened 2023-04-17
Feature/sg 772 predict using validation

shaydeci:
Added support for translating the transforms pipe in DetectionDataset to processing parameters.
Flow:
- If the validation datset is instance(HasProcessingParams) and our model isinstance(HasPredict), we set the model's  processing Parameters (tpreprocessing transforms , NMS params etc).
- If the above condition holds, we also save them under the "processing_params" entry inside the checkpoint, then later these can be decoded when loading them in models.get (see tests).
- Added factory for processing.
- Added unit tests.

We still need to implement the mechanism for our other datasets (note that it is important that we do so on the dataset side (i.e making it HasProcessingParams) since each dataset might have its own way of loading the sample + hard-coded transforms. This lets the user define exactly what is needed to have the same pipeline working with our way of loading images in Predict.


src/super_gradients/module_interfaces/module_interfaces.py
```diff
@@ -0,0 +1,36 @@
...
+        """Set the processing parameters for the dataset."""
+        pass
+
+    @abstractmethod
+    def predict(self, images, *args, **kwargs) -> None:
+        pass
+
+    @abstractmethod
+    def predict_webcam(self, *args, **kwargs) -> None:
+        pass
```
Louis-Dupont, line 36: images is missing typing, and return type is not None
`def predict(self, images: ImageSource, *args, **kwargs) -> ImagesPredictions:`

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
```diff
@@ -14,11 +17,11 @@
 from super_gradients.training.models.detection_models.pp_yolo_e.post_prediction_callback import PPYoloEPostPredictionCallback, DetectionPostPredictionCallback
 from super_gradients.training.models.prediction_results import ImagesDetectionPrediction
 from super_gradients.training.pipelines.pipelines import DetectionPipeline
-from super_gradients.training.transforms.processing import Processing
+from super_gradients.training.processing.processing import Processing
 from super_gradients.training.utils.media.image import ImageSource
 
 
-class PPYoloE(SgModule):
+class PPYoloE(SgModule, HasPredict):
```
Louis-Dupont, line 24: I would prefer to avoid double inheritance. In this case this can be handled with protocols:

```
from typing import Protocol, runtime_checkable


@runtime_checkable
class MyInterface(Protocol):
    def method1(self) -> None:
        ...

    def method2(self) -> None:
        ...


class ClassRespectingInterface:
    def method1(self):
        pass

    def method2(self):
        pass


class ClassNOTRespectingInterface:
    def method1(self):
        pass


print("ClassRespectingInterface: ", isinstance(ClassRespectingInterface(), MyInterface))
print("ClassNOTRespectingInterface: ", isinstance(ClassNOTRespectingInterface(), MyInterface))
```
```
ClassRespectingInterface:  True
ClassNOTRespectingInterface:  False
```

src/super_gradients/training/models/detection_models/yolo_base.py
```diff
@@ -390,7 +393,7 @@ def forward(self, intermediate_output):
         )
 
 
-class YoloBase(SgModule):
+class YoloBase(SgModule, HasPredict):
```
Louis-Dupont, line 396: Same as above (I would go for protocol type checking)

src/super_gradients/module_interfaces/module_interfaces.py
shaydeci: I dont want to import any type from inside SG, as these interfaces should be able to be imported anywhere.
Dropped the return type for now.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1314,6 +1320,16 @@ def forward(self, inputs, targets):
...
                 self.sg_logger.close()
 
+    def _set_net_preprocessing_from_valid_loader(self):
+        if isinstance(self.net.module, HasPredict) and isinstance(self.valid_loader.dataset, HasPreprocessingParams):
+            try:
+                self.net.module.set_dataset_processing_params(**self.valid_loader.dataset.get_dataset_preprocessing_params())
+            except Exception as e:
+                logger.warning(
+                    f"Could not set preprocessing pipeline from the validation dataset: {e}. Before calling"
+                    "predict make sure to call set_dataset_processing_params."
```
Louis-Dupont, line 1330: Maybe adding "\n" before "Before calling [...]" would help because otherwise it would be in the same line as the error (so ppl probably won't see it)

src/super_gradients/training/transforms/transforms.py
```diff
@@ -441,6 +444,9 @@ def __call__(self, sample: dict) -> dict:
         sample["image"] = (sample["image"] / self.max_value).astype(np.float32)
         return sample
 
+    def get_equivalent_preprocessing(self) -> List[Dict]:
+        return [{Processings.StandardizeImage: {"max_value": self.max_value}}]
```
Louis-Dupont, line 448: Why don't we directly instantiate it? I know we pass it through the factory, but the factory can also be support instantiated object. Is there another reason?

It would look more simple, more explicit, and easier to debug  + here we cannot directly click on the class StandardizeImage.

Louis-Dupont (commented):
Looks great, just check out the protocol thing which I believe would be much cleaner.

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -473,3 +474,22 @@ def plot(self, max_samples_per_plot: int = 16, n_plots: int = 1, plot_transforme
...
+            pipeline += [{Processings.DetectionLongestMaxSizeRescale: {"output_shape": self.input_dim}}]
+        for t in self.transforms:
+            pipeline += t.get_equivalent_preprocessing()
+        params = dict(
+            class_names=self.classes,
+            image_processor={Processings.ComposeProcessing: {"processings": pipeline}},
+            iou=0.65,
+            conf=0.5,
+        )
+        return params
```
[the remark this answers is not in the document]
shaydeci: I was contemplating about this, on one hand there would be some fixed values that might not be the best fit for all models. On the other hand, we dont need to set anything other then this before predict.

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
shaydeci: @BloodAxe WDYT?
Also after talking to @Louis-Dupont , we were wondering why you did not make the module interfaces abstract classes to force the methods implementations ?

src/super_gradients/training/transforms/transforms.py
shaydeci: Since we also need to revert back to "factory processable" input to save them inside the checkpoint file, I just do it first here (also thought it is not a good practice to import these objects here).
It is less risky translating this into instances rather then the other way around.

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
BloodAxe: To start, I was not aware of typing.Protocol :)
Second, Protocol supported since python 3.8. I beleive we still want to support 3.7, so Protocol is not an option yet :/
But it's looking good and https://jellis18.github.io/post/2022-01-11-abc-vs-protocol/ explains quite nicely the difference between ABC and protocols and if we decide to ditch 3.7 support I would be quite happy to  migrate to them.
If the double inheritance is a concernt, we don't have to use (ABC) as base class. We still can use decorators @abstractmethod to mark methods as abstract and put their default implementation manually:
```
class MyInterface):
    @abstractmethod
    def method1(self) -> None:
        raise NotImplementedError()
```

shaydeci: Nice catch!
Seems that typing-extensions supports Protocol and runtime_checkable this 3.7 and up though.
Any objections for using them?
I do like @Louis-Dupont suggestion, as we might want to have multiple interfaces to obey to in the future.

BloodAxe: Sounds good to me 👍 to use `typing-extensions`

shaydeci: changed in https://github.com/Deci-AI/super-gradients/pull/835/commits/d9e3019ce07736224b9e07687bc770e793a2c612

src/super_gradients/training/models/detection_models/yolo_base.py
shaydeci: changed in https://github.com/Deci-AI/super-gradients/commit/d9e3019ce07736224b9e07687bc770e793a2c612

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: Added in latest commit.

Louis-Dupont (approved):
LGTM

PR #843, opened 2023-04-18
Support empty label files for yolo format and be robust to invalid format

Louis-Dupont:


src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -198,18 +201,26 @@ def _load_annotation(self, sample_id: int) -> dict:
...
-        return np.array(labels_yolo_format)
+        for line in filter(lambda x: x != "\n", lines):
+            try:
+                label_id, cx, cw, w, h = line.split(" ")
+                labels_yolo_format.append([int(label_id), float(cx), float(cw), float(w), float(h)])
+            except Exception as e:
+                if ignore_invalid_labels:
+                    logger.warning(f"Line `{line}` of file {label_file_path} will be ignored because not in LABEL_NORMALIZED_CXCYWH format: {e}")
+                else:
+                    raise e
```
Louis-Dupont, line 225: I added an option to ignore when the label could not be parsed properly (and log a warning).
I can definitely imagine this happening from time to time.
But maybe we want to handle it differently and raise an explicit error instead of just dropping the label.
Any thought?

ofrimasad (approved):
LGTM

PR #846, opened 2023-04-18
Hotfix/sg 000 fix predict show

Louis-Dupont:


src/super_gradients/training/utils/detection_utils.py
```diff
@@ -385,20 +386,12 @@ def _draw_box_title(
...
-        # Draw the box
-        image_np = cv2.rectangle(image_np, (x1, y1), (x2, y2), color, box_thickness)
-
-        # Caption with class name and confidence if given
-        text_color = (255, 255, 255)  # white
-
         if is_target:
             title = f"[GT] {class_name}"
-        if not is_target:
+        else:
```
ofrimasad, line 391: looks like it will still print "[Pred]"

ofrimasad (commented):
looks good, but the "[Pred]" tag is still printed

src/super_gradients/training/utils/detection_utils.py
Louis-Dupont: When using `_draw_box_title` yes because we do use in in visualization that plots both predictions and ground truth.
In the case of predict, we don't call this `_draw_box_title` method anymore, and instead directly call `draw_bbox`

ofrimasad (approved):
LGTM

PR #849, opened 2023-04-20
Feature/sg src add examples to init

Louis-Dupont:


src/super_gradients/evaluate_from_recipe.py
```diff
@@ -0,0 +1,44 @@
...
+
+For recipe's specific instructions and details refer to the recipe's configuration file in the recipes directory.
+"""
+from omegaconf import DictConfig
+import hydra
+
+from super_gradients import Trainer, init_trainer
+
+
+@hydra.main(config_path="recipes", version_base="1.2")
```
Louis-Dupont, line 33: `@hydra.main(config_path=package_ressources("super_gradients.recipes", ""), ...)` crashes the code after moving to the root of sg
```
Primary config module 'super_gradients..home.louis.dupont.PycharmProjects.super-gradients.src.super_gradients.recipes' not found.
Check that it's correct and contains an __init__.py file
```

This can be fixed by just using `@hydra.main(config_path="recipe", ...)` .
I don't see a reason not to use `config_path="recipes"` here anyway, do you have any thoughts about that?

shaydeci (approved):
LGTM

src/super_gradients/evaluate_from_recipe.py
BloodAxe: IMO, should be fine to use config_path="recipes". Can't think of cases when it can break.

PR #850, opened 2023-04-20
Tab completion

BloodAxe:
See documentation file in this PR for more details


Louis-Dupont:
Just follow the instructions and it ran pretty fast on my computer (Mac), but I only see some keys. Is it normal?
image

BloodAxe:
I would say yes, it does not show you "all" keys for a given config. For some reason root settings are not there. Perhaps we should evaluate this feature more thoroughly to see whether it is really helps.

BloodAxe:
It is. I assume you expected to see all root parameters, right? We have no control on tab completion so if it is not providing useful completion - i guess we can reconsider whether we want to have this feature.

documentation/source/TabCompletion.md
```diff
@@ -0,0 +1,42 @@
...
+compinit
+autoload -Uz bashcompinit && bashcompinit
+```
+
+See [Tab Completion](https://hydra.cc/docs/tutorials/basic/running_your_app/tab_completion/) for more information.
+
+## Caveats & Limitations
+
+* Tab completion is not supported on Windows.
+* You need to enable tab completion explicitly for each entry point in your application. In other words, if your project.
```
shaydeci, line 25: if your project....
Line swapped here I believe.

documentation/source/TabCompletion.md
```diff
@@ -0,0 +1,42 @@
...
+* Tab completion is not supported for running train_from_recipe using `-m` form: `python -m super_gradients.train_from_recipe ...`.
```
shaydeci, line 26: Why not actually?
Cant we just do the same wherever SG package is stored in ? Or is it a problem with the fact that its being ran as a module?

shaydeci (changes requested):

shaydeci:
Im closing this down, as I dont think we currently benefit enough from tab completion.
Perhaps we can revisit this in the future, when tab completion is more ready from Hydra's side.

PR #852, opened 2023-04-23
Feature/sg 810 recipe shortcuts

shaydeci:
- Added shortcuts for our yaml based scripts using hydra.callbacks (experimental, but seems to do the job and is available from hydra 1.2).

- I added these through variable_setup.yaml that is compiled last after the main recipe config, then populating it to avid confusion in logs if we dont use the shortcuts (i.e lr will be set to training_hyperparams.inital_lr just for clarity).
- I also moved up the ckpt_root_dir resolvement.
- Updated all recipes to use variable_setup last in their defaults list.
- Updated training recipe tests to use shortcuts as coverage and to simplify the commands.

Note that https://github.com/Deci-AI/super-gradients/pull/852/commits/ecefd46b166cd4e2df9c945da4ac9137db237f2b  is to trigger the recipe tests in CI, please do not merge until I set the release tag filter back.


src/super_gradients/recipes/variable_setup.yaml
```diff
@@ -0,0 +1,17 @@
+lr:
```
BloodAxe, line 1: Could you please add a comment to each shortcut telling to which "main" parameter it aliases to?

src/super_gradients/common/environment/omegaconf_utils.py
```diff
@@ -1,9 +1,56 @@
...
+            resume: False
+            num_workers
+
+    When any of the above are not set, they will be populated with the original values (for example
+        config.lr will be set with config.training_hyperparams.initial_lr) for clarity in logs.
+
+    """
+
+    def on_run_start(self, config: DictConfig, **kwargs: Any) -> None:
+        config.training_hyperparams.initial_lr = config.lr or config.training_hyperparams.initial_lr
```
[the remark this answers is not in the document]
shaydeci: Nice one.
I modified it to explicitly check for None so we can include boolean params / 0 value.

src/super_gradients/recipes/variable_setup.yaml
shaydeci: sure, added in latest commit + more docs.

src/super_gradients/recipes/variable_setup.yaml
```diff
@@ -0,0 +1,42 @@
...
+#
+#
+#  Interpolates the shortcuts defined below, with their aliases (see comments near each parameter).
+#  When any of the above are not set, they will be populated with the original values (for example
+#      config.lr will be set with config.training_hyperparams.initial_lr) for clarity in logs.
+#
+# Note that interpolation is done by triggering RecipeShortcutsCallbackm which is a Hydra Callback (see http://hydra.cc/docs/experimental/callbacks/)
+#  so interpolation of these in other yaml configuration files won't be present.
+
+lr: # config.training_hyperparams.initial_lr
```
Louis-Dupont, line 20: I would add an inline example, I think it makes the use more straightforward:

`python train_from_recipe --config-name=recipe lr=0.003`

`python train_from_recipe --config-name=recipe config.training_hyperparams.initial_lr=0.003`

Louis-Dupont (commented):
LGTM beside a small comment

src/super_gradients/recipes/variable_setup.yaml
shaydeci: NP, added in c2d8e7ca47a3c72b5a5830122728c708f3bd8d30

Louis-Dupont (approved):
LGTM

PR #858, opened 2023-04-24
Improve visualization of detection by drawing semi-transparent bboxes…

BloodAxe:
… (75% opacity) and using tight-layout when plotting an image using MPL.


shaydeci (approved):
LGTM

PR #862, opened 2023-04-25
MKdocs compatability

RanZilberstein:
1. adding structure yaml for the doc structure
2. changing the readme to be markdown safe as the mkdocs compiler is much more stricter


Louis-Dupont (approved):
LGTM

PR #869, opened 2023-04-28
Add DagsHub Logger to Super Gradients

deanp70:
@timho102003 has added code to introduce a DagsHub logger to Super Gradients. Please let us know if there are any questions.

It uses MLflow and DVC under the hood to track experiment parameters and metrics while versioning the datasets and models used, and pushes all the above to your DagsHub repository.

Example DagsHub repository with the logger:
https://dagshub.com/timho102003/super_grad_test_v2

### DIFF
--- a/src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
+++ b/src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
@@ -1,0 +1,227 @@
+import os
+from pathlib import Path
+from typing import Optional
+
+import torch
+
+from super_gradients.common.registry.registry import register_sg_logger
+from super_gradients.common.abstractions.abstract_logger import get_logger
+
+from super_gradients.common.sg_loggers.base_sg_logger import BaseSGLogger
+from super_gradients.common.environment.ddp_utils import multi_process_safe
+
+logger = get_logger(__name__)
+
+try:
+    import dagshub
+    from dagshub.upload import Repo
+
+    _import_dagshub_error = None
+except (ModuleNotFoundError, ImportError, NameError) as dagshub_import_err:
+    _import_dagshub_error = dagshub_import_err
+
+try:
+    import mlflow
+
+    _import_mlflow_error = None
+except (ModuleNotFoundError, ImportError, NameError) as mlflow_import_err:
+    _import_mlflow_error = mlflow_import_err
+
+
+@register_sg_logger("dagshub_sg_logger")
+class DagsHubSGLogger(BaseSGLogger):
+    def __init__(
+        self,
+        project_name: str,
+        experiment_name: str,
+        storage_location: str,
+        resumed: bool,
+        training_params: dict,
+        checkpoints_dir_path: str,
+        tb_files_user_prompt: bool = False,
+        launch_tensorboard: bool = False,
+        tensorboard_port: int = None,
+        save_checkpoints_remote: bool = True,
+        save_tensorboard_remote: bool = True,
+        save_logs_remote: bool = True,
+        monitor_system: bool = None,
+        dagshub_repository: Optional[str] = None,
+        log_mlflow_only: bool = False,
+    ):
+        """
+
+        :param experiment_name:         Name used for logging and loading purposes
+        :param storage_location:        If set to 's3' (i.e. s3://my-bucket) saves the Checkpoints in AWS S3 otherwise saves the Checkpoints Locally
+        :param resumed:                 If true, then old tensorboard files will **NOT** be deleted when tb_files_user_prompt=True
+        :param training_params:         training_params for the experiment.
+        :param checkpoints_dir_path:    Local root directory path where all experiment logging directories will reside.
+        :param tb_files_user_prompt:    Asks user for Tensorboard deletion prompt.
+        :param launch_tensorboard:      Whether to launch a TensorBoard process.
+        :param tensorboard_port:        Specific port number for the tensorboard to use when launched (when set to None,
+                                        some free port number will be used
+        :param save_checkpoints_remote: Saves checkpoints in s3 and DagsHub.
+        :param save_tensorboard_remote: Saves tensorboard in s3.
+        :param save_logs_remote:        Saves log files in s3 and DagsHub.
+        :param monitor_system:          Save the system statistics (GPU utilization, CPU, ...) in the tensorboard
+        :param dagshub_repository:      Format: <dagshub_username>/<dagshub_reponame> format is set correctly to avoid
+                                        any potential issues. If you are utilizing the dagshub_sg_logger, please specify
+                                        the dagshub_repository in sg_logger_params to prevent any interruptions from
+                                        prompts during automated pipelines. In the event that the repository does not
+                                        exist, it will be created automatically on your behalf.
+        :param log_mlflow_only:         Skip logging to DVC, use MLflow for all artifacts being logged
+        """
+        if monitor_system is not None:
+            logger.warning("monitor_system not available on DagsHubSGLogger. To remove this warning, please don't set monitor_system in your logger parameters")
+
+        self.s3_location_available = storage_location.startswith("s3")
+        super().__init__(
+            project_name=project_name,
+            experiment_name=experiment_name,
+            storage_location=storage_location,
+            resumed=resumed,
+            training_params=training_params,
+            checkpoints_dir_path=checkpoints_dir_path,
+            tb_files_user_prompt=tb_files_user_prompt,
+            launch_tensorboard=launch_tensorboard,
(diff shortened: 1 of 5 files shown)


src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
```diff
@@ -0,0 +1,222 @@
...
+    def _init_env_dependency(self):
+        """
+        The function creates paths for the DVC directory, models, and artifacts, obtains an authentication token from Dagshub, and sets MLflow tracking credentials. 
+        It also checks whether the repository name and owner have been set and prompts the user to enter them if they haven't. 
+        If the remote URI is not set or does not include "dagshub", Dagshub is initialized with the repository name and owner, and the remote URI is obtained. 
+        The method then creates a Repo object with the repository information and sets the DVC folder to the DVC directory path. 
+        """
+
+        self.paths = {
+            "dvc_directory": Path("artifacts"),
```
ofrimasad, line 124: Where will these directories be created? 
Please make sure they are created inside the `checkpoints_dir_path`

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @multi_process_safe
+    def _dvc_commit(self, commit=""):
+        self.dvc_folder.commit(commit, versioning="dvc", force=True)
+
+    @multi_process_safe
+    def add_config(self, tag: str, config: dict):
+        super(DagsHubSGLogger, self).add_config(tag=tag, config=config)
+        param_keys = config.keys()
+        for pk in param_keys:
```
ofrimasad, line 167: Why not:  `mlflow.log_params(config)`?
Wouldn't this work?

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
```diff
@@ -0,0 +1,222 @@
...
+    def add_scalars(self, tag_scalar_dict: dict, global_step: int = 0):
+        super(DagsHubSGLogger, self).add_scalars(tag_scalar_dict=tag_scalar_dict, global_step=global_step)
+        mlflow.log_metrics(metrics=tag_scalar_dict, step=global_step)
+
+    @multi_process_safe
+    def close(self):
+        super().close()
+        try:
+            if not self.log_mlflow_only:
+                self._dvc_commit(commit=f"Adding all artifacts from run {mlflow.active_run().info.run_id}")
```
ofrimasad, line 189: Note that although this approach will be faster during training, it might be risky when training on a Spot instance in AWS (or equivalent in other cloud providers).
In such cases, the process has only limited time to stop, and uploading the files only on process closing may cause some files to be lost. 

Consider maybe calling commit every x epochs...

ofrimasad (changes requested):
Looks great. Thank you for this contribution! 
I have some change requests inline. Most important is the auth token issue.

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
```diff
@@ -0,0 +1,222 @@
...
+        :param tb_files_user_prompt:    Asks user for Tensorboard deletion prompt.
+        :param launch_tensorboard:      Whether to launch a TensorBoard process.
+        :param tensorboard_port:        Specific port number for the tensorboard to use when launched (when set to None, some free port number will be used
+        :param save_checkpoints_remote: Saves checkpoints in s3 and DagsHub.
+        :param save_tensorboard_remote: Saves tensorboard in s3.
+        :param save_logs_remote:        Saves log files in s3 and DagsHub.
+        :param monitor_system:          Save the system statistics (GPU utilization, CPU, ...) in the tensorboard
+        :param dagshub_repository:      Format: <dagshub_username>/<dagshub_reponame> format is set correctly to avoid any potential issues. If you are utilizing the dagshub_sg_logger, please specify the dagshub_repository in sg_logger_params to prevent any interruptions from prompts during automated pipelines. In the event that the repository does not exist, it will be created automatically on your behalf. 
+        :param dagshub_auth:            Provide dagshub authentication token for automatic pipeline
```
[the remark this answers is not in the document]
deanp70: Good point. Will remove it.

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
deanp70: This is actually one of the cool benefits of using DagsHub with DVC. These are all directories that are created on the DagsHub repo, without touching the local environment, so this can run from anywhere without creating files locally.

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
deanp70: Great point, let me explain the logic here:
1. MLflow is the "log everything" tool here, so after every epoch, when we add checkpoints, I log everything (including the model) to MLflow
2. DVC is the reproducibility tool, so I only want to commit things I'm likely to need to reproduce, therefore I'm logging the final and best results to it. The user can always log their artifacts manually if they want, but this makes sure they don't make their git history too overwhelming.

If the user is using spot instances and it disconnects in the middle, they will have all the previous checkpoints logged to MLflow.

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
deanp70: I did a bit of digging and apparently MLflow doesn't support nested dicts as params. I'll write a flatten function and then log that to MLflow.

deanp70: After looking over this, this would either require changing a lot of parameter names in order to log the 2 broken parameters which are the "list of packages" and the "log schema".

If we were to move to mlflow log_params as is it would break on logging, so for now I suggest we leave it as is. We're looking into a way to fix this on our end.

deanp70:
I addressed all the comments. Please let me know if you have any follow up questions

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
ofrimasad: if the issue is the space in the key, we can do 
```
config = {k.replace(" ", "_"):v for k, v in config.items()}
```
before calling `mlflow.log_params(config)`

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
ofrimasad: Got it. thanks

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
ofrimasad: ok. thanks

src/super_gradients/common/sg_loggers/dagshub_sg_logger.py
deanp70: Actually, the problem is that both parameters are too long (since they are big nested params) and MLflow has a character limit for parameter values. So we can either remove them manually, or write code to flatten the entire dict, which makes it less readable.

Happy to do either, though especially since the list of packages will cause the number of parameters to explode.

ofrimasad: ok. so let's keep it like this for now.

shaydeci (approved):
LGTM

ofrimasad (approved):
LGTM!
thank you!
Try to update branch to see if the CI runs

shaydeci:
> @timho102003 has added code to introduce a DagsHub logger to Super Gradients. Please let us know if there are any questions.
> 
> It uses MLflow and DVC under the hood to track experiment parameters and metrics while versioning the datasets and models used, and pushes all the above to your DagsHub repository.
> 
> Example DagsHub repository with the logger: https://dagshub.com/timho102003/super_grad_test_v2

@deanp70 If you merge master into your branch it should pass the build now.

deanp70:
@shaydeci thanks, did it, and I realized I missed some Flake8 issues. They are all fixed now (AFAICT). Can you approve it for running again?

Thanks 🙏

ofrimasad (approved):
LGTM

deanp70:
Not sure what to do about the signing of commits, I'll see if I can change this

ofrimasad:
If this is possible, then yes. Not sure Github will see this as a solution....

deanp70:
@ofrimasad combined all changes into one commit which is verified. You need to re-approve it (sorry for the multiple approvals, but I think this is the last one 🤞)

ofrimasad (approved):

PR #871, opened 2023-04-30
remove unnecessary login to codeartifact on normal branch

ranrubin:
This change will allow external users to use the CI when contributing code


shaydeci (approved):
Thanks!
LGTM

PR #876, opened 2023-05-02
Change Documentation Link 

Louis-Dupont:
Note: I didnt change the `docs` folder which, from what I remember, handles sphinx. Do we want to keep that folder now that we moved to a new doc ?


BloodAxe (approved):
LGTM

PR #882, opened 2023-05-03
freeze setuptools version

ofrimasad:


requirements.txt
```diff
@@ -8,7 +8,7 @@ scipy>=1.6.1
 matplotlib>=3.3.4
 psutil>=5.8.0
 tensorboard>=2.4.1
-setuptools>=21.0.0
+setuptools==59.5.0
```
BloodAxe, line 11: Does it have to be that strict, or >= would also work?
Given the fact 59.5 was released in late 2021, do we want to pin our users for this specific version? I imagine this may conflict with other packages in environment.

shaydeci:
@ofrimasad is this still relevant?

PR #885, opened 2023-05-03
Add missing calls to RepVGG blocks to fuse layers

BloodAxe:
On the left is unfused, on the right - fused
image


shaydeci (approved):
LGTM

PR #887, opened 2023-05-03
Fix leaking file-handle in logger

BloodAxe:
This is relevant for tests, that run multiple test cases in the single process.
This PR ensures that we close all file handles before setting up new logger


shaydeci (approved):
LGTM

PR #906, opened 2023-05-04
Update deprecate -> deprecated

BloodAxe:
This small PR fixes wrong import of `deprecate` package that we do not declare in our requirements.
Instead we use `deprecated`. So I changed that. The motivation for this fix is when we loose some of our requirements, the `deprecated` package is not installed anymore and we get import error because of that.


shaydeci (approved):
LGTM

PR #924, opened 2023-05-05
Update the notebooks to install super_gradients==3.1

BloodAxe:
@Louis-Dupont as requested.
Looks like it does not strip outputs, innit?


Louis-Dupont (approved):
Yeah you don't have this issue, I'll try to understand why
Anyway, LGTM

Issue #930, opened 2023-05-05
How to find the class labels of the detected objects (img.prediction.labels) only returns the numeric value


maninka123:

Louis-Dupont:
Each image prediction also includes an attribute `class_names` with all the classes names that the model tries to predict, here is an example:

```python
images_predictions = model.predict(IMAGES)

for image_prediction in images_predictions:
    class_names = image_prediction.class_names  # List of class names the model tries to predict for in the image
    labels = image_prediction.prediction.labels
    confidence = image_prediction.prediction.confidence
    bboxes = image_prediction.prediction.bboxes_xyxy
    for i, (label, conf, bbox) in enumerate(zip(labels, confidence, bboxes)):
        print("prediction: ", i)
        print("label_id: ", label)
        print("label_name: ", class_names[int(label)])
        print("confidence: ", conf)
        print("bbox: ", bbox)
        print("--" * 10)
```

Louis-Dupont:
@maninka123 we just released a new [documentation](https://docs.deci.ai/super-gradients/documentation/source/ModelPredictions.html) on `model.predict()` was added, check it out for more information, and feel free to let us know if anything is unclear.

Issue #907, opened 2023-05-04
How can i see the prediction result of a frame/image as bbox coordinates ?


ngocnd2402:
I want to combine yolo-nas with deepsort for object tracking problem, so I need to get the bbox of each object.

Louis-Dupont:
[Tutorial](https://docs.deci.ai/super-gradients/documentation/source/ModelPredictions.html) was released with answers to the questions mentioned in this issue, so I'm closing it :)

Issue #892, opened 2023-05-03
Filtering Classes


sivaji123256:
Can we filter the classes ?Is there any argument that we can use ?

BloodAxe:
Could you please elaborate what the use case? Do you want to detect only certain classes using pretrained model or something different?

BloodAxe:
Regarding filtering of classes - no, currently this is not supported.

Regarding inference time - Numbers that we shared on the frontier plot are obtained using TensorRT inference engine (batch size 1). This is in consistency with other models on that plot (YOLO-V8/7/6/5,PPYOLO). So we compare apples to apples.

Native pytorch inference is expected to be much slower. There are a number of reasons:
1) Eager execution has it's price
2) PyTorch model by default uses non-fused RepVGG blocks which is suitable for training, but for inference you may want to fuse them (You can achieve this by calling `model.prep_model_for_conversion((640, 640))` but bear in mind that you cannot train this model after you've called prep_model_for_conversion)
3) Video decoding / Moving images to GPU / Visualization of results takes additional time.

I'm not ready to comment regarding yolov8 as I'm not familiar with implementation of their inference.

sivaji123256:
@BloodAxe ,Thanks .But first of all , can you confirm me whether I was calculating the FPS value in a correct way as per the previous image attached?

Louis-Dupont:
The processing FPS can actually be seen directly next to the loading bar. In the following example, the FPS is `39.49fps`
`Predicting Video: 100%|███████████████████████| 306/306 [00:07<00:00, 39.49it/s]`
If you do it manually you might get a different (wrong) value because the time in sec (7 here) is rounded. In my example, this would give: `306/7=43.71fps`.

Note that there are 2 fps values: the one of the original video and the one at which we process the video.
To avoid confusion, we chose not to write the processing fps on the video because it could be mistaken for the video fps. Both type of fps are different because the video is not streamed (we first process it, and then visualize/save in the original fps)

NatanBagrov:
Hi @sivaji123256, regarding your original question of filtering classes.
[...]

PR #943, opened 2023-05-06
Add detection prediction tutorial

Louis-Dupont:
I made sure to include answers of the following issues:
- get bbox information - https://github.com/Deci-AI/super-gradients/issues/930
- get bbox information - https://github.com/Deci-AI/super-gradients/issues/907
- fps & filtering - https://github.com/Deci-AI/super-gradients/issues/892
- set_processing_params: https://github.com/Deci-AI/super-gradients/issues/903

Let me know if you feel like I'm still missing important questions/explanations or if it's just not clear.

**TODO: update the collab before merging**

### DIFF
--- a/documentation/source/ModelPredictions.md
+++ b/documentation/source/ModelPredictions.md
@@ -1,6 +1,6 @@
 # Using Pretrained Models for Predictions 
 
-In this tutorial, we will demonstrate how to use the `model.predict()` method in the Super Gradients library for object detection tasks.
+In this tutorial, we will demonstrate how to use the `model.predict()` method for object detection tasks.
 
 The model used in this tutorial is [YOLO-NAS](YoloNASQuickstart.md), pre-trained on the [COCO dataset](https://cocodataset.org/#home), which contains 80 object categories.
 
--- a/documentation/source/ModelPredictions.md
+++ b/documentation/source/ModelPredictions.md
@@ -9,6 +9,7 @@ The model used in this tutorial is [YOLO-NAS](YoloNASQuickstart.md), pre-trained
 
 *Note that the  `model.predict()` method is currently only available for detection tasks.*
 
+
 ## Detect Objects in Multiple Images
 
 #### Load the Model and Prepare the Images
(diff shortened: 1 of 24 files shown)


documentation/source/ModelPredictions.md
```diff
@@ -0,0 +1,181 @@
...
+]
+
+images_predictions = model.predict(IMAGES)
+```
+You can use the default IoU and Confidence threshold or override them like this:
+
+```python
+images_predictions = model.predict(IMAGES, iou=0.5, conf=0.7)
+```
+- `iou`: IoU threshold for the non-maximum suppression (NMS) algorithm. If None, the default value associated with training is used.
```
shaydeci, line 42: Thats (currently) not true, its defined by the model class.

documentation/source/ModelPredictions.md
```diff
@@ -0,0 +1,181 @@
+# Using Pretrained Models for Predictions 
+
+In this tutorial, we will demonstrate how to use the `model.predict()` method for object detection tasks.
+
+The model used in this tutorial is [YOLO-NAS](YoloNASQuickstart.md), pre-trained on the [COCO dataset](https://cocodataset.org/#home), which contains 80 object categories.
+
+
+**Warning**: If your model was trained on your own custom dataset, make sure to follow these [steps first](DetectionPredictionCustom.md).
```
shaydeci, line 8: Its a bit unclear...
A custom dataaset can also be one that uses our COCO Format Detection Dataset class.
In that case (and actually in any case where we just trian with SG objects and use our transforms for detection) there is nothing to be done (recall my feature that takes the validation transforms and creates a pipeline from it).

Please also mention that, as in that case you save the users trouble defining these things.

shaydeci (changes requested):

documentation/source/ModelPredictions.md
Louis-Dupont: 35af53371555093a4ab5aea8bb73120296bb8ee6
I refered to SG datasets, and not transforms explicitly because the SG detection datasets all work with SG transforms, so I think there is no point to go too deep in implementations details:
What I wrote;
You are using SG dataset ? -> No need to set the params
You are not ? -> set the params

documentation/source/ModelPredictions.md
Louis-Dupont: https://github.com/Deci-AI/super-gradients/commit/35af53371555093a4ab5aea8bb73120296bb8ee6

shaydeci (approved):
LGTM

PR #944, opened 2023-05-07
Feature/sg 850 support torch 2.0

shaydeci:
- Fixed some failing tests in build after moving it to run on 3.8.
- Moved build to run on 3.8 (minimum python version requirement for torch 2.0)
- Fixed missing arch_params.num classes that were missing for pose estimation regression test.
- Updated requirement, pip version in CI.

### DIFF
--- a/src/super_gradients/recipes/coco2017_pose_dekr_w32_no_dc.yaml
+++ b/src/super_gradients/recipes/coco2017_pose_dekr_w32_no_dc.yaml
@@ -55,6 +55,9 @@ ckpt_root_dir:
 train_dataloader: coco2017_pose_train
 val_dataloader: coco2017_pose_val
 
+arch_params:
+  num_classes: ${dataset_params.num_joints}
+
 checkpoint_params:
   # Original training recipe uses pretrained weights for HRNet on ImageNet.
   # You will need to download the pretrained weights from the original repo and place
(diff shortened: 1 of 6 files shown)


Louis-Dupont (approved):
LGTM

src/super_gradients/recipes/coco2017_pose_dekr_w32_no_dc.yaml
```diff
@@ -55,6 +55,9 @@ ckpt_root_dir:
 train_dataloader: coco2017_pose_train
 val_dataloader: coco2017_pose_val
 
+arch_params:
```
ofrimasad, line 58: How is that related?

tests/unit_tests/detection_utils_test.py
```diff
@@ -24,7 +24,7 @@ def setUp(self):
     def test_visualization(self):
 
         valid_loader = coco2017_val(dataloader_params={"batch_size": 16})
-        trainer = Trainer("visualization_test", device=self.device)
+        trainer = Trainer("visualization_test")
```
ofrimasad, line 27: how is that related?

ofrimasad (commented):
please see a few questions before merge

tests/unit_tests/detection_utils_test.py
shaydeci: Not related, but I ran all of our unit tests (including those outside the suite) and fixed anything that was crashing.
Passing a divece through trainer was deprecated and causes to crash since few versions ago.

src/super_gradients/recipes/coco2017_pose_dekr_w32_no_dc.yaml
shaydeci: Not related.
I ran our regression tests, and this one was crashing because of missing arch_params.num_classes.
Thouht Itd be a good idea to fix this now and not forget about it later.

PR #945, opened 2023-05-07
Force predictions as list

Louis-Dupont:
This will make image prediction a list instead of a generator when running on single image (which was not intended)


shaydeci (approved):
LGTM

PR #948, opened 2023-05-08
Bug/sg 764 wrong ckpt when resuming with external ckpt root dir

shaydeci:
Since we dont pass ckpt_root_dir (till now) to get_checkpoints_dir_path it takes by default the pkg_resources one. Then if someone calls train with training_hyperparams.resume=True ckpt_root_dir=/some/path/to/ckpt_root/ we discard the ckpt_root_dir and get that we are trying to load a checkpoint that does not exist.

### DIFF
--- a/src/super_gradients/common/environment/checkpoints_dir_utils.py
+++ b/src/super_gradients/common/environment/checkpoints_dir_utils.py
@@ -58,25 +58,32 @@ def get_checkpoints_dir_path(experiment_name: str, ckpt_root_dir: str = None) ->
     return os.path.join(ckpt_root_dir, experiment_name)
 
 
-def get_ckpt_local_path(source_ckpt_folder_name: str, experiment_name: str, ckpt_name: str, external_checkpoint_path: str):
+def get_ckpt_local_path(source_ckpt_folder_name: str, experiment_name: str, ckpt_name: str, external_checkpoint_path: str, ckpt_root_dir: str = None) -> str:
     """
     Gets the local path to the checkpoint file, which will be:
-        - By default: YOUR_REPO_ROOT/super_gradients/checkpoints/experiment_name.
+        - By default: YOUR_REPO_ROOT/super_gradients/checkpoints/experiment_name/ckpt_name.
+        - external_checkpoint_path when external_checkpoint_path != None
+        - ckpt_root_dir/experiment_name/ckpt_name when ckpt_root_dir != None.
         - if the checkpoint file is remotely located:
             when overwrite_local_checkpoint=True then it will be saved in a temporary path which will be returned,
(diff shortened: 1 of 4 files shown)


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1489,7 +1489,9 @@ def _load_checkpoint_to_model(self):  # noqa: C901 - too complex
 
         if self.load_checkpoint or self.external_checkpoint_path:
             # GET LOCAL PATH TO THE CHECKPOINT FILE FIRST
+            ckpt_root_dir = str(Path(self.checkpoints_dir_path).parent)
             ckpt_local_path = get_ckpt_local_path(
+                ckpt_root_dir=ckpt_root_dir,
                 source_ckpt_folder_name=self.source_ckpt_folder_name,
```
Louis-Dupont, line 1495: Not directly related, but it looks like `self.source_ckpt_folder_name` is always None int he Trainer. I think it would be better to remove because it makes it confusing. Typically, I first wondered what is the difference between `ckpt_root_dir` vs `self.source_ckpt_folder_name` ?

src/super_gradients/common/environment/checkpoints_dir_utils.py
```diff
@@ -58,25 +58,32 @@ def get_checkpoints_dir_path(experiment_name: str, ckpt_root_dir: str = None) ->
     return os.path.join(ckpt_root_dir, experiment_name)
 
 
-def get_ckpt_local_path(source_ckpt_folder_name: str, experiment_name: str, ckpt_name: str, external_checkpoint_path: str):
+def get_ckpt_local_path(source_ckpt_folder_name: str, experiment_name: str, ckpt_name: str, external_checkpoint_path: str, ckpt_root_dir: str = None) -> str:
```
Louis-Dupont, line 61: Adding to my previous comment, you can see that `get_ckpt_local_path` is called only in 1 place which is the Trainer with `self. source_ckpt_folder_name=None`. So I think we could just remove this `source_ckpt_folder_name` parameter and go with the parameter you introduced which is more explicit.

Besides, the name `source_ckpt_folder_name` was missleading because looking at the code, it was expected to be the local name of the folder including experiments and not the full path

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: removed (including from weight averaging) in latest commit.

src/super_gradients/common/environment/checkpoints_dir_utils.py
shaydeci: removed (including from weight averaging) in latest commit.

PR #973, opened 2023-05-10
added find_detection_score_threshold

itaylevy-deci:
computes the best threshold for each class in an efficient manner without added loops


utils_script/find_detection_score_threshold.py
```diff
@@ -0,0 +1,49 @@
...
+from super_gradients.common.environment.cfg_utils import add_params_to_cfg
+from super_gradients import Trainer, init_trainer
+
+
+@hydra.main(config_path=pkg_resources.resource_filename("super_gradients.recipes", ""), version_base="1.2")
+def main(cfg: DictConfig) -> None:
+    add_params_to_cfg(cfg.training_hyperparams.valid_metrics_list[0].DetectionMetrics, params=["calc_best_score_thresholds=True"])
+    _, valid_metrics_dict = Trainer.evaluate_from_recipe(cfg)
+
+    class_names = COCO_DETECTION_CLASSES_LIST  # change this line to use a different dataset
```
BloodAxe, line 33: Great work. I really love the way to find thresholds using command-line args approach: ` python find_detection_score_threshold.py --config-name=coco2017_yolox architecture=yolox_n
            checkpoint_params.pretrained_weights=coco`
However it seems that for other dataset one have to change the hard-coded `class_names` constant. Is it possible to improve this part and move class names to dataset_params maybe?

Also, consider moving this script to `super_gradients.scripts` to enable using this script as follows:
`python -m super_gradients.find_detection_score_threshold --config-name=coco2017_yolox architecture=yolox_n checkpoint_params.pretrained_weights=coco`

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -1196,4 +1233,4 @@ def compute_detection_metrics_per_cls(
     # Average over the recall_thresholds
     ap = sampled_precision_points.mean(0)
 
-    return ap, precision, recall
+    return ap, precision, recall, best_score_threshold
```
shaydeci, line 1236: I think it would be safer to not return the threshold here..
perhaps set it as an attribute, or return it somewhere else.

shaydeci (changes requested):
Very nice work!

- See one comment inline.
- There's still a missing piece we spoke of - we get the different thresholds per class, then we wish to use these in our NMS.
- It would also be great to just have the threshold that maximizes F1.
- Nice to have: same mechanism for the other metrics (mAP etc).

src/super_gradients/training/metrics/detection_metrics.py
```diff
@@ -30,6 +30,7 @@ class DetectionMetrics(Metric):
     :param dist_sync_on_step:               Synchronize metric state across processes at each ``forward()`` before returning the value at the step.
     :param accumulate_on_cpu:               Run on CPU regardless of device used in other parts.
                                             This is to avoid "CUDA out of memory" that might happen on GPU.
+    :param calc_best_score_thresholds       Whether to calculate the best score threshold per class
```
shaydeci, line 33: If we leave it this way, please give some guidance what we will have retiurned when we use this flag.

shaydeci (changes requested):

shaydeci:
@itaylevy-deci This PR has been open for quite a while now and has some comments that were not addressed.

BloodAxe (approved):
LGTM

shaydeci (approved):
LGTM

Issue #967, opened 2023-05-10
'Trainer' object has no attribute 'train_loader' error


fleventy-5:
I'm Trying to train yolo-nas on a custom dataset and I get the below error while running the training on google colab using 
```
trainer.train(model=model, 
              training_params=train_params, 
              train_loader=train_data, 
              valid_loader=val_data)
```
![image](https://github.com/Deci-AI/super-gradients/assets/78148524/b725c386-8b13-41d3-9644-6ceb16f7a69c)

Louis-Dupont:
Hi @fleventy-5 
I think that the `train_loader` you passed is not a `Dataloader` but instead either `None`, `[]`, `False`, or any "0" value.
Can you please check it ?

Louis-Dupont:
@momonoki3nenn, @fleventy-5 ,
I did not manage to reproduce it, but we pushed a change that should fix it. It will be in the next release, but meanwhile, you can install SG from our repo directly :)
``` shell
pip install git+https://github.com/Deci-AI/super-gradients
```

Louis-Dupont:
@momonoki3nenn 
The `StopIteration` appears because we are trying to iterate over a Dataloader that is apparently empty (you can try `next(iter(train_data))` and you will get the same error)

Another thing that supports this hypothesis is the fact that on the right of the "Caching annotation" line, there is "1/1" written, which shows that your dataset doesn't include multiple images/labels.

So now the question is why is the dataloader empty ?
I would guess that either you dont point to the right path of your dataset or maybe your dataset doesnt have the right structrure. 
Feel free to check this [documentation page](https://docs.deci.ai/super-gradients/src/super_gradients/training/datasets/Dataset_Setup_Instructions.html#computer-vision-datasets-setup) to see how your data should be structured to use this (or another) dataset.
To test it simply, you can check the length or iterate over it `len(train_data)` or `next(iter(train_data))`.

Vikram12301:
Is there any fix found for the _StopIteration_ issue?

Louis-Dupont:
@Vikram12301 , did you install the nightly ?
``` shell
pip install git+https://github.com/Deci-AI/super-gradients
```
If yes, do you have a single sample, in your train set?
Please also provide the full snippet of code you are using to run, with among other dataset instantiation.

Louis-Dupont:
[...]

PR #981, opened 2023-05-11
Fix train_loader not initalized properly

Louis-Dupont:
I am not sure why this error happens to begin (see https://github.com/Deci-AI/super-gradients/issues/967), and I couldnt reproduce it, but this should fix it:
- Make the code more robust to this type of error
- Raise a more explicit error when user doesnt provide a loader and not loader found in cache.


shaydeci (approved):
LGTM

Issue #912, opened 2023-05-04
MaskAttentionLoss in DiceCEEdgeLoss doesn't handle images without any edges


davidtvs:
### Describe the bug

Training models that use `DiceCEEdgeLoss` results in NaN loss on images that only contain one semantic class. The `edge_target` becomes a tensor filled with zeros because there are no edges in the image:

https://github.com/Deci-AI/super-gradients/blob/aa2745450e4c2617847874eb7f54bcab4ebb8f10/src/super_gradients/training/losses/dice_ce_edge_loss.py#L101-L103

Then, when computing the `MaskAttentionLoss`, `mask_loss` is a tensor filled with zeros, gets reassigned to an empty tensor, and, finally, computing the mean of an empty tensor results in NaN.

https://github.com/Deci-AI/super-gradients/blob/aa2745450e4c2617847874eb7f54bcab4ebb8f10/src/super_gradients/training/losses/mask_loss.py#L45-L47

### To Reproduce
I've written a new test in `tests/unit_tests/mask_loss_test.py` that reproduces the problem.

```python
def test_with_cross_entropy_loss_maskless(self):
    """
    Test case with mask filled with zeros, corresponding to a scenario without
    attention. It's expected that the mask doesn't contribute to the loss.

    This scenario may happen when using edge masks on an image without
    edges - there's only one semantic region in the whole image.

    Shapes: predict [BxCxHxW], target [BxHxW], mask [Bx1xHxW]
    """
    predict = torch.randn(self.batch, self.num_classes, self.img_size, self.img_size)
    target = self._get_default_target_tensor()
    # Create a mask filled with zeros to disable the attention component
    mask = self._get_default_mask_tensor() * 0.0

    loss_weigths = [1.0, 0.5]
    ce_crit = nn.CrossEntropyLoss(reduction="none")
    mask_ce_crit = MaskAttentionLoss(criterion=ce_crit, loss_weights=loss_weigths)

    # expected result - no contribution from mask
    ce_loss = ce_crit(predict, target)
    expected_loss = ce_loss.mean() * loss_weigths[0]

    # mask ce loss result
    loss = mask_ce_crit(predict, target, mask)

    self._assertion_torch_values(expected_loss, loss)
```

Running this test results in:
```txt
AssertionError: False is not true : Unequal torch tensors: excepted: 1.7192925214767456, found: nan
```

### Expected behavior
A mask filed with zeros should "disable" attention. Thus, the mask should not contribute to the loss.

[...]

PR #982, opened 2023-05-11
Fix empty mask in MaskAttentionLoss

lkdci:
Fix case where MaskAttentionLoss return nan values when the input mask is empty. Reported in https://github.com/Deci-AI/super-gradients/issues/912


shaydeci (approved):
LGTM

Issue #978, opened 2023-05-11
bug in super-gradients/src/super_gradients/training/models/conversion.py


JagdishKolhe:
### To Reproduce
1. net = models.get(model_name, pretrained_weights="coco")
2.  models.convert_to_onnx(model=net, out_path=f"{model_name}.onnx", torch_onnx_export_kwargs={'input_size': (3,640,640)})

### Expected behavior
Conversion should happen without giving validation errer

### Screenshots
![super_gradients_kwrg_validation_bug](https://github.com/Deci-AI/super-gradients/assets/54310972/b4541766-26c5-48a5-b0e1-9517f3bfcb46)

    models.convert_to_onnx(model=net, out_path=f"{model_name}.onnx", torch_onnx_export_kwargs={'input_size': (3,640,640)})
  File "/opt/conda/lib/python3.8/site-packages/super_gradients/common/decorators/factory_decorator.py", line 36, in wrapper
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/super_gradients/common/decorators/factory_decorator.py", line 36, in wrapper
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/super_gradients/training/models/conversion.py", line 89, in convert_to_onnx
    prep_model_for_conversion_kwargs["input_size"] = (1, *input_shape)
TypeError: 'NoneType' object is not iterable

The change should be:
if input_shape is not None:
        logger.warning(
            "input_shape is deprecated and will be removed in the next major release." " Use the input_size kwarg in prep_model_for_conversion_kwargs instead"
        )
        prep_model_for_conversion_kwargs["input_size"] = (1, *input_shape)

shaydeci:
Thanks @JagdishKolhe , we will look into it and update.

BloodAxe:
I don't think you need to use `torch_onnx_export_kwargs` directly. Passing `input_shape` argument should do the work just fine:

```
from super_gradients.common.object_names import Models
from super_gradients.training import models

net = models.get(Models.YOLO_NAS_L, pretrained_weights="coco")
models.convert_to_onnx(model=net, out_path=f"yolo_nas.onnx", input_shape=(3,640,640))
```

@JagdishKolhe may I ask you where did you get the code snippet? It may that we have outdated documentation somewhere and I'd like to address that to prevent such issue in future. Thanks!

BloodAxe:
Ah I see what you mean. Yeah, it's definitely a bug

Issue #979, opened 2023-05-11
SystemError: new style getargs format but argument is not a tuple


janardan-ds:
While doing train_data.dataset.plot() getting the system error :

SystemError                               Traceback (most recent call last)
Cell In[42], line 1
----> 1 train_data.dataset.plot()

File ~/anaconda3/lib/python3.8/site-packages/super_gradients/training/datasets/detection_datasets/detection_dataset.py:463, in DetectionDataset.plot(self, max_samples_per_plot, n_plots, plot_transformed_data)
    460 index = img_i + plot_i * 16
    462 if plot_transformed_data:
--> 463     image, targets, *_ = self[img_i + plot_i * 16]
    464     image = image.transpose(1, 2, 0).astype(np.int32)
    465 else:
SystemError: new style getargs format but argument is not a tuple

<Figure size 1000x1000 with 0 Axes>

And also during traing getting the runtime error

RuntimeError: DataLoader worker (pid(s) 1762330) exited unexpectedly

BloodAxe:
It is hard to tell exactly what cause this problem. 

Can you please provide more information about your issue:
- Link to colab if possible
- What dataset are you using and dataset params you've used to instantiate it?
- What is the number of samples in the dataset 
- Version of SuperGradients you're using

janardan-ds:
=> What dataset are you using and dataset params you've used to instantiate it? --> jpg and txt (yolo format)
=> What is the number of samples in the dataset --> around 4k in train (images and labels), 1k in val and 200 in test.
=> Version of SuperGradients you're using --> latest 3.1.1

Louis-Dupont:
Hi @janardan-ds 

Concerning the Dataset plot, it is a little buggy and we will fix it soon, but until then I don't have a quick workaround for that.

Concerning the error you got in training, this is a known pytorch issue (https://github.com/pytorch/pytorch/issues/5301), I'm not sure why it happens, but usually setting `num_workers` to 0 solves it, you can try this out:

``` python
train_data = coco_detection_yolo_format_train(
    dataset_params={
        'data_dir': dataset_params['data_dir'],
        'images_dir': dataset_params['train_images_dir'],
        'labels_dir': dataset_params['train_labels_dir'],
        'classes': dataset_params['classes']
    },
    dataloader_params={
        'batch_size':16,
        'num_workers': 0 # instead of 2
    }
)
```

Let me know if this works for you :)

PR #989, opened 2023-05-12
Fix convert_to_onnx to correctly handle case when input_shape is None

BloodAxe:
Fixes bug https://github.com/Deci-AI/super-gradients/issues/978


BloodAxe:
#979

PR #998, opened 2023-05-14
Predict on fused model

Louis-Dupont:
Eventually, I took the fused model out of the SgModule to move it to the pipeline, which makes more sense and simplifies the code.
I also added cache with `lru_cache` to solve the issue of knowing when the pipeline needs to be instantiated or not.


src/super_gradients/modules/qarepvgg_block.py
```diff
@@ -316,5 +316,14 @@ def fuse_block_residual_branches(self):
...
         raise NotImplementedError
 
-    def prep_model_for_conversion(self, input_size: Optional[Union[tuple, list]] = None, **kwargs):
-        self.partial_fusion()
+    def prep_model_for_conversion(self, input_size: Optional[Union[tuple, list]] = None, full_fusion: bool = False, **kwargs):
+        """Prepare the QARepVGGBlock for conversion.
+
+        :param full_fusion: If True, performs full fusion, converting the block into a non-trainable, fully fused block.
+                            If False, performs partial fusion, slower for inference but still trainable.
+        """
```
Louis-Dupont, line 324: I set it to False by default for backward compatibility, but I am not sure if it is relevant.

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -97,7 +102,7 @@ def _initialize_weights(self, bn_eps: Optional[float] = None, bn_momentum: Optio
     def prep_model_for_conversion(self, input_size: Optional[Union[tuple, list]] = None, **kwargs):
         for module in self.modules():
             if module != self and hasattr(module, "prep_model_for_conversion"):
-                module.prep_model_for_conversion(input_size, **kwargs)
+                module.prep_model_for_conversion(input_size, full_fusion=True, **kwargs)  # FIXME: Where should we set `full_fusion` ?
```
Louis-Dupont, line 105: What is the best way to set `full_fusion=True`?
If we set `full_fusion` to `True` by default this question would not be relevant. 

But If we don't I am not sure whether to hard code it here (might also create backward compatibility issue).
We could also just pass it in the kwargs. But it would be very likely that people will forget about it and even worse not even know about it.

So this makes me think that maybe we should just set it to `True` by default

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
```diff
@@ -59,12 +64,14 @@ def set_dataset_processing_params(
         self._default_nms_iou = iou or self._default_nms_iou
         self._default_nms_conf = conf or self._default_nms_conf
 
-    def _get_pipeline(self, iou: Optional[float] = None, conf: Optional[float] = None) -> DetectionPipeline:
+    @lru_cache(maxsize=1)
```
Louis-Dupont, line 67: I added a cache of 1, to keep in memory the latest pipeline used:
- If the user uses the same `iou`/`conf`, we keep the pipeline as is
    - to not instantiate everything again
    - to avoid fusing the model every time
- If the user sets different iou/conf, we instantiate a new pipeline, which will take a little more time

src/super_gradients/modules/qarepvgg_block.py
shaydeci: Im pretty sure we need partial fusion for QAT, correct me if im wrong @spsancti .

shaydeci (changes requested):

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
Louis-Dupont: In practice they do, either working with a lot of images that dont fit into memory, processing a big video, or doing custom stuff with webcam stream

src/super_gradients/modules/qarepvgg_block.py
Louis-Dupont: I talked to @spsancti and we chose to set full_fusion as default, since as you said `fuse_repvgg_blocks` is used for QAT and not `prep_model_for_conversion`

src/super_gradients/examples/benchmark/benchmark_gif.py
```diff
@@ -0,0 +1,42 @@
+import torch
+from typing import Callable
+from super_gradients.common.object_names import Models
+from super_gradients.training import models
+from super_gradients.training.utils.media.video import load_video
+
+
+def benchmark_model(action: Callable, benchmark_cycles: int, warmup_cycles: int):
```
shaydeci, line 8: I am not sure this is something we wish to show...
It has not been in our favor to show our benchmark results in Python.

src/super_gradients/modules/qarepvgg_block.py
```diff
@@ -316,5 +316,16 @@ def fuse_block_residual_branches(self):
     def from_repvgg(self, src: RepVGGBlock):
         raise NotImplementedError
 
-    def prep_model_for_conversion(self, input_size: Optional[Union[tuple, list]] = None, **kwargs):
-        self.partial_fusion()
+    def prep_model_for_conversion(self, input_size: Optional[Union[tuple, list]] = None, full_fusion: bool = True, **kwargs):
```
shaydeci, line 319: Im wondering whether it would be agood idea to let our users know that once we fuse the model it is not trainable anymore.
On one hand, its not true in the context of predict, on the other hand if someonce calls prep_model_for conversion explicitly- they might try for some reason to train the model again.
Just thinking out loud here.

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -59,6 +67,15 @@ def __init__(self, model: SgModule, image_processor: Union[Processing, List[Proc
             image_processor = ComposeProcessing(image_processor)
         self.image_processor = image_processor
 
+        self.fuse_model = fuse_model  # If True, the model will be fused in the first forward pass, to make sure it gets the right input_size
+
+    def _fuse_model(self, input_example: torch.Tensor):
+        logger.info("Fusing some of the model's layers. If this takes too much memory, you can deactivate it by setting `fuse_model=False`")
+        self.model = copy.deepcopy(self.model)
```
shaydeci, line 74: We need to release GPU memory here once we are done predicting (that is, when the original model goes into train mode).
We can delete the model copy / entire pipeline and then call torch.cuda.empty_cache() (when cuda is a available of course).

shaydeci (changes requested):

src/super_gradients/modules/qarepvgg_block.py
BloodAxe: I'd plan for future the following enchancement:
For RepVGGBlock & QARepVGGBlock:
```
class QARepVGGBlock:
   def train():
      if self.fused:
          raise RuntimeError("Fused module cannot be used in training")
```

This would effectively prevent using model with fused blocks in trainer.

shaydeci: We can figure it out though

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -136,12 +137,14 @@ def set_dataset_processing_params(
         self._default_nms_iou = iou or self._default_nms_iou
         self._default_nms_conf = conf or self._default_nms_conf
 
-    def _get_pipeline(self, iou: Optional[float] = None, conf: Optional[float] = None) -> DetectionPipeline:
+    @lru_cache(maxsize=1)
```
BloodAxe, line 140: I'm not sure using lru_cache here is a good idea since it can retain the fused model inside internal cache state of `lru_cache`. And since we don't have control on the cache lifetime my fear it would cause unwanted memory leaks. That's the first issue.

Second - I think we introducing too much knowledge about the internal state of the model for predict method. I'm talking about fuse_model=True, that to my eye clearly not the responsibility of the prediction pipeline to take care of model fusing.
To me, it's better to have an explicit step:
```
model.fuse()
model.predict(...) 
```
WDYT?

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -155,6 +172,8 @@ def _generate_prediction_result_single_batch(self, images: Iterable[np.ndarray])
         # Predict
         with eval_mode(self.model):
```
BloodAxe, line 173: Maybe add no_grad and autocast? Should not hurt and give us a nice boost.

```
with eval_mode(self.model), torch.no_grad(), torch.cuda.amp.autocast():
   ...
```

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -155,6 +172,8 @@ def _generate_prediction_result_single_batch(self, images: Iterable[np.ndarray])
...
             torch_inputs = torch.Tensor(np.array(preprocessed_images)).to(self.device)
```
BloodAxe, line 174: Consider using `torch.from_numpy` as it would skip a data copy (share memory with numpy array).
What type preprocessed_images has that we convert it to `np.array`

src/super_gradients/training/models/detection_models/customizable_detector.py
Louis-Dupont: (Just wrote a big answer and refreshed the page 😢 )

Concerning lru_cache, I added this in the model:
``` python
    def train(self, mode: bool = True):
        super().train(mode)

        self._get_pipeline.cache_clear()
        torch.cuda.empty_cache()
```
You think there is still a risk of unwanted memory leak ?

Concerning the second point, I agree with you generally speaking, it's definitly a cleaner design.
The issue is see is that:
- People would forget to run it
- Yolov8 fuses in the first predict call, so when people will want to compare 'apples to apples' (i.e., for them, just run predict, without any additional code) they will actually be comparing a fused YoloV8 to an non-fused YoloNAS - and this is what is already hapenning

Eventually, it all comes to the trade-off:
- Separation of concern and clean code
VS
- Simple interface for users to use it

Based on the issues we get, I think it is ambitious to expect people to go over the doc and see that they can fuse the model (they might not even know what it means). But I'm open to discussion, I think there should be a limit on how much we are willing to give up on having a clean code in order to simplify the users life.

src/super_gradients/examples/benchmark/benchmark_gif.py
Louis-Dupont: I took it out

src/super_gradients/training/pipelines/pipelines.py
Louis-Dupont: Done with:
``` python
    def train(self, mode: bool = True):
        super().train(mode)

        self._get_pipeline.cache_clear()
        torch.cuda.empty_cache()
```

src/super_gradients/training/models/detection_models/customizable_detector.py
BloodAxe: > You think there is still a risk of unwanted memory leak ?

I think that's should be good. 

> Based on the issues we get, I think it is ambitious to expect people to go over the doc and see that they can fuse the model 

Yeah, probably you have the point here.

src/super_gradients/training/pipelines/pipelines.py
Louis-Dupont: changed

src/super_gradients/modules/qarepvgg_block.py
Louis-Dupont: Why not:
``` python
class QARepVGGBlock:
   def train():
      if self.full_fusion: # Partially fused model still be trainable for QAT
          raise RuntimeError("Fused module cannot be used in training")
```

Louis-Dupont: @BloodAxe Is is part of the scope of this ticket ? https://deci.atlassian.net/browse/SG-839

BloodAxe: Yup, added it not to forget about it)

BloodAxe (approved):
LGTM

shaydeci (approved):
LGTM

PR #1001, opened 2023-05-14
Bug/sg 861 decouple qat from train from config

shaydeci:
This PR decouples QAT/PTQ from train_from_config.
The goal is to let users launch PTQ/QAT using Python instead of CLI + configs, while using as much defaults as possible.
Note that automatic adaptation of the parameters using the best practices is the users responsibilty when they choose to launch this way, since the new "quantize" method I added can expect objects rather then parameters.

- Renamed train_from_config in QATTrainer to quantize_from_config.
- Introduced a "quantize" method.
- Added the option to pass little parameters as possible to let users use existing objects which are already set in train().
- I added a simple test that takes a model from train to QAT to our recipe test suite (modified th ci config to install pytorch quantisation beforehand - it is possible since these tests run on gpu on spot instances).
- Added alot of docs as there were barely any.

### DIFF
--- a/src/super_gradients/qat_from_recipe.py
+++ b/src/super_gradients/qat_from_recipe.py
@@ -14,7 +14,7 @@ from super_gradients.training.qat_trainer.qat_trainer import QATTrainer
 
 @hydra.main(config_path="recipes", version_base="1.2")
 def _main(cfg: DictConfig) -> None:
-    QATTrainer.train_from_config(cfg)
+    QATTrainer.quantize_from_config(cfg)
 
 
 def main():
--- a/tests/recipe_training_tests/coded_qat_launch_test.py
+++ b/tests/recipe_training_tests/coded_qat_launch_test.py
@@ -1,0 +1,39 @@
+import unittest
+
+from super_gradients import QATTrainer
+from super_gradients.training.dataloaders.dataloaders import classification_test_dataloader
+from super_gradients.training.metrics import Accuracy, Top5
+from super_gradients.training.models import ResNet18
+
+
+class CodedQATLuanchTest(unittest.TestCase):
+    def test_qat_launch(self):
+        trainer = QATTrainer("test_launch_qat_with_minimal_changes")
+        net = ResNet18(num_classes=5, arch_params={})
+        train_params = {
+            "max_epochs": 2,
+            "lr_updates": [1],
+            "lr_decay_factor": 0.1,
+            "lr_mode": "step",
+            "lr_warmup_epochs": 0,
+            "initial_lr": 0.1,
+            "loss": "cross_entropy",
+            "optimizer": "SGD",
+            "criterion_params": {},
+            "optimizer_params": {"weight_decay": 1e-4, "momentum": 0.9},
+            "train_metrics_list": [Accuracy(), Top5()],
+            "valid_metrics_list": [Accuracy(), Top5()],
+            "metric_to_watch": "Accuracy",
+            "greater_metric_to_watch_is_better": True,
+        }
+        trainer.train(
+            model=net,
+            training_params=train_params,
+            train_loader=classification_test_dataloader(batch_size=10),
+            valid_loader=classification_test_dataloader(batch_size=10),
+        )
+        trainer.quantize(calib_dataloader=classification_test_dataloader(batch_size=10))
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -454,6 +454,8 @@ jobs:
             python3.8 -m pip install -r requirements.txt
             python3.8 -m pip install .
             python3.8 -m pip install torch torchvision torchaudio
+            python3.8 -m pip install pytorch-quantization==2.1.2 --extra-index-url https://pypi.ngc.nvidia.com
+
             python3.8 src/super_gradients/examples/train_from_recipe_example/train_from_recipe.py --config-name=coco2017_pose_dekr_w32_no_dc experiment_name=shortened_coco2017_pose_dekr_w32_ap_test batch_size=4 val_batch_size=8 epochs=1 training_hyperparams.lr_warmup_steps=0 training_hyperparams.average_best_models=False training_hyperparams.max_train_batches=1000 training_hyperparams.max_valid_batches=100 multi_gpu=DDP num_gpus=4
             python3.8 src/super_gradients/examples/train_from_recipe_example/train_from_recipe.py --config-name=cifar10_resnet experiment_name=shortened_cifar10_resnet_accuracy_test epochs=100 training_hyperparams.average_best_models=False multi_gpu=DDP num_gpus=4
             python3.8 src/super_gradients/examples/convert_recipe_example/convert_recipe_example.py --config-name=cifar10_conversion_params experiment_name=shortened_cifar10_resnet_accuracy_test
--- a/src/super_gradients/__init__.py
+++ b/src/super_gradients/__init__.py
@@ -3,6 +3,7 @@ from super_gradients.training import losses, utils, datasets_utils, DataAugmenta
 from super_gradients.common.registry.registry import ARCHITECTURES
 from super_gradients.sanity_check import env_sanity_check
 from super_gradients.training.utils.distributed_training_utils import setup_device
+from super_gradients.training.pre_launch_callbacks import AutoTrainBatchSizeSelectionCallback, QATRecipeModificationCallback
 
 __all__ = [
     "ARCHITECTURES",
@@ -18,6 +19,8 @@ __all__ = [
     "is_distributed",
     "env_sanity_check",
     "setup_device",
+    "QATRecipeModificationCallback",
+    "AutoTrainBatchSizeSelectionCallback",
 ]
 
 __version__ = "3.1.1"
--- a/src/super_gradients/recipes/roboflow_yolo_nas_s_qat.yaml
+++ b/src/super_gradients/recipes/roboflow_yolo_nas_s_qat.yaml
@@ -4,6 +4,7 @@ defaults:
   - _self_
 
 checkpoint_params:
+  checkpoint_path: ???
   strict_load: no_key_matching
 
 pre_launch_callbacks_list:
--- a/src/super_gradients/recipes/roboflow_yolo_nas_s_qat.yaml
+++ b/src/super_gradients/recipes/roboflow_yolo_nas_s_qat.yaml
@@ -4,7 +4,6 @@ defaults:
   - _self_
 
 checkpoint_params:
-  checkpoint_path: ???
   strict_load: no_key_matching
 
 pre_launch_callbacks_list:
--- a/src/super_gradients/training/qat_trainer/qat_trainer.py
+++ b/src/super_gradients/training/qat_trainer/qat_trainer.py
@@ -1,5 +1,6 @@
 import os
-from typing import Union, Tuple
+from typing import Union, Tuple, Dict, Mapping, List
+from torchmetrics import Metric
 
 import copy
 import hydra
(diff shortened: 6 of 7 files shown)


tests/recipe_training_tests/coded_qat_launch_test.py
```diff
@@ -0,0 +1,39 @@
...
+            "metric_to_watch": "Accuracy",
+            "greater_metric_to_watch_is_better": True,
+        }
+        trainer.train(
+            model=net,
+            training_params=train_params,
+            train_loader=classification_test_dataloader(batch_size=10),
+            valid_loader=classification_test_dataloader(batch_size=10),
+        )
+        trainer.quantize(calib_dataloader=classification_test_dataloader(batch_size=10))
```
Louis-Dupont, line 35: `trainer.train()` does QAT. This is very clear and intuitive.
`trainer.quantize()` does PTQ + optionally QAT. And by default it does both. I think this can be missleading, people might want to run `trainer.train()` and then `trainer.quantize()` and not think about setting `ptq_only=True`.

Maybe we should drop the name `trainer.quantize()` and instead have `trainer.post_training_quantization()`
If we still want the option to run to do both qat and ptq together, we could have something like `trainer.full_quantization()`

Louis-Dupont (commented):
Looks great!
I wrote a small note about naming (non-blocking)

Other than that I think we should also add this as a "quickstart" section in the ptq/qat tutorial: https://github.com/Deci-AI/super-gradients/blob/master/documentation/source/ptq_qat.md

tests/recipe_training_tests/coded_qat_launch_test.py
BloodAxe: Agree to @Louis-Dupont. I haven't been into QAT so all this is relatively new to me.
So as a newcomer, I'd expect `trainer.quantize` to do a quantization of the already trained model. The method name does not give any hints that training may take place. Perhaps for sake of continence we can have:

```python
def train_and_quantize(...):
   self.train(...)
   self.quantize(...)
```

Second, I believe that both Trainer and QATTrainer should both have `quantize` method.
Why? 
Regardless whether we use QAT or not, we should be able to use PTQ. For that reason, having a `quantize` method does make sense to me.
However, to get the maximum accuracy, one may employ QATTrainer which would do it's QAT-specific magic:

So user flow would be:
1) Do I want to get quantized model?
[NO] -> Use `Trainer.train()`
[YES] -> Use `Trainer.train_and_quantize()`
2) Do I want to train model with quantization-aware tricks?
[NO] -> Use `Trainer.train_and_quantize()`
[YES] -> Use `QATTrainer.train_and_quantize()`

src/super_gradients/qat_from_recipe.py
```diff
@@ -14,7 +14,7 @@
 
 @hydra.main(config_path="recipes", version_base="1.2")
 def _main(cfg: DictConfig) -> None:
-    QATTrainer.train_from_config(cfg)
+    QATTrainer.quantize_from_config(cfg)
```
Louis-Dupont, line 17: Since it is deprecated I think we should already change this script to using `Trainer`

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2018,3 +2033,341 @@ def _init_loss_logging_names(self, loss_logging_items):
...
+
+        The quantized model will be exported to ONNX along with other checkpoints.
+
+        The call to self.quantize (see docs in the next method) is done with the created
+         train_loader and valid_loader. If no calibration data loader is passed through cfg.calib_loader,
+         a train data laoder with the validation transforms is used for calibration.
+
+        :param cfg: The parsed DictConfig object from yaml recipe files or a dictionary.
+        :return: A tuple containing the quantized model and the output of trainer.train() method.
+        :rtype: Tuple[nn.Module, Tuple]
```
Louis-Dupont, line 2056: We can take out :rtype:, it's already in typing

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2018,3 +2033,341 @@ def _init_loss_logging_names(self, loss_logging_items):
...
+        """
+
+        if quantization_params is None:
+            quantization_params = load_recipe("quantization_params/default_quantization_params").quantization_params
+            logger.info(f"Using default quantization params: {quantization_params}")
+
+        selective_quantizer_params = get_param(quantization_params, "selective_quantizer_params")
+        calib_params = get_param(quantization_params, "calib_params")
+        model = model or get_param(self.ema_model, "ema") or self.net
+        model.to(device_config.device)
```
Louis-Dupont, line 2327: If `model is None` and `Trainer.train(model)` was not called before, self.net will be None.
Then, running `model.to(...)` will raise an exception

tests/recipe_training_tests/coded_qat_launch_test.py
```diff
@@ -0,0 +1,131 @@
...
+        train_params, train_dataset_params, val_dataset_params, train_dataloader_params, val_dataloader_params = modify_params_for_qat(
+            train_params, train_dataset_params, val_dataset_params, train_dataloader_params, val_dataloader_params
+        )
+
+        train_loader = cifar10_train(dataset_params=train_dataset_params, dataloader_params=train_dataloader_params)
+        valid_loader = cifar10_val(dataset_params=val_dataset_params, dataloader_params=val_dataloader_params)
+
+        trainer.ptq(model=net, valid_loader=valid_loader, calib_loader=train_loader, valid_metrics_list=train_params["valid_metrics_list"])
+
+
```
Louis-Dupont, line 129: I would add 1 or 2 tests for the cases when:
- loading from pretrained_weights
- loading from local checkpoint (might be redundant)

(I think these test would break with current implementation, see my message above on self.net)

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2018,3 +2033,341 @@ def _init_loss_logging_names(self, loss_logging_items):
...
+        self,
+        calib_loader: DataLoader,
+        model: nn.Module,
+        valid_loader: DataLoader,
+        valid_metrics_list: List[torchmetrics.Metric],
+        quantization_params: Dict = None,
+        deepcopy_model_for_export: bool = False,
+    ):
+        """
+        Performs calibration.
```
Louis-Dupont, line 2277: I think it would be better to explicitly use the phrase "Post Training Quantization" in the doc.
Maybe also add link to torch page https://pytorch.org/docs/stable/quantization.html#post-training-static-quantization (this is the PTQ static right ?)

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2018,3 +2033,341 @@ def _init_loss_logging_names(self, loss_logging_items):
...
+        valid_metrics_dict = self.test(model=model, test_loader=valid_loader, test_metrics_list=valid_metrics_list)
+        results = ["PTQ Model Validation Results"]
+        results += [f"   - {metric:10}: {value}" for metric, value in valid_metrics_dict.items()]
+        logger.info("\n".join(results))
+
+        input_shape = next(iter(valid_loader))[0].shape
+        os.makedirs(self.checkpoints_dir_path, exist_ok=True)
+        qdq_onnx_path = os.path.join(self.checkpoints_dir_path, f"{self.experiment_name}_{'x'.join((str(x) for x in input_shape))}_ptq.onnx")
+
+        # TODO: modify SG's convert_to_onnx for quantized models and use it instead
```
Louis-Dupont, line 2365: Is this for another ticket ?

Louis-Dupont (commented):
Looks great like that
I think there might be one bug though, see below, and the rest is just minor comments

tests/recipe_training_tests/coded_qat_launch_test.py
shaydeci: Dont really need to check that.
We pass a model instance, from the Trainer class perspective it doesnt matter what weights are used.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2018,3 +2033,341 @@ def _init_loss_logging_names(self, loss_logging_items):
...
+            training_params=training_params,
+            additional_configs_to_log=additional_qat_configs_to_log,
+        )
+
+        # EXPORT QUANTIZED MODEL TO ONNX
+        input_shape = next(iter(valid_loader))[0].shape
+        os.makedirs(self.checkpoints_dir_path, exist_ok=True)
+        qdq_onnx_path = os.path.join(self.checkpoints_dir_path, f"{self.experiment_name}_{'x'.join((str(x) for x in input_shape))}_qat.onnx")
+
+        # TODO: modify SG's convert_to_onnx for quantized models and use it instead
```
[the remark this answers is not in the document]
shaydeci: yes, ill open one. Dont think its urgent for now.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: yes, ill open one. Dont think its urgent for now.

src/super_gradients/qat_from_recipe.py
shaydeci: fixed in latest commit.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: done in latest commit.

src/super_gradients/training/utils/quantization/export.py
```diff
@@ -14,10 +16,14 @@
...
+    model: torch.nn.Module, onnx_filename: str, input_shape: tuple, train: bool = False, to_cpu: bool = True, deepcopy_model=False, **kwargs
+):
     """
     Method for exporting onnx after QAT.
 
+    :param deepcopy_model: Whether to export deepcopy(model). Necessary in case further training is performed and
+     prep_model_for_conversion makes the network un-trainable (i.e RepVGG blocks).
     :param to_cpu: transfer model to CPU before converting to ONNX, dirty workaround when model's tensors are on different devices
     :param train: export model in training mode
     :param model: torch.nn.Module, model to export
```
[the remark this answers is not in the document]
shaydeci: fixed in latest commit.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: fixed in latest commit.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2018,3 +2033,341 @@ def _init_loss_logging_names(self, loss_logging_items):
...
+
+        :return: Validation results of the QAT model in case quantization_params['ptq_only']=False and of the PTQ
+        model otherwise.
+        """
+
+        if quantization_params is None:
+            quantization_params = load_recipe("quantization_params/default_quantization_params").quantization_params
+            logger.info(f"Using default quantization params: {quantization_params}")
+        valid_metrics_list = valid_metrics_list or get_param(training_params, "valid_metrics_list")
+        model = model or get_param(self.ema_model, "ema") or self.net
```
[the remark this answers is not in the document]
shaydeci: fixed in latest commit.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: Added at the top of ptq docs.
In QAT method I only use the complete terms at the top, think its good enough and tiring to do so each time.

I have a ticket for writing the PTQ/QAT notebook, so its better off sending our users there rather to torch's quantization which is implemented differently (we use Nvidia's Pytorch-quantization package)

Louis-Dupont (approved):
LGTM

PR #1007, opened 2023-05-15
Added Apple CoreML compilation tutorials for any SG model

avideci:
# CoreML Compilation Support For Any SG Model
I added a notebook for CoreML compilation, since it was requested by many of our users, and because it is a low-hanging fruit.

## Convert to CoreML:
1. Select any SG model (yolo-nas-s used by default), sing model.get(...)
2. Convert to CoreML using the notebook - providing nn.Module and target checkpoint path for the CoreML model.

It uses torch tracing, which is recommended by apple as the go-to in terms of coreml source.
Note: The netron visualization does not work on colab, it assumes the notebook is running in localhost.

## To compile you SD architecture or model to CoreML:
1. Run the notebook on colab
2. Download the file from` /content/yolo-nas.mlmodel`
3. Drag to XCode
4. Use in your iOS/OS X application!

The names or the inputs will always be input_1, input_2, and so on, to make it easy on researchers.
CoreML classes use keyword arguments for inference, so having the same input names helps a lot while working with different models and trying different things.


avideci:
Pre-Reviewed notebook in colab (where I ran all cells):
https://colab.research.google.com/drive/1vqa2_TJ3rgxbNaR5n--uVfpmkEkJjbrN?usp=sharing

avideci:
Please note that is worked without calling `prep_model_for_conversion`,
We can add it if needed any maybe run regression on all the models, to see what breaks and what works.

shaydeci (changes requested):
Nice! Thanks for the contribution (:
Some notes-

- We need to call prep_model_for_conversion, otherwise we will not fuse the repvgg branches (especially important for YoloNAS).

- If we take our pre-trained YoloNAS as an example, its best that the dummy input will be 640X640.

- I think it wuold be great to add just some text before the code cells, that briefly explain what is being performed. Maybe even some of the text you added to this PR description.

avideci:
- Added text cells
- Added prep_for_conversoin
- Changed to 640x640

BloodAxe:
Great work here! 💪 
Maybe it make sense to move helper functions to 
`super_gradients.coreml` namespace?

avideci:
I added the code to super_gradients.models, just like the ONNX.
I also added tests for CoreML mlmodel / mlpackage.

I didn't want to abuse abstractions here, since we don't forecast many export functions as part of SG,
so I re-used some of the parameters (but kept their names).
Let me know if it makes sense. We can move the input preperation outside to a third function.

src/super_gradients/training/models/conversion.py
```diff
@@ -48,6 +56,109 @@ def forward(self, x):
...
+         When true, the simplified model will be saved in out_path (default=True).
+        :param export_as_ml_program: Whether to convert to the new program format (better) or legacy coreml proto file
+                            (Supports more iOS versions and devices, but this format will be deprecated at some point).
+        :param torch_trace_kwargs: kwargs for torch.jit.trace
+    :return: Path
+    """
+    if ct is None:
+        raise ImportError(
+            '"coremltools" is required for CoreML export, but is not installed. Please install CoreML Tools using:\n'
+            '   "python3 -m pip install coremltools" and try again (Tested with version 6.3.0);'
```
BloodAxe, line 92: Is there a reason we a are using -m syntax  and explicit python3 here? If not, I suggest to follow default installation from the official repo https://github.com/apple/coremltools#version-6

src/super_gradients/training/models/conversion.py
```diff
@@ -48,6 +56,109 @@ def forward(self, x):
...
+    logger.debug(model)
+    logger.debug("Model child nodes:")
+    logger.debug(next(model.named_children()))
+
+    if not os.path.isdir(pathlib.Path(out_path).parent.resolve()):
+        raise FileNotFoundError(f"Could not find destination directory {out_path} for the ONNX file.")
+    torch_trace_kwargs = torch_trace_kwargs or dict()
+    prep_model_for_conversion_kwargs = prep_model_for_conversion_kwargs or dict()
+
+    if input_shape is not None:
```
BloodAxe, line 105: Since this is a new function, I don't think adding a deprecated parameter to it is a right idea. To me it make sense to drop it. I realize this is to mimic the API of `convert_to_onnx`, but probably it's time to follow our own promises and at least not introduce a new deprecated stuff.
@shaydeci @Louis-Dupont WDYT?

tests/unit_tests/export_onnx_test.py
```diff
@@ -15,6 +15,7 @@ def test_models_onnx_export_with_deprecated_input_shape(self):
         with tempfile.TemporaryDirectory() as tmpdirname:
             out_path = os.path.join(tmpdirname, "resnet18.onnx")
             models.convert_to_onnx(model=pretrained_model, out_path=out_path, input_shape=(3, 256, 256), pre_process=preprocess)
+            self.assertTrue(os.path.exists(out_path))
```
BloodAxe, line 18: I think you may want to use `os.path.isfile` here, since `os.path.exists` would return true if `out_path` would be a directory.

tests/unit_tests/export_onnx_test.py
```diff
@@ -24,6 +25,7 @@ def test_models_onnx_export(self):
             models.convert_to_onnx(
                 model=pretrained_model, out_path=out_path, pre_process=preprocess, prep_model_for_conversion_kwargs=dict(input_size=(1, 3, 640, 640))
             )
+            self.assertTrue(os.path.exists(out_path))
```
BloodAxe, line 28: Same here with os.path.isfile

src/super_gradients/training/models/conversion.py
avideci: python3 -m pip is the perefferable way to use pip, in general, not just here. 
When executin pip/pip3 you don't control the interpreter being used. 
In python, it is always better to be explicit.
You might have more than 2 python 3.X interpretrs installed.
Python3 -m pip let's OS determine which one to use.

'pip' as binary or command is implicit, might use other interpreters and venvs, and so on.
Long story short - use python3 -m pip whenver possible to avoid environment installation bugs.

Plus, I believe users understand that's replecable with pip install.

src/super_gradients/training/models/conversion.py
avideci: Why are you saing input_shape is deprecated?

avideci: Oh I see now. 
Whatever you think...

tests/unit_tests/export_onnx_test.py
avideci: We are using .exists() all over the codebase, nt on directories (see logging filenmes as an example).
I will fix that. Just letting you know it is everywhere and not a convention. 

If it is bugging us, let's fix all the occurenced in the codebase that are not directories and use .exists()

tests/unit_tests/export_onnx_test.py
avideci: Fixed

src/super_gradients/training/models/conversion.py
avideci: Fixed - Removed shape and using only "input_size" in CoreML

ofrimasad (approved):
LGTM

Louis-Dupont (approved):
LGTM

avideci:
Squashed and signed in https://github.com/Deci-AI/super-gradients/pull/1068/commits/62936fc7d32e5330bb2bb969099118c2fcfa6d65

PR #1009, opened 2023-05-15
Cache model from platform locally

BloodAxe:


src/super_gradients/common/plugins/deci_client.py
```diff
@@ -54,9 +61,10 @@ def _get_file(self, model_name: str, file_name: str) -> Optional[str]:
         :return:            Path were the downloaded file was saved to. None if not found.
         """
         try:
-            response = self.lab_client.get_autonac_model_file_link(
+            response, status, response_headers = self.lab_client.get_autonac_model_file_link_with_http_info(
```
shaydeci, line 64: I assume we need to update our pro requirements then ?

shaydeci (changes requested):
I think the recent changes are just in the newer versions of client right?
Lets update the pro requirements and then LGTM.

src/super_gradients/common/plugins/deci_client.py
```diff
@@ -54,9 +61,10 @@ def _get_file(self, model_name: str, file_name: str) -> Optional[str]:
...
                 model_name=model_name, file_name=file_name, super_gradients_version=super_gradients.__version__
             )
+            etag = response_headers.get("etag", "")
```
Louis-Dupont, line 67: Do they provide an etag?

src/super_gradients/common/plugins/deci_client.py
BloodAxe: According to the versions in pro requirements (deci-lab-client==4.17.0) - it's already there. 
In fact get_autonac_model_file_link using get_autonac_model_file_link_with_http_info under the hood. So no version update needed IMO

src/super_gradients/common/plugins/deci_client.py
BloodAxe: Yep, https://github.com/Deci-AI/super-gradients/pull/1009#discussion_r1193677766

src/super_gradients/common/plugins/deci_client.py
BloodAxe: Updated deci-lab-client to 4.29 since there was some import error for BodyRegisterUserArchitecture request model.

shaydeci (dismissed):
LGTM

shaydeci (approved):

Issue #1008, opened 2023-05-15
Save model checkpoints using Weights & Biases Artifacts


soumik12345:
Currently, model checkpoints are saved using `wandb.save()` in [`WandBSGLogger`](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/common/sg_loggers/wandb_sg_logger.py#L244). It will be better to save the model checkpoints using [Artifacts](https://docs.wandb.ai/guides/artifacts) instead.

- Logging checkpoints as artifacts would enable checkpoints to be automatically versioned.
- It would also make it possible to register checkpoints as models enabling users to use the [Model Registry](https://docs.wandb.ai/guides/models).
- This would make the checkpoints easier and more accessible to further [usages](https://docs.wandb.ai/guides/artifacts/download-and-use-an-artifact) like evaluation, inference, and further training.
- This would also ensure its [lineage](https://docs.wandb.ai/guides/app/pages/project-page#lineage-panel) is tracked automatically.

Issue #980, opened 2023-05-11
Weights & Biases initialization should be decoupled from the Logger


soumik12345:
Currently, a Weights & Biases run is initialized inside the constructor of [`WandBSGLogger`](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/common/sg_loggers/wandb_sg_logger.py#L100). However, if the run initialization is not decoupled from the logger, it would not be possible for a user to use [Weights & Biases Artifacts](https://docs.wandb.ai/guides/artifacts) for storage and versioning of their datasets. Hence, `wandb.init()` should be **optionally** decoupled from `WandBSGLogger`.

The ideal workflow should be:

```python

wandb.init(project=..., entity=..., **kwargs)

artifact = wandb.use_artifact('geekyrakshit/yolo-nas-integration/bdd100k-subset-yolo:v2', type='dataset')
artifact_dir = artifact.download()

train_data, val_data = create_dataset(artifact_dir)
model = ...

trainer = Trainer(experiment_name=..., ckpt_root_dir=...)
trainer.train(model=model, training_params=train_params, train_loader=train_data, valid_loader=val_data)
```

BloodAxe:
I see your point, however it is not quite how we thought users would use SG ^_^
We build SG to rely heavily on use of recipes, so for that purpose initialization of logger inside trainer made perfect sense. 

What we can do is a small change inside the [WandBSGLogger](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/common/sg_loggers/wandb_sg_logger.py#L100) to check whether there is active run (`wandb.run is not None`) and skip call to `wandb.init(...)` if it was already made. 

Although it looks simple from the first glance, I feel there could be edge cases with resume (?) @shaydeci @Louis-Dupont what you guys think about it?

soumik12345:
> WandBSGLogger

Hi @BloodAxe 
I think this could be resolved by changing [this line](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/common/sg_loggers/wandb_sg_logger.py#L100) to

```python
run = wandb.init(
    project=project_name, name=experiment_name, entity=entity, resume=resumed, id=wandb_id, **kwargs
) if  wandb.run is None else wandb.run
```

This would mean that if a run is already initialized, `WandBSGLogger` won't attempt to initialize a run again, else it would be automatically initialized as per the default behaviour. Would love to know your thoughts on this.

[...]

Issue #953, opened 2023-05-09
Log inference results to Weights & Biases


soumik12345:
Optionally log inference results in [Weights & Biases](https://wandb.ai/site) dashboard using [bounding-box image overlays](https://docs.wandb.ai/guides/track/log/media#image-overlays).

Add an optional parameter `visualize_on_wandb` to `ImagesDetectionPrediction.show()` which when set to `True` would log the predictions using the [bounding-box image overlays](https://docs.wandb.ai/guides/track/log/media#image-overlays) in the [Weights & Biases](https://wandb.ai/site) dashboard of the user.

Here's a sample code:

```python
from super_gradients.training import models
from super_gradients.common.object_names import Models

import wandb

# Initialize Weights & Biases run
wandb.init(project="yolo-nas", entity="geekyrakshit")

net = models.get(Models.YOLO_NAS_S, pretrained_weights="coco")
prediction = net.predict([
    "https://cdn.autonomous.ai/static/upload/images/common/upload/20210919/5-Desk-Setup-with-a-Laptop--Monitor-Ideas-for-2021_10f89201411b.jpg",
    "https://p.imgci.com/db/PICTURES/CMS/352000/352019.jpg",
    "https://p.imgci.com/db/PICTURES/CMS/352200/352298.jpg",
    "https://www.encirclephotos.com/wp-content/uploads/Laos-Ban-Pak-Ou-Man-Riding-Asian-Elephant-1440x961.jpg"
])
prediction.show(visualize_on_wandb=True)
```

This would log the images to be visualized in a media panel in a Weights & Biases dashboard.

Here's how the results would look 👇
[![](https://raw.githubusercontent.com/wandb/assets/main/wandb-github-badge-28-gray.svg)](https://wandb.ai/geekyrakshit/yolo-nas-integration/runs/lz9mjhwy)

Here's a video of how to interact with the bounding-box overlay interface on Weights & Biases 👇

https://github.com/Deci-AI/super-gradients/assets/19887676/fb027d35-adc1-4e93-9d5a-2ea7b32c50f7

PR #1010, opened 2023-05-15
Update Weights & Biases Integration + Add Prediction Visualization for Object Detection

soumik12345:
# Changes made in this PR

## Log inference results to Weights & Biases

Fixes https://github.com/Deci-AI/super-gradients/issues/953

Optionally log inference results for object detection in [Weights & Biases](https://wandb.ai/site) dashboard using [bounding-box image overlays](https://docs.wandb.ai/guides/track/log/media#image-overlays).

This PR adds an optional parameter `visualize_on_wandb` to `ImagesDetectionPrediction.show()` which when set to `True` would log the predictions using the [bounding-box image overlays](https://docs.wandb.ai/guides/track/log/media#image-overlays) in the [Weights & Biases](https://wandb.ai/site) dashboard of the user.

Here's a sample code:

```python

import wandb

# Initialize Weights & Biases run
wandb.init(project="yolo-nas", entity="geekyrakshit")

net = models.get(Models.YOLO_NAS_S, pretrained_weights="coco")
prediction = net.predict([
    "https://cdn.autonomous.ai/static/upload/images/common/upload/20210919/5-Desk-Setup-with-a-Laptop--Monitor-Ideas-for-2021_10f89201411b.jpg",
    "https://p.imgci.com/db/PICTURES/CMS/352000/352019.jpg",
    "https://p.imgci.com/db/PICTURES/CMS/352200/352298.jpg",
    "https://www.encirclephotos.com/wp-content/uploads/Laos-Ban-Pak-Ou-Man-Riding-Asian-Elephant-1440x961.jpg"
])
prediction.show(visualize_on_wandb=True)
```

This would log the images to be visualized in a media panel in a Weights & Biases dashboard.

👇 Here's how the results would look
[](https://wandb.ai/geekyrakshit/yolo-nas-integration/runs/lz9mjhwy)

Here's a video of how to interact with the bounding-box overlay interface on Weights & Biases 👇

https://github.com/Deci-AI/super-gradients/assets/19887676/fb027d35-adc1-4e93-9d5a-2ea7b32c50f7

## Update `WandBSGLogger`

Fixes https://github.com/Deci-AI/super-gradients/issues/980
Fixes https://github.com/Deci-AI/super-gradients/issues/1008

- `WandBSGLogger` will now call `wandb.init()` only if it has not been called earlier.
- Save model checkpoints versioned with Weights & Biases checkpoints.
- Add an additional parameter `sync_tensorboard` to `WandBSGLogger` that automatically syncs TensorBoard with the wandb run.

👇 Here's a sample training code that works with the changes made in this PR
[](https://colab.research.google.com/drive/1478GGwddwiyWJzb6umfVH5bNl8unx0UR?usp=sharing)

👇 Here's a sample Weights & Biases Run demonstrating the features
[](https://wandb.ai/geekyrakshit/yolo-nas-integration/runs/n1trro70)

👇 Tensorboard instance synced with Weights & Biases Run
[](https://wandb.ai/geekyrakshit/yolo-nas-integration/runs/n1trro70/tensorboard)

👇 Lineage of Model Checkpoints
[](https://wandb.ai/geekyrakshit/yolo-nas-integration/artifacts/model/n1trro70-checkpoint/27fc3b1a6c611bd3dc70/lineage)


soumik12345:
Hi @BloodAxe 
Can you please take a look at this PR?

src/super_gradients/wandb/artifacts.py
```diff
@@ -0,0 +1,13 @@
+import os
+import wandb
```
ofrimasad, line 2: Please use the same import logic we used in the WandbLogger 
```
try:
    import wandb
except (ModuleNotFoundError, ImportError, NameError):
    pass  # no action or logging - this is normal in most cases
```
then check if `wandb is not None`

src/super_gradients/wandb/artifacts.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def save_artifact(path):
```
ofrimasad, line 5: If you define the type to be model, please name the function `save_wandb_model_artifact`
otherwise, call it `save_wandb_artifact(path: str, type: str)`.

note the added explicit typing

ofrimasad: please add documentation:
something like:
```
 upload and wandb Artifact.
Note that this function can be called only after wandb.init()
 :param path: the local full path to the pth file to be uploaded 
 ```

File: src/super_gradients/wandb/log_predictions.py [hunk 3]
```diff
@@ -0,0 +1,40 @@
+import wandb
```
ofrimasad, line 1: same here regarding the import please

src/super_gradients/common/sg_loggers/wandb_sg_logger.py
```diff
@@ -102,7 +105,11 @@ def __init__(
                     )
                 wandb_id = self._get_wandb_id()
 
-        run = wandb.init(project=project_name, name=experiment_name, entity=entity, resume=resumed, id=wandb_id, **kwargs)
+        run = (
+            wandb.init(project=project_name, name=experiment_name, entity=entity, resume=resumed, id=wandb_id, sync_tensorboard=sync_tensorboard, **kwargs)
```
ofrimasad, line 109: why do we need to synch tensorboard if all information is logged to wandb? Wouldn't this create duplicated data?

src/super_gradients/common/sg_loggers/wandb_sg_logger.py
```diff
@@ -102,7 +105,11 @@ def __init__(
...
+            if wandb.run is None
```
ofrimasad, line 110: This is breaking the resume experiment feature.
Moreover, if the run was initialized from outside, we have no way of knowing it is logging into the correct experiment. 
Why would you want to init wandb from outside of the logger?

src/super_gradients/common/sg_loggers/wandb_sg_logger.py
```diff
@@ -246,7 +253,7 @@ def add_checkpoint(self, tag: str, state_dict: dict, global_step: int = 0):
         if self.save_checkpoints_wandb:
             if self.s3_location_available:
                 self.model_checkpoints_data_interface.save_remote_checkpoints_file(self.experiment_name, self._local_dir, name)
-            wandb.save(glob_str=path, base_path=self._local_dir, policy="now")
+            save_artifact(path)
```
ofrimasad, line 256: will this have the same effect? 
currently all files are saved inside the wandb experiment. Will the artifact be saved in the same way?

src/super_gradients/wandb/log_predictions.py
```diff
@@ -0,0 +1,40 @@
...
+
+from super_gradients.training.models.prediction_results import ImageDetectionPrediction, ImagesDetectionPrediction
+
+
+def visualize_image_detection_prediction_on_wandb(prediction: ImageDetectionPrediction, show_confidence: bool):
```
ofrimasad, line 6: please make this function private (start with _)

src/super_gradients/wandb/log_predictions.py
```diff
@@ -0,0 +1,40 @@
...
+            box["scores"] = {"confidence": float(round(prediction.prediction.confidence[pred_i], 2))}
+        boxes.append(box)
+
+    wandb_image = wandb.Image(image, boxes={"predictions": {"box_data": boxes, "class_labels": class_id_to_labels}})
+
+    wandb.log({"Predictions": wandb_image})
+
+
+def log_detection_results_to_wandb(prediction: ImagesDetectionPrediction, show_confidence: bool = True):
+    """Log predictions for object detection to Weights & Biases using interactive bounding box overlays.
```
ofrimasad, line 34: If I am not mistaken, this has to be called after `wandb.init()` please state this in the documentation and add an if-raise statement

src/super_gradients/wandb/artifacts.py
ofrimasad: looks like this code is used only inside the `WandBSGLogger`
why not include this as a private function of `WandBSGLogger`?

src/super_gradients/common/sg_loggers/wandb_sg_logger.py
soumik12345: Hi @ofrimasad 
You can refer to https://github.com/Deci-AI/super-gradients/pull/1010#issuecomment-1561244431 regarding this.

src/super_gradients/common/sg_loggers/wandb_sg_logger.py
ofrimasad: Assume I trained my model fro 10 epochs and then stopped it for some reason. now I want to resume the training.
I can just use the `resume` flag in the training params and the wandb run will be connected to the right run ID.
but if I init the run from outside, how would I do this? I need to go to wandb, check the run Id, init the run with the correct run Id and the resume flag. If I just call wandb.init() without the id, the run will not resume.

src/super_gradients/common/sg_loggers/wandb_sg_logger.py
ofrimasad: After testing this, I can see two issues with this change:

1. it dramatically inflates the storage required by the user. Instead of saving only the last version of the model, it saves *all* the checkpoints. Users that are using an on-prem wandb server (like we do) will see a huge increase in storage usage.  

2. you are uploading different files to the same Artifact. During the training, we save ckpt_best.pth, ckpt_latest.pth, ckpt_average.pth. and this code maps all of them to the same Artifact (as different versions)  /

Please add a parameter (in the init), so the users can choose if they want the files to be saved as Artifact or not. 
please differentiate between files with different names.

ofrimasad (commented):
I see that you still kept the option to init wandb from outside the logger.
In that case please add a warning that will be presented if the run was initialized from outside the logger. 
"The wandb run was initialized from outside of the WandbSGLogger. This means that SG cannot control the run ID to which this session will be logged. The current run id is: {run.id}"

Thanks

ofrimasad (commented):
please use the common logging behavior of SG

soumik12345:
Hi @ofrimasad is there any more changes we need to make as part of this PR?

PR #1047, opened 2023-05-21
Update deci-lab-client usage

roikoren755:


src/super_gradients/common/plugins/deci_client.py
```diff
@@ -45,38 +52,37 @@ class DeciClient:
...
+                    "Unauthorized. wrong credentials or credentials not defined. "
+                    "Please provide credentials via environment variables (DECI_CLIENT_ID, DECI_CLIENT_SECRET)"
                 )
             elif e.status == 400 and e.body is not None and "message" in e.body:
                 logger.error(f"Deci client: {json.loads(e.body)['message']}")
             else:
                 logger.debug(e.body)
             return None
-
+        etag = "123"
```
Louis-Dupont, line 85: @roikoren755 I hard coded `etag = "123"` to test, but once the new version is released we need to get etag from
`download_link, etag = self.lab_client.get_autonac_model_file_link` as discussed

shaydeci (approved):
LGTM

Issue #1061, opened 2023-05-23
Bounding boxes are drawn in random order instead of ordered by confidence


laszlovandenhoek:
## Problem

This is the method that draws bounding boxes around predictions:

https://github.com/Deci-AI/super-gradients/blob/a59449726e1d55f68c96b7be24f8a00a126b59e0/src/super_gradients/training/models/prediction_results.py#L57-L85

However, the code does not take into account that bounding boxes are drawn on top of each other. Since the arrays are not in any particular order, if an object has multiple overlapping predictions, it can happen that a less likely prediction is drawn later, obscuring the more likely prediction in the resulting image.

## Solution

This could be fixed by changing this line:

https://github.com/Deci-AI/super-gradients/blob/a59449726e1d55f68c96b7be24f8a00a126b59e0/src/super_gradients/training/models/prediction_results.py#L69

into:

```python
for pred_i in np.argsort(self.prediction.confidence):
```

This way, the most likely prediction is drawn last, i.e. on top of any overlapping other ones, and is therefore the most visible in the resulting image.

## Example

The examples below were created using this code:

```python
from super_gradients.training import models
from super_gradients.common.object_names import Models

net = models.get(Models.YOLO_NAS_L, pretrained_weights="coco")

prediction = net.predict("https://www.krijskraan.nl/catbird-input.mp4")

prediction.save(output_path="catbird-output.mp4")
```

As you can see in these examples, there is less "flashing" in the resulting movie. Also, the fact that the most likely class for this animal is "bird" more often than "cat" immediately becomes apparent, highlighting the necessity for finetuning.

Before (=current code): https://krijskraan.nl/catbird-output-original.mp4
After ordering by confidence (=with proposed change): https://krijskraan.nl/catbird-output-ordered.mp4

### Versions

PyTorch version: 1.12.1+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.2 LTS (x86_64)
GCC version: (Ubuntu 11.3.0-1ubuntu1~22.04.1) 11.3.0
Clang version: Could not collect
CMake version: version 3.26.0
Libc version: glibc-2.35

Python version: 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] (64-bit runtime)
Python platform: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Is CUDA available: False
[...]

PR #1063, opened 2023-05-23
Draw most likely bbox on top (#1061)

laszlovandenhoek:
This is the proposed fix for issue #1061


Louis-Dupont (approved):
Great idea, LGTM!

PR #1092, opened 2023-05-26
Feature/sg 815 fix override dataset params

BloodAxe:
This PR fixes interpolation issue in our dataset configs.
There are two distinct use cases which are almost mutually exclusive, but looks like I managed to get them both working.

## Scenario 1: Colab users

In colab, users override params via dataloader factory methods as follows:

```
dataset = coco2017_train_yolo_nas(dataset_params=dict(data_dir="e:/coco2017", input_dim=[320,320]))
```

Under the hood, we load dataset config from YAML file and apply user-provided overrides.

## Scenario 2: Pycharm / Command line users

In command-line users are launching training with `train_from_recipe` and instantiation of the datasets is somewhat crooked. Since we still calling `coco2017_train_yolo_nas`, but we do a double work of merging a configuration for dataset.

In the call to `coco2017_train_yolo_nas(dataset_params)`  the `dataset_params` variable comes from experiment configuration file  with all the interpolations/instantiations done by hydra for us.
And inside the function `coco2017_train_yolo_nas` still loads `default` dataset params from YAML file.
And them we do the "merging".

The interpolation did not work as intended because of wrong order of instantiation/merging/overrides.
Looks like this PR addresses both scenarios. There are additional unit test cases which tests new logic of merging configs.
I've also ran full set of unit & integration tests on tzags

## Not fixed yet 
(And I'm not sure how)
The COCO YOLO-NAS dataset recipe has this strange image preprocessing of resizing image to longest size of 636 and then padding to 640. Why this is the problem:
*We cannot use interpolation in this case for obvious reasons*
```yaml
val_dataset_params:
  ...
  input_dim: [636, 636]
  transforms:
    - DetectionPadToSize:
        output_size: [640, 640]
    ...
    - DetectionTargetsFormatTransform:
        input_dim: [640, 640]
```

Solution 1: Change input_dim: [636, 636] to input_dim: [640, 640]
Pros: Can use interpolation
Cons: Lose a few decimals of mAP

Solution 2: Leave as is
Pros: None? 
Cons: This recipe would not support changing image resolution via command line or colab

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,8 @@
+unit_tests:
+	python -m unittest tests/deci_core_unit_test_suite_runner.py
+
+integration_tests:
+	python -m unittest tests/deci_core_integration_test_suite_runner.py
+
+yolo_nas_integration_tests:
+	python -m unittest tests/integration_tests/yolo_nas_integration_test.py
--- a/src/super_gradients/training/dataloaders/dataloaders.py
+++ b/src/super_gradients/training/dataloaders/dataloaders.py
@@ -3,6 +3,7 @@ from typing import Dict, Mapping
 import hydra
 import numpy as np
 import torch
+from omegaconf import OmegaConf
 from torch.utils.data import BatchSampler, DataLoader, TensorDataset, RandomSampler
 
 import super_gradients
@@ -81,14 +82,24 @@ def get_data_loader(config_name: str, dataset_cls: object, train: bool, dataset_
     return dataloader
 
 
-def _process_dataset_params(cfg, dataset_params, train):
-    default_dataset_params = cfg.train_dataset_params if train else cfg.val_dataset_params
-    default_dataset_params = hydra.utils.instantiate(default_dataset_params)
-    for key, val in default_dataset_params.items():
-        if key not in dataset_params.keys() or dataset_params[key] is None:
-            dataset_params[key] = val
+def _process_dataset_params(cfg, dataset_params, train: bool):
+    # Due to some internal details of Hydra, we need to merge the dataset_params with the default dataset params while retaining the
+    # top-level config structure.
+    # We can't do:
+    # >>> default_dataset_params = cfg.train_dataset_params if train else cfg.val_dataset_params
+    # because it will break interpolation down the road.
+    if train:
+        cfg.train_dataset_params = OmegaConf.merge(cfg.train_dataset_params, dataset_params)
+    else:
+        cfg.val_dataset_params = OmegaConf.merge(cfg.val_dataset_params, dataset_params)
 
-    return dataset_params
+    # default_dataset_params = cfg.train_dataset_params if train else cfg.val_dataset_params
+    # default_dataset_params = hydra.utils.instantiate(default_dataset_params)
+    # for key, val in default_dataset_params.items():
+    #     if key not in dataset_params.keys() or dataset_params[key] is None:
+    #         dataset_params[key] = val
+
+    return hydra.utils.instantiate(cfg.train_dataset_params if train else cfg.val_dataset_params)
 
 
 def _process_dataloader_params(cfg, dataloader_params, dataset, train):
--- a/src/super_gradients/recipes/dataset_params/coco_detection_yolo_nas_dataset_params.yaml
+++ b/src/super_gradients/recipes/dataset_params/coco_detection_yolo_nas_dataset_params.yaml
@@ -31,7 +31,7 @@ train_dataset_params:
         prob: 0.5                       # probability to apply per-sample mixup
         flip_prob: 0.5                  # probability to apply horizontal flip
     - DetectionPaddedRescale:
-        input_dim: [640, 640]
+        input_dim: ${dataset_params.train_dataset_params.input_dim}
         max_targets: 120
         pad_value: 114
     - DetectionStandardize:
--- a/tests/unit_tests/detection_dataset_test.py
+++ b/tests/unit_tests/detection_dataset_test.py
@@ -1,6 +1,7 @@
 import unittest
 from pathlib import Path
 
+from super_gradients.training.dataloaders import coco2017_train_yolo_nas
 from super_gradients.training.datasets import COCODetectionDataset
 from super_gradients.training.exceptions.dataset_exceptions import DatasetValidationException, ParameterMismatchException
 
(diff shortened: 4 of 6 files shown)


shaydeci:
> 

Nice!
Regarding the COCO config issue you brought up - did we check how significant is the mAP decrease? I mean its such a tiny difference...

Also, from your command example:

 dataset_params.train_dataset_params.input_dim.0=320
 dataset_params.train_dataset_params.input_dim.1=320
 
 I think the transforms that Interpolate input_dim should handle single int values and turning them into [input_dim, input_dim]. I don't think we have too much of them in use tbh, and some already have this option implemetnted.
 But...I do vaguely remember there might have been issues with overriding different data types inside hydra..

Should we also add the interpolations for our other detection recipes while at it?

Makefile
```diff
@@ -0,0 +1,8 @@
+unit_tests:
```
shaydeci, line 1: Whats this?

src/super_gradients/training/dataloaders/dataloaders.py
```diff
@@ -81,14 +82,24 @@ def get_data_loader(config_name: str, dataset_cls: object, train: bool, dataset_
...
+    # top-level config structure.
+    # We can't do:
+    # >>> default_dataset_params = cfg.train_dataset_params if train else cfg.val_dataset_params
+    # because it will break interpolation down the road.
+    if train:
+        cfg.train_dataset_params = OmegaConf.merge(cfg.train_dataset_params, dataset_params)
+    else:
+        cfg.val_dataset_params = OmegaConf.merge(cfg.val_dataset_params, dataset_params)
+
+    # default_dataset_params = cfg.train_dataset_params if train else cfg.val_dataset_params
```
shaydeci, line 96: Left on purpose?

shaydeci (changes requested):

Makefile
BloodAxe: This is to allow us easily run unit & integration tests locally on demand. `make unit_tests` to quickly run all unit tests. On linux there is built-in tab completion for Makefiles. So you type make [tab] and get suggestions. I also see this could be extended in future with `make all_tests`, `make docs`, etc.
But I can drop it if you feel this should be moved away from this PR

src/super_gradients/training/dataloaders/dataloaders.py
BloodAxe: Yes, I imagine having (for some time) old config merging code can be helpful if we ran into some unexpected problems with `OmegaConf.merge`. For this purpose I left this code block commented to save some time for one debugging it.

BloodAxe:
> Nice!
Regarding the COCO config issue you brought up - did we check how significant is the mAP decrease? I mean its such a tiny difference...

It's like 5155 -> 51.25 for M and 0.5222 -> 0.517 for L if I recall correctly. It is small, but non-negligible...





Also, from your command example:

dataset_params.train_dataset_params.input_dim.0=320
dataset_params.train_dataset_params.input_dim.1=320

I think the transforms that Interpolate input_dim should handle single int values and turning them into [input_dim, input_dim]. I don't think we have too much of them in use tbh, and some already have this option implemetnted.
But...I do vaguely remember there might have been issues with overriding different data types inside hydra..

Should we also add the interpolations for our other detection recipes while at it?

src/super_gradients/training/dataloaders/dataloaders.py
```diff
@@ -81,14 +82,24 @@ def get_data_loader(config_name: str, dataset_cls: object, train: bool, dataset_
...
+def _process_dataset_params(cfg, dataset_params, train: bool):
+    # Due to some internal details of Hydra, we need to merge the dataset_params with the default dataset params while retaining the
+    # top-level config structure.
+    # We can't do:
+    # >>> default_dataset_params = cfg.train_dataset_params if train else cfg.val_dataset_params
+    # because it will break interpolation down the road.
+    if train:
+        cfg.train_dataset_params = OmegaConf.merge(cfg.train_dataset_params, dataset_params)
+    else:
+        cfg.val_dataset_params = OmegaConf.merge(cfg.val_dataset_params, dataset_params)
```
Louis-Dupont, line 94: Does this work when `dataset_params` includes already instantiated objects?
e.g. 
``` python
transforms = [DetectionMosaic(...), DetectionRandomAffine(...), ...]
dataloaders.get('some_dataset', dataset_params={'transforms': transforms})
```

Louis-Dupont (commented):
Concerning `input_dim`, there is a third option which is to have some sort of `_resizing_dim` and `_padding_dim` but only one public param `input_dim`

- By default `input_dim=None` and we use `_resizing_dim=633`, `_padding_dim=640`.
- When setting `input_dim`, it overrides both  `_resizing_dim` and `_padding_dim` accordingly.

Kind of like `resizing_dim = input_dim if input_dim else _resizing_dim`



**Pro**
- Keep our results in the default case
- Single param to override when someone wants to set input_dim

**Con**
- Increase the complexity of the recipe (and/or resolver if we need to use them to do this `if ... else ...` logic)

Not sure we want it to go for that, it might not be worth it if we're fine with losing this few digit in MaP. But if we do care about these few digits I think it might be worth considering this approach

src/super_gradients/training/dataloaders/dataloaders.py
BloodAxe: I haven't tested this scenario, tbh.

BloodAxe: 😢 😢 😢  No it's not
```
```

BloodAxe: I found the solution, cannot say I'm very proud of it, but it works on all test cases 🙈

src/super_gradients/training/dataloaders/dataloaders.py
```diff
@@ -81,14 +82,40 @@ def get_data_loader(config_name: str, dataset_cls: object, train: bool, dataset_
...
+
+    except UnsupportedValueType:
+        # This is somewhat ugly fallback for the case when the user provides overrides for the dataset params
+        # that contains non-primitive types (E.g instantiated transforms).
+        # In this case interpolation is not possible so we just override the default params with the user-provided ones.
+        default_dataset_params = hydra.utils.instantiate(cfg.train_dataset_params if train else cfg.val_dataset_params)
+        for key, val in default_dataset_params.items():
+            if key not in dataset_params.keys() or dataset_params[key] is None:
+                dataset_params[key] = val
+        return dataset_params
```
Louis-Dupont, line 119: Well I guess that's the best we can get

src/super_gradients/training/dataloaders/dataloaders.py
```diff
@@ -81,14 +82,40 @@ def get_data_loader(config_name: str, dataset_cls: object, train: bool, dataset_
...
```
Louis-Dupont, line 118: Well I guess that's the best we can get

Louis-Dupont (approved):
LGTM

Louis-Dupont (approved):
LGTM

shaydeci (approved):
Feeling a bit unease about the fact that users wont know in case they mixed non-primitve instantiated objects and some values they wish to interpolate, but tbh I really dont have a better solution \: 

LGTM

BloodAxe:
I can think only of one case where this may go wrong:
When user combines config-based transform declaration and already instantiated instances of transform object in `transforms`:
```
dataset_params={
"input_dim": 512,
"transforms": [
 DetectionRandomAffine(...),
 {"DetectionPaddedRescale": {"input_dim": "${dataset_params.train_dataset_params.input_dim}" }}
]
```

Frankly speaking I'm not sure whether we should ever allow this mixing.

PR #1107, opened 2023-05-31
Hotfix/alg 1470 drop boxes padding

yurkovak:
**Removed max_targets param from:**
* `DetectionTargetsFormatTransform`
* `DetectionPaddedRescale`, it didn't do anything anyway 
* `DetectionHorizontalFlip` for the case of empty targets, because there won't be a collate to undo it 

**Updated:**
* detection collates (crowd and without, usual and for PPYolo);
* all places that were passing max_targets (yamls, tests, notebook)

**Trained:**
* yolox
* ppyolo

(both just to see train/val epoch runs, didn't do metric checks)


documentation/source/ObjectDetection.md
```diff
@@ -310,9 +309,7 @@ class MyNewDetectionDataset(DetectionDataset):
             lines = targets_file.read().splitlines()
             target = np.array([x.strip().strip(',').split(',') for x in lines], dtype=np.float32)
 
-        res_target = np.zeros((self.max_targets, 5))  # cls, cx, cy, w, h
```
BloodAxe, line 313: Same here, if `max_targets` passed explicitly - the old functionality should be preserved for now.

yurkovak: I'll update transforms with the warning, asking about the notebook

BloodAxe: Code examples in notebooks & docs are same-class citizens as regular code. To some extent they are even more important, since this is what users will see first when searching.
So yes, they should reflect up-to-date, recommended way of using SG.

documentation/source/ObjectDetection.md
```diff
@@ -259,12 +259,11 @@ MY_CLASSES = ['cat', 'dog', 'donut']
 @register_dataset("MyNewDetectionDataset")
 class MyNewDetectionDataset(DetectionDataset):
     def __init__(self, data_dir: str, samples_dir: str, targets_dir: str, input_dim: Tuple[int, int],
-                 transforms: List[DetectionTransform], max_targets: int = 100, max_num_samples: int = None,
```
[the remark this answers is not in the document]
yurkovak: Done 7568f478

yurkovak:
Thanks for the fast review, addressed

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -664,30 +663,42 @@ class DetectionCollateFN:
...
-        targets_merged = []
-        for i in range(targets.shape[0]):
-            targets_im = targets[i, : nlabel[i]]
-            batch_column = targets.new_ones((targets_im.shape[0], 1)) * i
-            targets_merged.append(torch.cat((batch_column, targets_im), 1))
-        return torch.cat(targets_merged, 0)
+    def _format_images(self, images_batch: List[Union[torch.Tensor, np.array]]) -> torch.Tensor:
+        images_batch = [torch.tensor(img) for img in images_batch]
+        images_batch_stack = torch.stack(images_batch, 0)
+        if images_batch_stack.shape[3] == 3:
```
BloodAxe, line 672: Let's not introduce HWC->CHW here, since it's quite unexpected responsibility of collate fn. We have permutation transform that does this in an explicit manner.

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -664,30 +663,42 @@ class DetectionCollateFN:
     """
 
     def __call__(self, data) -> Tuple[torch.Tensor, torch.Tensor]:
-        batch = default_collate(data)
-        ims, targets = batch[0:2]
-        return ims, self._format_targets(targets)
+        images_batch, labels_batch = list(zip(*data))
```
Louis-Dupont, line 666: `images_batch, labels_batch = list(zip(*data))` will only work if each sample of data is a tuple of 2 elements. It's true that we should have 2, but if someone uses for instance a dataset with crowd_target=True, and eventually uses `DetectionCollateFN` then the code would crash (there might be other scenarios)

I think the way to do would be something like this:
```python
def __call__(self, data) -> Tuple[torch.Tensor, torch.Tensor]:
    batch = default_collate(data)
    if len(batch) != 2:
         raise RuntimeError(f"`DetectionCollateFN` only supports Datasets that return a tuple (image, target), but got a tuple of len={len(batch)}")
    ims, targets = batch[0:2]
    return self._format_images(images_batch), self._format_targets(labels_batch)
```
What do you think ?

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -768,12 +759,8 @@ def __call__(self, data) -> Tuple[torch.Tensor, torch.Tensor, Dict[str, torch.Te
...
             data = self.random_resize(data)
 
-        batch = default_collate(data)
-        ims, targets, crowd_targets = batch
-        if ims.shape[3] == 3:
-            ims = torch.moveaxis(ims, -1, 1).float()
-
-        return ims, self._format_targets(targets), {"crowd_targets": self._format_targets(crowd_targets)}
+        images_batch, labels_batch, crowd_labels_batch = list(zip(*data))
+        return self._format_images(images_batch), self._format_targets(labels_batch), {"crowd_targets": self._format_targets(crowd_labels_batch)}
```
Louis-Dupont, line 763: Same idea as above, but reversed: It would be great to raise an error if `len(batch) != 3`
I know we did not in the past, but this was a mistake, the code would just break without any explanation of why it doesnt work.

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -701,12 +712,8 @@ def __call__(self, data) -> Tuple[torch.Tensor, torch.Tensor]:
...
             data = self.random_resize(data)
 
-        batch = default_collate(data)
-        ims, targets = batch
-        targets = self._format_targets(targets)
-        ims = torch.moveaxis(ims, -1, 1).float()
-
-        return ims, targets
+        images_batch, labels_batch = list(zip(*data))
+        return self._format_images(images_batch), self._format_targets(labels_batch)
```
Louis-Dupont, line 716: Maybe we can call super() not not duplicate this part

Louis-Dupont (commented):
Looks great, thanks for your contribution! I left a few comments, let me know your thoughts

src/super_gradients/training/utils/detection_utils.py
yurkovak: I would argue that crashing is a desired behavior for such usage. Why would we want a collate that silently ignores items from a loader? This is by the way what happened in CI of 41674fb, I fixed collates in the test

src/super_gradients/training/utils/detection_utils.py
yurkovak: The code will break with "too many/not enough values to unpack", do you mean a more verbose and explanatory error?

src/super_gradients/training/utils/detection_utils.py
yurkovak: Good point, thanks! 91b0352

src/super_gradients/training/utils/detection_utils.py
Louis-Dupont: Crashing due to `ims, targets = batch[0:2]` would not be informative for people who don't know SG, this is why I propose to make it explicit:
``` python
    if len(batch) != 2:
         raise RuntimeError(f"`DetectionCollateFN` only supports Datasets that return a tuple (image, target), but got a tuple of len={len(batch)}")
```

src/super_gradients/training/utils/detection_utils.py
Louis-Dupont: Yeah, not for me but for others who would be lost with this error.

src/super_gradients/training/utils/detection_utils.py
BloodAxe: ++ for more user friendly error messages.
Yes, that's classic - change one place, and get +10 more improvement requests.

yurkovak: `if len(batch) != 2` is wrong, it's a batch length, not a sample length. Added an expcetion not to drag the message over all collates, see 452fb19
Using `try except` for [“zero-cost” exceptions](https://docs.python.org/3.11/whatsnew/3.11.html#misc), don't want any extra work on each iteration

src/super_gradients/training/utils/detection_utils.py
yurkovak: See the answer above

src/super_gradients/training/utils/detection_utils.py
yurkovak: @BloodAxe could you please follow-up on this?

yurkovak:
Something's wrong with CI, I think it's stuck in master and so commits and master merge into this branch don't appear in the pull request. A linked the relevant commit, please take a look anyway

Louis-Dupont (approved):
LGTM

PR #1111, opened 2023-06-01
Fixed resuming training with DDP when average_best_models=true

shaydeci:
- In line 52 we save the state dict even if it was just read. Saving is just for the first time we create the state_dict of the snapshots, so the saving is moved inside the else statement when we check if we resume or not.


Louis-Dupont (approved):
LGTM

PR #1114, opened 2023-06-01
Actualize pose estimatino zoo.

BloodAxe:
Remove old models that are not supported and we don't provide recipe & weights for them


shaydeci (dismissed):
LGTM

ofrimasad (approved):
LGTM

PR #1121, opened 2023-06-04
Bug/sg 896 add deprecation for previous breaking changes

shaydeci:
Added deprecations, and backward compatibility to previous breaking changes.

In this PR:
- InfiniteSampler has been removed, added deprecation using equivalent DistributedSampler.
- Broken imports
`from super_gradients.training.utils.optimizers.all_optimizers import OPTIMIZERS`
`from super_gradients.training.models import BasicBlock, Bottleneck, HpmStruct`

- DDRNetCustom: the __init__ expected a parameter “aux_head” from “arch_params.aux_head”. In the new version of SG, this parameter is: “use_aux_heads=arch_params.use_aux_heads” - added deprecation.
- BasicDDRBackBone . In old SG version, this block didn’t expect layer3_repeats . Now SG expects it. Also, ddrnet model now assumes that “arch_params” has layer3_repeats parameter. Fixed by setting default=1.

Verified backward compatibility handled outside this PR:

Some were already handled in other PRs, this is the complete list + where they were handled:
- In `ema_params`, they had: `exp_activation: True`. Changed to `decay_type: exp` handled by @BloodAxe in 69a82bc4b82f0676210065606181de907affd5e9
- The architecture `custom_stdc` name was changed to `stdc_custom`. Handled by @ofrimasad in 190283ebcbf7a8a0feeb028d853c8492c0d3db85
- Broken import: `from super_gradients.training.models import make_divisible`, handled by @BloodAxe in 6b5785d9bb513c0bf0c126a63cc0eb2bf457b1a2
- Had: `trainer = Trainer(experiment_name=cfg.experiment_name, multi_gpu=cfg.multi_gpu)`. Changed to use `setup_device(()` instead - this is the only case where we throw an error, to avoid setting devices inside the Trainer. f50d438c35661b7f2ffea3130ae3b97e19ccc31a
- Models iinheriting from `CustomizableDetector` Used to call `super().__init__` only with `arch_params`. Now its changed to accept each parameter separately:
```
super().__init__(
    backbone=arch_params.backbone,
    heads=arch_params.heads,
    neck=arch_params.neck,
    num_classes=arch_params.num_classes,
    bn_eps=arch_params.bn_eps,
    bn_momentum=arch_params.bn_momentum,
    inplace_act=arch_params.inplace_act
)

```
handled by @Louis-Dupont  in a3618e87b105808729576408fbcb878441696bc7

### DIFF
--- a/src/super_gradients/training/models/__init__.py
+++ b/src/super_gradients/training/models/__init__.py
@@ -35,7 +35,7 @@ from super_gradients.training.models.classification_models.preact_resnet import 
     PreActResNet152,
 )
 from super_gradients.training.models.classification_models.resnet import (
-    Bottleneck,
+    Bottleneck as NewBottleneck,
     BasicResNetBlock,
     ResNet,
     ResNet18,
@@ -164,6 +164,16 @@ BasicBlock = make_deprecated(
     "[-] from super_gradients.training.models import BasicBlock\n"
     "[+] from super_gradients.training.models import BasicResNetBlock\n",
 )
+
+Bottleneck = make_deprecated(
+    func=NewBottleneck,
+    reason="You're importing `Bottleneck` class from `super_gradients.training.models`. This is deprecated since SuperGradients 3.1.0.\n"
+    "This block was renamed to BasicResNetBlock for better clarity.\n"
+    "Please update your code to import it as follows:\n"
+    "[-] from super_gradients.training.models import Bottleneck\n"
+    "[+] from super_gradients.training.models.classification_models.resnet import Bottleneck\n",
+)
+
 HpmStruct = make_deprecated(
     func=CurrVersionHpmStruct,
     reason="You're importing `HpmStruct` class from `super_gradients.training.models`. This is deprecated since SuperGradients 3.1.0.\n"
--- a/src/super_gradients/training/models/__init__.py
+++ b/src/super_gradients/training/models/__init__.py
@@ -35,6 +35,7 @@ from super_gradients.training.models.classification_models.preact_resnet import 
     PreActResNet152,
 )
 from super_gradients.training.models.classification_models.resnet import (
+    Bottleneck,
     BasicResNetBlock,
     ResNet,
     ResNet18,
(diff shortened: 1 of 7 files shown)


src/super_gradients/training/models/__init__.py
```diff
@@ -163,8 +164,18 @@ def inner(*args, **kwargs):
...
+    reason="You're importing `HpmStruct` class from `super_gradients.training.models`. This is deprecated since SuperGradients 3.1.0.\n"
+    "Please update your code to import it as follows:\n"
+    "[-] from super_gradients.training.models import HpmStruct\n"
+    "[+] from super_gradients.training.utils import HpmStruct\n",
+)
+
 
 __all__ = [
+    "HpmStruct",
+    "Bottleneck",
```
ofrimasad, line 178: where is the deprecation for this?

src/super_gradients/training/utils/optimizers/all_optimizers.py
```diff
@@ -0,0 +1,11 @@
+import warnings
+
+from super_gradients.common.registry.registry import OPTIMIZERS as CURRENT_VERSION_OPTIMIZERS
+
+OPTIMIZERS = CURRENT_VERSION_OPTIMIZERS
+
+warnings.warn(
```
ofrimasad, line 7: why are we not using the same concept of `make_deprecated`?

ofrimasad (commented):
two minor comments

src/super_gradients/training/utils/optimizers/all_optimizers.py
shaydeci: Because the entirely_optimizers.py file was removed, it made more sense to me to warn any user that would try and access this file.

src/super_gradients/training/models/__init__.py
shaydeci: slipped out, fixed in 1ae6315af6e40de9205e517096c7200986e57ec5

ofrimasad (approved):
LGTM

PR #1126, opened 2023-06-05
Feature/sg 000 yolo nas pose

BloodAxe:
Depends on https://github.com/Deci-AI/super-gradients/pull/1142 and have changes from that PR already in to test all things are working as intended. To be merged (and probably reviewed) after merge of predict()


src/super_gradients/training/models/pose_estimation_models/yolo_nas_pose.py
```diff
@@ -0,0 +1,239 @@
...
+        self.stem = ConvBNAct(sum(in_channels), inter_channels, kernel_size=1, stride=1, padding=0, bias=False, activation_type=activation_type)
+        self.heatmap = self.build_heatmap_path(inter_channels, activation_type, num_blocks, self.num_classes_with_center, upsample_factor)
+
+        self.offset_transition = ConvBNAct(
+            inter_channels, channels_per_kpt * num_classes, kernel_size=3, stride=1, padding=1, bias=False, activation_type=activation_type
+        )
+
+        offset_modules = []
+        for _ in range(num_classes):
+            offset_modules.append(self.build_offset_path(inter_channels, activation_type, num_blocks, channels_per_kpt, upsample_factor))
```
ofrimasad, line 52: If I am not mistaken, this can be implemented more efficiently by using 
Grouped Convolution (where groups=num_classes). It will save slice and concat but also can be optimized by the GPU much better. 

image

src/super_gradients/training/models/pose_estimation_models/yolo_nas_pose.py
```diff
@@ -0,0 +1,239 @@
...
+        blocks += [("final", nn.Conv2d(channels_per_kpt, 2 * (2**upsample_factor), kernel_size=1)), ("shuffle", nn.PixelShuffle(upsample_factor))]
+        return nn.Sequential(collections.OrderedDict(blocks))
+
+    @property
+    def out_channels(self):
+        return (self.num_classes_with_center), (2 * self.num_classes)
+
+
+class YoloNASPose(CustomizableDetector):
+    def __init__(self, arch_params: Union[HpmStruct, DictConfig]):
```
ofrimasad, line 105: lets drop the arch_params here please.
we said we will not create new models with that concept.
This init should accept the actual parameters by name (only the variants below should accept arch_params)

ofrimasad (commented):
Very nice work!
I added some comments inline.

src/super_gradients/training/metrics/pose_estimation_metrics.py
```diff
@@ -103,7 +103,14 @@ def __init__(
         self.greater_component_is_better = dict((k, True) for k in self.stats_names)
 
         if oks_sigmas is None:
-            oks_sigmas = np.array([0.26, 0.25, 0.25, 0.35, 0.35, 0.79, 0.79, 0.72, 0.72, 0.62, 0.62, 1.07, 1.07, 0.87, 0.87, 0.89, 0.89]) / 10.0
+            if num_joints == 17:
+                oks_sigmas = np.array([0.26, 0.25, 0.25, 0.35, 0.35, 0.79, 0.79, 0.72, 0.72, 0.62, 0.62, 1.07, 1.07, 0.87, 0.87, 0.89, 0.89]) / 10.0
```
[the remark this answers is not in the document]
BloodAxe: These are default values for COCO2017 PE

src/super_gradients/training/processing/processing.py
```diff
@@ -455,6 +455,87 @@ def default_dekr_coco_processing_params() -> dict:
...
+            KeypointsLongestMaxSizeRescale(output_shape=(640, 640)),
+            KeypointsBottomRightPadding(output_shape=(640, 640), pad_value=127),
+            StandardizeImage(max_value=255.0),
+            NormalizeImage(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
+            ImagePermute(permutation=(2, 0, 1)),
+        ]
+    )
+
+    edge_links = [
+        [0, 1],
```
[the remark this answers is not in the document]
BloodAxe: Yes, they are defined in `coco_pose_dataset_params`, however the goal of having them defined here is to provide a default params if they are missing in the checkpoint. Which is the case for your DEKR & YOLO-NAS checkpoints, since they were trained before we introduced Predict API for PE.

src/super_gradients/training/models/pose_estimation_models/yolo_nas_pose.py
BloodAxe: This looks like a good idea from inference time perspective, however we need to empiricaly test for this change affects mAP. In the [paper](https://arxiv.org/pdf/2104.02300.pdf) authors discuss this briefly in section `3.4. Discussions`. Sadly they do not provide numbers for group-conv approach.

src/super_gradients/recipes/coco2017_pose_yolo_pose_s.yaml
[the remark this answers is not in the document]
BloodAxe: Fixed in 8de2595bf07153c201031fe5e9f6b6e671552c18

src/super_gradients/training/models/pose_estimation_models/yolo_nas_pose.py
BloodAxe: Just to leave some traces for future generations - we did try the group conv approach. 
Yes, ONNX looked much better, but in fact we gained zero improvement of the throughput.

PR #1128, opened 2023-06-05
Bug/sg 897 support for min samples with ddp

shaydeci:
Added support for min_samples in DDP with RepeatAugSampler.

I also added tests for min_samples functionality, bot for DDP and single node.


src/super_gradients/training/dataloaders/dataloaders.py
```diff
@@ -150,6 +152,18 @@ def _process_sampler_params(dataloader_params, dataset, default_dataloader_param
...
+            if len(dataset) < min_samples:
+                world_size = dist.get_world_size()
+                num_repeats = min_samples / len(dataset)
+                selected_ratio = world_size / num_repeats
+
+                # WE SET selected_ratio = world_size / num_repeats SO THAT IN RepeatAugSampler THE NUMBER OF SAMPLES
+                # PER EPOCH PER RANK WILL BE DETERMINED BY
+                # int(math.ceil(len(self.dataset) / selected_ratio)) =  min_samples / world_size
+
+                dataloader_params["sampler"] = RepeatAugSampler(dataset=dataset, num_repeats=num_repeats, selected_round=0, selected_ratio=selected_ratio)
```
BloodAxe, line 166: Can you please explain why we use `RepeatAugSampler` and not `torch.utils.data.RandomSampler(dataset, num_samples=min_samples//world_size)`?
a) Use of RepeatAugSampler looks like an overkill to me
b) I don't trust implementation of RepeatAugSampler

shaydeci: Using Random sampler with DDP is wrong as it results in duplicating the dataset across the different processes instead of dividing it between them.

So what we want is a DistributedSampler with the functionality of truncating the dataset, and RepeatAugSampler is exactly that.

BloodAxe (approved):
LGTM

PR #1138, opened 2023-06-06
Allow checkpoint_path to be an url (Required for transfer learning)

BloodAxe:
- Allow checkpoint_path to be an url (Required for transfer learning)
- Added synchronization primitives to prevent race conditions when downloading weights

One can now load checkpoints like so:
```
models.get(Models.YOLO_NAS_S, num_classes=80, checkpoint_path="https://sghub.deci.ai/models/yolo_nas_s_coco.pth")
```

Why we want this to have: For training YOLO-NAS-POSE we want to use YOLO-NAS-OD pretrained weights like so:
```
checkpoint_params:
  checkpoint_path: "https://sghub.deci.ai/models/yolo_nas_s_coco.pth"
```

This PR enables us to do this.

### DIFF
--- a/src/super_gradients/training/utils/checkpoint_utils.py
+++ b/src/super_gradients/training/utils/checkpoint_utils.py
@@ -10,6 +10,7 @@ from super_gradients.common.decorators.explicit_params_validator import explicit
 from super_gradients.module_interfaces import HasPredict
 from super_gradients.training.pretrained_models import MODEL_URLS
 from super_gradients.common.data_types import StrictLoad
+from super_gradients.training.utils.distributed_training_utils import get_local_rank, wait_for_the_master
 
 from torch import nn, Tensor
 from typing import Union, Mapping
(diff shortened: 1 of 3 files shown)


src/super_gradients/training/utils/checkpoint_utils.py
```diff
@@ -206,10 +217,6 @@ def load_checkpoint_to_model(
     if isinstance(strict, str):
         strict = StrictLoad(strict)
 
-    if ckpt_local_path is None or not os.path.exists(ckpt_local_path):
-        error_msg = "Error - loading Model Checkpoint: Path {} does not exist".format(ckpt_local_path)
-        raise RuntimeError(error_msg)
-
```
Louis-Dupont, line 212: Why remove this ? Because we check it before already?

Louis-Dupont (approved):
LGTM, just small question below

src/super_gradients/training/utils/checkpoint_utils.py
BloodAxe: We have exactly the same check inside the method itself, yes.
But since we are adding URL support, this one triggers as false-positive

Louis-Dupont: Yeah that's right

PR #1140, opened 2023-06-06
Bug/sg 000 merge failure for datasetparams

shaydeci:
Recent changes in https://github.com/Deci-AI/super-gradients/pull/1092 caused pose estimation recipe tests to crash https://app.circleci.com/pipelines/github/Deci-AI/super-gradients/7748/workflows/5d210255-4bed-4736-b4d4-b693d6bc91e1/jobs/23382

This is becaue there are overrides for dataset params from the command line, causing the merge to fail.

Before calling merge we neeed to set the two configs's struct to false, other wise if one includes an attribute which the other one does not have - the merge will fail.


BloodAxe (dismissed):
I'm wondering into how many unexpected things we will run with this config merging, but..
LGTM )

Louis-Dupont (dismissed):
LGTM

ofrimasad (dismissed):
LGTM

shaydeci:
> Pls add test to tests/unit_tests/detection_dataset_test.py ^_^

Added a simple test that uses dataloaders.get with dataset_params that indclude an entry which is not present in the yaml file (this is what caused the crash).

BloodAxe (approved):
LGTM

Issue #1161, opened 2023-06-12
DetectionVerticalFlip


LukeAI:
### 🚀 Feature Request

In https://github.com/Deci-AI/super-gradients/blob/a43cfcd70072c7be1231f9183b1a717c136ff657/src/super_gradients/training/transforms/transforms.py#LL815C1-L815C1

There is `DetectionHorizontalFlip` but no `DetectionVerticalFlip` which would be a useful augmentation to have for my dataset, it would be trivial to implement and I'd be happy to open a PR if it would be accepted?

NatanBagrov:
Hello, PRs are welcome 🥇

LukeAI:
@NatanBagrov  have implemented https://github.com/Deci-AI/super-gradients/issues/1161

harpreetsahota204:
Hi @LukeAI -

Thanks for opening an issue for SG. I'm gathering some feedback on SuperGradients and YOLO-NAS.

Would you be down for a quick call to chat about your experience?

If a call doesn't work for you, no worries. I've got a short survey you could fill out: https://bit.ly/sgyn-feedback.

I know you’re super busy, but your input will help us shape the direction of SuperGradients and make it as useful as possible for you.

I appreciate your time and feedback. Let me know what works for you.

Cheers,

Harpreet

LukeAI:
yeah a call is fine

harpreetsahota204:
@LukeAI -  Great! Feel free to find sometime here:https://calendly.com/harpreet-sahota/30min

LukeAI:
Hi Harpreet,

Sorry to let you down short notice but I'm not going to be able to make the 11.30 call, something has come up. I can reschedule.

Luke
________________________________

PR #1165, opened 2023-06-12
Add Transforms.DetectionVerticalFlip

LukeAI:
There is currently a `DetectionHorizontalFlip` but no `DetectionVerticalFlip`. This PR adds DetectionVerticalFlip which is a common, useful augmentation, (which I want to use for my application with yoloNAS)

Resolves: https://github.com/Deci-AI/super-gradients/issues/1161
The diff looks a bit weird, but it's legit.
https://github.com/Deci-AI/super-gradients/blob/a43cfcd70072c7be1231f9183b1a717c136ff657/src/super_gradients/training/transforms/transforms.py#LL815C1-L815C1


src/super_gradients/training/transforms/transforms.py
```diff
@@ -829,7 +829,31 @@ def __call__(self, sample):
...
             targets = np.zeros((0, 5), dtype=np.float32)
         boxes = targets[:, :4]
-        image, boxes = _mirror(image, boxes, self.prob)
+        image, boxes = _horizontal_mirror(image, boxes, self.prob)
+        targets[:, :4] = boxes
+        sample["target"] = targets
+        sample["image"] = image
+        return sample
+
+@register_transform(Transforms.DetectionVerticalFlip)
```
BloodAxe, line 838: Is Transforms.DetectionVerticalFlip exists? I don't see it in a changeset.

src/super_gradients/training/transforms/transforms.py
```diff
@@ -829,7 +829,31 @@ def __call__(self, sample):
         if len(targets) == 0:
             targets = np.zeros((0, 5), dtype=np.float32)
         boxes = targets[:, :4]
-        image, boxes = _mirror(image, boxes, self.prob)
+        image, boxes = _horizontal_mirror(image, boxes, self.prob)
```
BloodAxe, line 832: I realize that this augmentation does not respect the crowd targets.
Can you please transform `crowd_target` as well? See DetectionRescale for an example.
For this you may want to remove random sampling from _horizontal_mirror implementation and do the if random() < self.prob inside the __call__ method.

BloodAxe (changes requested):
Thanks for your PR. I think this would be a good addition. However there are few things to address, see inline comments.

src/super_gradients/training/transforms/transforms.py
LukeAI: Hi @BloodAxe , thanks for feedback.
I've made two changes to `_mirror()`:
(1) The name -> `_horizontal_mirror`
(2) The expression, but not the functionality (I believe!) of the transform of the X coordinates:
`flipped_boxes[:, 0::2] = width - boxes[:, 2::-2]`
-> `flipped_boxes[:, [0,2]] = width - boxes[:, [2,0]]`

I made change (2) because I think directly, explicitly stating the columns is more readable than implicitly with the reverse periodic slice.

Is your comment "this augmentation does not respect the crowd targets" related to change (2)? Is it because crowd_target is expressed as a list of coordinates like [ [xyxyxyxyxy...] [xyxyxyxy...] ?

BloodAxe: > flipped_boxes[:, [0,2]] = width - boxes[:, [2,0]]
Looks good!

BloodAxe: > Is your comment "this augmentation does not respect the crowd targets" related to change (2)? Is it because crowd_target is expressed as a list of coordinates like [ [xyxyxyxyxy...], [xyxyxyxy...] ?

Essentially what you want to do is the following:
```
image, targets, crowd_targets = sample["image"], sample["target"], sample.get("crowd_targets")
if self.prob < randon.random():
    image = fliplr_image(image)
    targets = fliplr_boxes(targets)
    if crowd_targets  is not None:
      crowd_targets = fliplr_boxes(targets)
      
    ...
```

Here you may notice I introduced new functions: `fliplr_image`, `fliplr_boxes` and moved probability check to the class method itself.

LukeAI: ok, so this is not related to the changes I made? It's adding additional unrelated functionality? That's fine, just trying to understand

src/super_gradients/training/transforms/transforms.py
```diff
@@ -1277,21 +1318,40 @@ def _filter_box_candidates(box1, box2, wh_thr=2, ar_thr=20, area_thr=0.1):
     return (w2 > wh_thr) & (h2 > wh_thr) & (w2 * h2 / (w1 * h1 + 1e-16) > area_thr) & (ar < ar_thr)  # candidates
 
 
-def _mirror(image, boxes, prob=0.5):
+def _flip_horizontal_image(image: np.array):
```
BloodAxe, line 1321: A correct type hint for numpy array is `np.ndarray`, not `np.array`. `np.array` is a method.

src/super_gradients/training/transforms/transforms.py
```diff
@@ -1277,21 +1318,40 @@ def _filter_box_candidates(box1, box2, wh_thr=2, ar_thr=20, area_thr=0.1):
...
+    return image[:, ::-1]
+
 
+def _flip_horizontal_boxes(boxes: np.array, img_width: int):
+    """
+    Horizontally flips bboxes
+    :param boxes: (np.array) bboxes to be flipped. (xyxy format)
+    :return: flipped_boxes
+    """
+    return img_width - boxes[:, [2,1,0,3]]
```
BloodAxe, line 1336: Please add a unit test for this method since it looks wrong to me.
You are selecting `X1,Y1,X2,Y2` in the right order via indexing `[2,1,0,3]` into `X2, Y1, X1, Y2`. However you then subtract from img_width all four components. So final boxes are:
`img_width - X2, img_width - Y1, img_width - X1, img_width - Y2`
Y coordinate should remain unchanged during flip operation.

BloodAxe:
Once you have the tests in place and other commends addressed please set-up signed commits as described here: https://github.com/Deci-AI/super-gradients/blob/master/CONTRIBUTING.md
We require all PRs to be signed. 

The easiest way to move your work is to branch from master again, checkout new branch and the do:
`git merge --no-commit --no-ff <your feature branch>`

LukeAI:
Please could you point me in the right direction for writing tests? Are there any examples for the other transforms?

________________________________

BloodAxe:
You can take this test case as an example and add more tests to validate your implementation of _flip_horizontal_boxes and other functions you introduced.

https://github.com/Deci-AI/super-gradients/blob/2cb3a7c2709bcbe2f01856b11c5b5a69fcdbd13f/tests/unit_tests/transforms_test.py#L172

Start with simple examples.
Say you have image for 16x24 and you have bbox of (1, 2, 3, 4) (x1, y1, x2, y2)
It's quite easy to count what coordinates of the bbox should be after H & V flip. 
So the unit test should check that.

LukeAI:
@BloodAxe please could you advise me how to actually run the tests in the local cloned repo 🙃
If I `cd src` so that I'm in the same dir, as `super-gradients` and then try to run the test file like:
`python -m unittest ../tests/unit_tests/transforms_test.py` I get:
```
  File "/home/luke/.pyenv/versions/3.8.17/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
ValueError: Empty module name
```

If I run `python -m unittest discover ../tests/unit_tests` it appears to run something.
```
[2023-06-17 18:57:06] INFO - crash_tips_setup.py - Crash tips is enabled. You can set your environment variable to CRASH_HANDLER=FALSE to disable it
[2023-06-17 18:57:08] WARNING - __init__.py - Failed to import pytorch_quantization
[2023-06-17 18:57:08] WARNING - calibrator.py - Failed to import pytorch_quantization
[2023-06-17 18:57:08] WARNING - export.py - Failed to import pytorch_quantization
[2023-06-17 18:57:08] WARNING - selective_quantization_utils.py - Failed to import pytorch_quantization
...[2023-06-17 18:57:08] WARNING - utils.py - The CustomizableDetector.__init__ received `arch_params` argument which is deprecated and will be removed in next versions. Please change the signature of the __init__ method to take explicit list arguments instead: CustomizableDetector.__init__(backbone, heads, neck, bn_eps, bn_momentum, inplace_act, in_channels)
.[2023-06-17 18:57:09] WARNING - utils.py - The CustomizableDetector.__init__ received `arch_params` argument which is deprecated and will be removed in next versions. Please change the signature of the __init__ method to take explicit list arguments instead: CustomizableDetector.__init__(backbone, heads, neck, num_classes, bn_eps, bn_momentum, inplace_act, in_channels)
.[2023-06-17 18:57:09] WARNING - utils.py - The DummyModelV2.__init__ received `arch_params` argument which is deprecated and will be removed in next versions. Please change the signature of the __init__ method to take explicit list arguments instead: DummyModelV2.__init__(backbone, head, neck)
......[2023-06-17 18:57:09] INFO - dataloaders.py - Using min_samples=80
.[2023-06-17 18:57:11] WARNING - model_factory.py - Passing num_classes through arch_params is deprecated and will be removed in the next version. Pass num_classes explicitly to models.get
.
----------------------------------------------------------------------
Ran 13 tests in 2.942s

OK
```
but I started off by adding a unit test designed to fail, and that isn't registering, so I don't think it ran anything in `unit_tests/transforms_test.py`

I've tried a few other incantations but I can't seem to work out how to run the contents of transforms_test.py in a way that uses the code in the repo. some guidance would be greatly appreciated.

BloodAxe:
```
# Install SG in development mode (From repo root)
pip install --no-deps -e .

# Run tests
python -m unittest tests/unit_tests/transforms_test.py
``` 

You can always use IDE's builint functionality to run unit tests from it (Pycharm for sure and  I assume vscode support automatic tests discrovery there days as well)

LukeAI:
How exactly are the xyxy bounding boxes defined? eg. if you had a bounding box in the top left corner with a width and height of 10, would that bounding box be `((0,0), (10,10))` - ie. the bounding box is considered to pass through the centre of each pixel?

If so, then on an image of 100x100 (ie. 0-99, 0-99) a horizontal flip of `((0,0), (10,10))` would give us `((89, 0), (99, 10))`

The calculation for horizontal flip in super-gradients is given as `flipped_boxes[:, 0::2] = width - boxes[:, 2::-2]`
https://github.com/Deci-AI/super-gradients/blob/a43cfcd70072c7be1231f9183b1a717c136ff657/src/super_gradients/training/transforms/transforms.py#LL1293C9-L1293C57

which would give in this case: `((90, 0), (100, 0))` which extends half a pixel beyond the edge of the image.

So either the `xyxy` interpretation is wrong or the transform is wrong.

Maybe the bounding box is considered to border the top left corner of each `xy` pair? In which case, the top left hand corner 10x10 bounding box would still be ((0,0), (10,10)) and the original transform would be correct.

LukeAI:
Ok, have added unit tests for both DetectionHorizontal and DetectionVertical Transforms. If that all seems fine I'll do the signed commit thing.

LukeAI:
@BloodAxe does that all seem ok? shall I go ahead and do the signed PR thing?

src/super_gradients/training/transforms/transforms.py
```diff
@@ -655,8 +655,10 @@ def __call__(self, sample: dict) -> dict:
             cp_sample = sample["additional_samples"][0]
             img, cp_labels = cp_sample["image"], cp_sample["target"]
             cp_boxes = cp_labels[:, :4]
-
-            img, cp_boxes = _mirror(img, cp_boxes, self.flip_prob)
+            if self.prob < random.random():
```
BloodAxe, line 658: The condition should be `if random.random() < self.prob`. The order matters:
Imagine your have self.prob = 1, which can be translated to "Always apply" that transformation.
A `random()` returns a value in range `0.0 <= X < 1.0`.
So your condition `if self.prob < random.random():` would never be a True: `if 1 < random()` which is by definition always less than one.

src/super_gradients/training/transforms/transforms.py
```diff
@@ -826,13 +828,52 @@ def __init__(self, prob: float, max_targets: Optional[int] = None):
...
-        image, boxes = _mirror(image, boxes, self.prob)
-        targets[:, :4] = boxes
+        if random.random() < self.prob:
+            image = _flip_horizontal_image(image)
+            boxes = targets[:, :4]
+            _, width, _ = image.shape
+            boxes = _flip_horizontal_boxes(boxes, width)
+            targets[:, :4] = boxes
+            if crowd_targets is not None:
+                crowd_targets = _flip_horizontal_boxes(crowd_targets)
```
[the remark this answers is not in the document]
BloodAxe: Probably a good idea to add crowd targets to your unit test as well

src/super_gradients/training/transforms/transforms.py
```diff
@@ -826,13 +828,52 @@ def __init__(self, prob: float, max_targets: Optional[int] = None):
...
+        if len(targets) == 0:
+            targets = np.zeros((0, 5), dtype=np.float32)
+        if random.random() < self.prob:
+            image = _flip_vertical_image(image)
+            boxes = targets[:, :4]
+            height, _, _ = image.shape
+            boxes = _flip_vertical_boxes(boxes, height)
+            targets[:, :4] = boxes
+            if crowd_targets is not None:
+                crowd_targets = _flip_vertical_boxes(crowd_targets)
```
BloodAxe, line 873: Same here, missing second argument

src/super_gradients/training/transforms/transforms.py
```diff
@@ -826,13 +828,52 @@ def __init__(self, prob: float, max_targets: Optional[int] = None):
...
             targets = np.zeros((0, 5), dtype=np.float32)
-        boxes = targets[:, :4]
-        image, boxes = _mirror(image, boxes, self.prob)
-        targets[:, :4] = boxes
+        if random.random() < self.prob:
+            image = _flip_horizontal_image(image)
+            boxes = targets[:, :4]
+            _, width, _ = image.shape
+            boxes = _flip_horizontal_boxes(boxes, width)
+            targets[:, :4] = boxes
```
BloodAxe, line 839: You can actually simplify
this
```
boxes = targets[:, :4]
_, width, _ = image.shape
boxes = _flip_horizontal_boxes(boxes, width)
targets[:, :4] = boxes
```

Into this:
```
_, width, _ = image.shape
targets[:, :4] = _flip_horizontal_boxes(targets[:, :4], width)
```

Which in my opinion quite easier to understand & more compact

BloodAxe:
@LukeAI not quite ^_^
But we are getting closer, that's for sure :)
Left a few comments addressing few remaining issues.

BloodAxe:
This PR looks really good to me! Feel free to proceed with signed commits :) Well done!

LukeAI:
Ok, have opened new PR with signed commit https://github.com/Deci-AI/super-gradients/pull/1234

LukeAI:
and again with squashed commits
https://github.com/Deci-AI/super-gradients/pull/1237

Issue #1136, opened 2023-06-06
Add visualization of predicted bounding boxes on Weights & Biases


soumik12345:
### 🚀 Feature Request

Add feature to **optionally** perform inference with the model while training on a small subset of the validation data at a fixed interval and log them on Weights & Biases. This helps us gain further insight into how the model performs on specific examples from the validation set.

This feature is part of similar Weights & Biases integrations like YOLOv5, YOLOv8, Keras, etc.

- [This report](https://wandb.ai/av-team/bdd100k-perception/reports/BoundingBoxDebugger-23-06-01-11-16-31---Vmlldzo0NTI4MTE2) shows a media panel with the bounding boxes logged at the end of every training epoch for [YOLOv5](https://github.com/ultralytics/yolov5). This slider lets us visualize the predicted bounding boxes logged after every epoch.
- [This report](https://wandb.ai/parambharat/YOLOv8/reports/Sample-Detections--Vmlldzo0NTY5MDk3) shows the same as an interactive [`wandb.Table`](https://docs.wandb.ai/guides/data-vis) for [YOLOv8](https://github.com/ultralytics/ultralytics).
- [This report](https://wandb.ai/ayush-thakur/mmdetection/reports/How-To-Use-Weights-Biases-With-MMDetection--VmlldzoyMTM0MDE2#visualize-dataset-and-model-prediction) shows the same for [mmdetection](https://github.com/open-mmlab/mmdetection) as an interactive [`wandb.Table`](https://docs.wandb.ai/guides/data-vis).

### Proposed Solution

Possible ways to implement:

- Implement as a [`Callback`](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/training/utils/callbacks/base_callbacks.py#L94) by overriding the [`on_validation_batch_end`](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/training/utils/callbacks/base_callbacks.py#L289).
- Implement as part of the [`WandBSGLogger`](https://github.com/Deci-AI/super-gradients/blob/master/src/super_gradients/common/sg_loggers/wandb_sg_logger.py#L30).

soumik12345:
@ofrimasad @BloodAxe 
I would love to discuss this!

PR #1167, opened 2023-06-13
Add utilities to plot datasets to Weights & Biases + Add callback to log validation predictions to Weights & Biases

soumik12345:
# Changes made in this PR

Fixes https://github.com/Deci-AI/super-gradients/issues/1136

👇 Here's a sample training code that works with the changes made in this PR
[](https://colab.research.google.com/drive/14kIWc9EYdqGmDkA_xrY7pmheSqsndmKq?usp=sharing)

👇 Here's a Weights & Biases run that demonstrates the changes made in this PR:
https://wandb.ai/geekyrakshit/yolo-nas-integration-2/runs/3lcpxzti

## Plot datasets to Weights & Biases

This PR adds a function `plot_detection_dataset_on_wandb()` which can be used to plot and visualize a `super_gradients.training.datasets.detection_datasets.DetectionDataset` to Weights & Biases as a [`wandb.Table`](https://docs.wandb.ai/guides/data-vis) which can be used for interactive exploratory analysis of respective datasets.

Here's a sample code:

```python
train_data = coco_detection_yolo_format_train(
    dataset_params={
        'data_dir': dataset_params['data_dir'],
        'images_dir': dataset_params['train_images_dir'],
        'labels_dir': dataset_params['train_labels_dir'],
        'classes': dataset_params['classes']
    },
    dataloader_params={
        'batch_size':16,
        'num_workers':2
    }
)
plot_detection_dataset_on_wandb(train_data.dataset, max_examples=20, dataset_name="Train-Dataset")
```

This would log the dataset to be visualized as a Table in a Weights & Biases dashboard. 

Here's a video of how to interact with the table UI on Weights & Biases 👇

https://github.com/Deci-AI/super-gradients/assets/19887676/b448e586-aefc-4d74-ba9d-5f6b5123b14e

## Log validation prediction as a table on Weights & Biases

This PR also includes a callback `WandBDetectionValidationPredictionLoggerCallback` that logs object detection predictions to a Weights & Biases Table with interactive bounding-box overlays during training on an epoch-wise basis.

Here's a sample code:

```python
train_params.update({
    "phase_callbacks": [
        WandBDetectionValidationPredictionLoggerCallback(class_names=labels),
    ]
})
trainer.train(model=net, training_params=train_params, train_loader=train_data, valid_loader=val_data)
```

This would log the validation predictions to be visualized as a Table in a Weights & Biases dashboard. 

Here's a video of how to interact with the table UI on Weights & Biases 👇

https://github.com/Deci-AI/super-gradients/assets/19887676/18fdc9ed-59c9-4195-a53b-8cf999649b31


soumik12345:
Hi @ofrimasad 
Would love your feedback on this PR.

kldarek:
Hi @ofrimasad & team, we'd really appreciate feedback on the PR, thanks so much! Darek from W&B

src/super_gradients/common/plugins/wandb/validation_logger.py
```diff
@@ -0,0 +1,77 @@
...
+
+        :param class_names:         A list of class names.
+        :param score_threshold:     Predictions confidence threshold. Predictions with score lower than score_threshold will not participate in Top-K & NMS
+        :param iou:                 IoU threshold for NMS step.
+        :param nms_top_k:           Number of predictions participating in NMS step
+        :param max_predictions:     maximum number of boxes to return after NMS step
+        """
+        super().__init__()
+        self.class_names = class_names
+        self.post_prediction_callback = PPYoloEPostPredictionCallback(
```
BloodAxe, line 32: Hard-coded post-prediction callback is not quite flexible solution and would work only for YOLO-NAS & PPYoloE models. Other models like YoloX, SSD and any other model would not work here. 
I wonder why don't you check `if isinstance(context.net, HasPredict)` to ensure whether model supports `predict` API and call it instead to obtain predictions? I beleive it should make the code much easier. One thing not to forget when calling predict from callback - pass `fuse_layers=False` explicitly to prevent irreversible runtime optimizations on the model.

BloodAxe:
Thanks for your PR, that is really nice addition. Pardon for delays with reviewing it as we were pretty busy recently preparing a new release.
We are here to help to make the process of accepting this PR as smooth as possible. 
So here is a check-list of things that I believe are missing/important to address, but which would allow other users to start using this callback easily:

- [ ] A callback class should either use predict API (All of our detection models support this) or get `post_prediction_callback` as an argument externally or you probably can do a `unwrap_model(context.net).get_post_prediction_callback() call to get it inside the callback method. I tend to lean towards the use of `post_prediction_callback` option for a number of reasons:
1) since predict method may not be up to the task, because the `context.inputs` is already normalized tensors. @Louis-Dupont correct me if I'm wrong - what would happen if we pass torch tensor to predict()? Would we double-apply normalization here?
2) `predict` call may be doing layer fusion to optimize the inference speed, however it may turn model non-trainable, which is not what we want during training. So `fuse_model=False` must be passed to predict()

- [ ] It looks like the callback is logging images on each batch in validation and accumulating generated images in RAM. I can easily see this is causing OOM errors for datasets like COCO. Is it intentional? If it mean to be used only for `validate_from_recipe` regime, please clarify this in the docstring description with the intended use case scenario of the callback.

soumik12345:
Hi @BloodAxe 
Made the requested changes and apologies for the delay 😅

Also a couple of questions...

1. Is it possible to visualize the dataset before applying augmentations?

2. In the validation table, when the data is being logged at the end of the epochs, the bottom half of the images seem to have some kind of padding. Do you recommend some post-processing in order to avoid this? (such as slicing the bottom half of the image before plotting)

image

BloodAxe:
> Is it possible to visualize the dataset before applying augmentations?

No

> In the validation table, when the data is being logged at the end of the epochs, the bottom half of the images seem to have some kind of padding. Do you recommend some post-processing in order to avoid this? (such as slicing the bottom half of the image before plotting)

This is due to padding when preparing samples in dataset. If your input images has aspect ratio 16:9 (example) but the dataset transforms output images in 640x64 (aspect 1:1) then you would get this issue. The right solution is to use the resolution during training that matches (more or less) the aspect ratio of the images in dataset. So I'm not sure whether there is any action is required.

BloodAxe (approved):
LGTM

PR #1172, opened 2023-06-14
Feature/sg 686 torch compile

BloodAxe:
Depends on https://github.com/Deci-AI/super-gradients/pull/1181/files


documentation/source/model_zoo.md
```diff
@@ -44,22 +44,22 @@ All the available models are listed in the column `Model name`.
...
-| SSD lite MobileNet v1 | ssd_mobilenet_v1      | COCO    | 320x320    | 24.3                     | **1.55ms**                    | **2.84ms**                          | **8.07ms**                               | **9.14ms**                                        |           **22.76ms**           |
+| Model                 | Model Name            | Dataset | Resolution | mAP<sup>val<br>0.5:0.95  | Latency (HW)*<sub>T4</sub>    | Latency (Production)**<sub>T4</sub> | Latency (HW)*<sub>Jetson Xavier NX</sub> | Latency (Production)**<sub>Jetson Xavier NX</sub> | Latency <sub>Cascade Lake</sub> |                              Torch Compile Supported |
+|-----------------------|-----------------------|---------|------------|--------------------------|-------------------------------|-------------------------------------|------------------------------------------|---------------------------------------------------|:-------------------------------:|-----------------------------------------------------:|
```
shaydeci, line 48: Torch Compile Support is a better title then Torch Compile Supported I believe.

documentation/source/model_zoo.md
```diff
@@ -44,22 +44,22 @@ All the available models are listed in the column `Model name`.
...
+| YOLO-NAS L            | yolo_nas_l            | COCO    | 640x640    | 52.22(FP16) 52.1(INT8)   | **7.87(FP16)** **4.78(INT8)** |                                     |                                          |                                                   |                                 |                                            Supported |
+| PP-YOLOE small        | ppyoloe_s             | COCO    | 640x640    | 42.52                    | **2.39ms**                    | **4.3ms**                           | **14.28ms**                              | **14.99ms**                                       |              **-**              |           TypeError: 'Tensor' object is not callable |
+| PP-YOLOE medium       | ppyoloe_m             | COCO    | 640x640    | 47.11                    | **5.16ms**                    | **7.05ms**                          | **32.71ms**                              | **33.46ms**                                       |              **-**              |           TypeError: 'Tensor' object is not callable |
```
shaydeci, line 53: Can we just put No/Yes or X/V instead of the error message ?

src/super_gradients/recipes/coco2017_yolo_nas_m.yaml
```diff
@@ -0,0 +1,43 @@
+# YoloNAS-S Detection training on COCO2017 Dataset:
```
shaydeci, line 1: Why do we need those in this PR ?

shaydeci (commented):
Great work!

Besides the comments inline I have some suggestions:
- I think its worth checking how much speedup the heavier classification models that we have recipes for gain from torch compile (since the bottleneck would actually be the model's forward pass). I would be very interested in seeing how much we would gain for our KD recipe for Resnet50 on Imagenet.
- I already discussed this with you - should we give a shot and try to torch.compiile the loss modules maybe? It wouild seem to me that for some recipes, especially in OD that the bottleneck might actually be there. We can add it for another time, yet it still is very intriguing. 
- I think we should have some info about what modules cannot be compiled inside the code (that is, in addition to the table you added to the model zoo). Then, if some user attempts to compile an unsupported model we can catch that in advance.

documentation/source/model_zoo.md
BloodAxe: Yep! Fixed that. I left messages mostly for myself (:

src/super_gradients/recipes/coco2017_yolo_nas_m.yaml
BloodAxe: These are temporary recipes to run the benchmarks on those models. Will be cleaned up when PR is finalized

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -664,6 +666,13 @@ def _prep_net_for_train(self) -> None:
         self.resume_from_remote_sg_logger = core_utils.get_param(self.training_params, "resume_from_remote_sg_logger", False)
         self.load_checkpoint = self.load_checkpoint or self.external_checkpoint_path is not None or self.resume_from_remote_sg_logger
 
+        if self.training_params.torch_compile:
+            logger.info("Using torch.compile feature. Compiling model. This may take a few minutes")
+            self.net = torch.compile(self.net, mode=self.training_params.torch_compile_mode)
```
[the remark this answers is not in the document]
BloodAxe: ✅ Added check for torch 2.0+
➖ Adding try/catch is not quite possible because compile operates on the first forward & backward as far as I can tell and more importantly for some reason it does the second compile during validation 🤷 I'm not sure whether this is an expected behavior or not, but this is what I'm observing. Long story short, we can see whether compile actually worked or not only inside the train loop. Do we really want to add try/catch everywhere? I feel this would complicate the logic quite a lot
✅ Added `torch_compile_loss`, but tried on PPYoloE loss with no luck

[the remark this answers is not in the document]
shaydeci: Alrighty, thanks

src/super_gradients/training/params.py
```diff
@@ -64,11 +64,12 @@
     # (i.e iterating over train_loader) when reaching this number of batches.
     "max_valid_batches": None,  # For debug- when not None- will break out of inner valid loop
     # (i.e iterating over valid_loader) when reaching this number of batches.
-    "resume_from_remote_sg_logger": False  # When true, ckpt_name (checkpoint filename to resume, ckpt_latest.pth by
+    "resume_from_remote_sg_logger": False,  # When true, ckpt_name (checkpoint filename to resume, ckpt_latest.pth by
     # default) will be downloaded into the experiment checkpoints directory prior to loading weights, then resumed
     # from that checkpoint. The source is unique to every logger, and currently supported for WandB loggers only.
     # Note that for this to work, the experiment must be ran with sg_logger_params.save_checkpoints_remote=True. For
     # WandB loggers, one must also pass the run id through the wandb_id arg in sg_logger_params.
+    "torch_compile": False,
```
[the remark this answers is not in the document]
BloodAxe: Fixed in 48c11d94da10b755af8ea3f113ef61aa55dcd3aa

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -661,6 +662,20 @@ def _prep_net_for_train(self) -> None:
...
+                self.net = torch.compile(self.net, **self.training_params.torch_compile_options)
+                logger.info("Model compilation complete. Continuing training")
+                if is_distributed():
+                    torch.distributed.barrier()
+            else:
+                logger.warning(
+                    "Your recipe has requested use of torch.compile. "
+                    "However torch.compile is not supported in this version of PyTorch. "
+                    "Ignoring torch_compile flag"
+                )
```
Louis-Dupont, line 677: Good logs, but I would also print the required version in the warning to let the user know how to update and actually use torch.compile

Louis-Dupont (dismissed):
Looks great

src/super_gradients/training/sg_trainer/sg_trainer.py
BloodAxe: Fixed in 4b584f5db0809b12fa422fd93a33a38bcaa3d735

BloodAxe:
@Louis-Dupont I've updated the warning message to include version requirement. Unfortunately it dismissed approval state. Can you please re-approve?

Louis-Dupont (approved):
LGTM

Issue #999, opened 2023-05-14
Error: 'charmap' codec can't encode characters


T0T4R4:
Hi !

I'm following your tutorial for fine-tuning, but received the following error during training on my GPU :

```
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\super_gradients\training\sg_trainer\sg_trainer.py", line 1273, in train
    validation_results_tuple = self._validate_epoch(epoch=epoch, silent_mode=silent_mode)
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\super_gradients\training\sg_trainer\sg_trainer.py", line 1759, in _validate_epoch
    return self.evaluate(
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\super_gradients\training\sg_trainer\sg_trainer.py", line 1870, in evaluate
    sg_trainer_utils.display_epoch_summary(
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\super_gradients\training\utils\sg_trainer_utils.py", line 257, in display_epoch_summary
    summary_tree.show()
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\treelib\tree.py", line 854, in show
    print(self._reader)
  File "C:\Python310\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 20-22: character maps to <undefined>

```

I tried to just comment the show() function, but at then end of the whole training i got that error :  

```
Exception ignored in atexit callback: <function reset_all at 0x000001DE4A6357E0>
Traceback (most recent call last):
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\colorama\initialise.py", line 34, in reset_all
    AnsiToWin32(orig_stdout).reset_all()
  File "C:\Users\clert\Documents\_Dev\yolo-nas\venv\lib\site-packages\colorama\ansitowin32.py", line 189, in reset_all
    self.wrapped.write(Style.RESET_ALL)
  File "C:\Python310\lib\codecs.py", line 378, in write
    self.stream.write(data)
TypeError: write() argument must be str, not bytes
```

### Environment:
 - OS : Windows 11
 -  GPU : RTX 2060
 - Super Gradients version 3.1.1
[...]

PR #1175, opened 2023-06-15
Fix missing encoding

allankouidri:
Fixing [Error: 'charmap' codec can't encode characters](https://github.com/Deci-AI/super-gradients/issues/999#top)


BloodAxe (approved):
LGTM

PR #1176, opened 2023-06-15
Support Predict on Classification Models

danielafrimi:
Until now the changes made only for Resnet (HasPredict interface).

1. Basic script to run a classifier with predict functionalities.
2. Created ClassificationPipeline
    - _decode_model_output - call argmax on he model output, needs to pass confidence via softmax activations. in 
    addition we create ClassificationPrediction per prediction with confidence, label and image_shaoe attr.
    - _instantiate_image_prediction
    - _combine_image_prediction_to_images
    - _combine_image_prediction_to_video -> we dont need it to classifier, what should be the implementation? we wont 
    call predict_video. 

3. Created Prediction objects for classification (with the implementation of show, save and draw functions).
4. Created some processes according to validation transforms (implement Resize, CenterCrop) + default_resnet_imagenet_processing_params() which create the image_processor.

Up Next
1. Creating a middle class BaseClassificationModel that inherit from Sg.moudle and will preform as base class for all the classifier models (which inherit from Sg.moudle). this BaseClassificationModel obj will  include the HasPredict interface function. - **Done**
2. Drawing top k with confidence on the image itself. 
3. default_resnet_imagenet_processing_params -> make it generic to classification models (needs to add more processing functions).
4. create a function of drawing in vizualization dir (use it in draw function in Prediction obj) - **Done**
5. run black + flake8 - **Done**


BloodAxe:
It looks like the code you've pushed is not properly formatted. We are using black and pre-commit hooks to ensure that code is formatted identically for everyone.  Feel free to check https://github.com/Deci-AI/super-gradients/blob/master/CONTRIBUTING.md#pre-commit-hooks on configuring hooks and running black

src/super_gradients/training/utils/predict/prediction_results.py
```diff
@@ -41,6 +42,70 @@ def save(self, *args, **kwargs) -> None:
...
+    :attr image:        Input image
+    :attr predictions:  Predictions of the model
+    :attr class_names:  List of the class names to predict
+    """
+
+    image: np.ndarray
+    prediction: ClassificationPrediction
+    class_names: List[str]
+
+    def draw(self, show_confidence: bool = True) -> np.ndarray:
```
BloodAxe, line 58: I suggest to move the actual implementation of "draw label overlay on top of the image" to regular function similar to how `draw_bbox` is made:
`draw_image_with_header(image, text, text_color) -> np.ndimage`
And use it here. 
Pros: If we at some point want to reuse the drawing funciton we would simply call that function.

danielafrimi: Thanks. ill fix the black formatter (its still a draft).

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -377,3 +379,64 @@ def _combine_image_prediction_to_video(
...
+            images_predictions = [next(iter(images_predictions))]
+        else:
+            images_predictions = [image_predictions for image_predictions in tqdm(images_predictions, total=n_images, desc="Predicting Images")]
+
+        return ImagesClassificationPrediction(_images_prediction_lst=images_predictions)
+
+    def _combine_image_prediction_to_video(
+        self, images_predictions: Iterable[ImageDetectionPrediction], fps: float, n_images: Optional[int] = None
+    ) -> VideoDetectionPrediction:
+        # TODO we dont need this in classification task
```
danielafrimi, line 440: In CalssifiocationPipeline we dont need to implement this function. what should i code their?

Louis-Dupont: Maybe something like `raise NotImplementedError('This feature is not available for Classification task')` ?

src/super_gradients/training/utils/predict/predictions.py
```diff
@@ -106,3 +106,34 @@ def _validate_input(self, poses: np.ndarray, scores: np.ndarray, edge_links, edg
...
     def __len__(self):
         return len(self.poses)
+
+
+@dataclass
+class ClassificationPrediction(Prediction):
+    """Represents a Classification prediction"""
+
+    confidence: np.float32
+    labels: np.int64
```
Louis-Dupont, line 116: why not int/float instead ?

src/super_gradients/training/utils/predict/predictions.py
```diff
@@ -106,3 +106,34 @@ def _validate_input(self, poses: np.ndarray, scores: np.ndarray, edge_links, edg
...
+
+        :param confidence:  Confidence scores for each bounding box
+        :param labels:      Labels for each bounding box.
+        :param image_shape: Shape of the image the prediction is made on, (H, W).
+        """
+        self._validate_input(confidence, labels)
+
+        self.confidence = confidence
+        self.labels = labels
+        self.image_shape = image_shape
```
Louis-Dupont, line 130: FYI `@dataclass` actually provide a `__post_init__` method meant for this type of thing. The advantage is that it doesnt override the default `@dataclass` `__init__`, so you don't need to write all this `self.confidence = confidence` logic.

I think this is the syntax
```python
def __post_init__(self, confidence: np.float32, labels: np.int64, image_shape: Optional[Tuple[int, int]]):
    self._validate_input(confidence, labels)
```

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -286,7 +289,7 @@ def _decode_model_output(self, model_output: Union[List, Tuple, torch.Tensor], m
         return predictions
 
     def _instantiate_image_prediction(self, image: np.ndarray, prediction: DetectionPrediction) -> ImagePrediction:
-        return ImageDetectionPrediction(image=image, prediction=prediction, class_names=self.class_names)
+        return ImageClassificationPrediction(image=image, prediction=prediction, class_names=self.class_names)
```
Louis-Dupont, line 292: I think that's a mistake

src/super_gradients/training/models/classification_models/base_classifer.py
```diff
@@ -0,0 +1,65 @@
...
+from super_gradients.common.factories.processing_factory import ProcessingFactory
+from super_gradients.training.processing.processing import Processing
+from typing import Optional, List
+
+
+class BaseClassifier(SgModule):
+    def __init__(
+        self,
+    ):
+        super(BaseClassifier, self).__init__()
```
Louis-Dupont, line 17: (Not important) You can skip this

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -377,3 +380,63 @@ def _combine_image_prediction_to_video(
...
+
+    def __init__(
+        self,
+        model: SgModule,
+        class_names: List[str],
+        device: Optional[str] = None,
+        image_processor: Optional[Processing] = None,
+        fuse_model: bool = True,
+    ):
+        super().__init__(model=model, device=device, image_processor=image_processor, class_names=class_names, fuse_model=fuse_model)
```
Louis-Dupont, line 404: Should the model be `BaseClassifier` instead? (Same for doc ` The classification model (instance of SgModule)`)

Louis-Dupont (commented):
Nice! I just left some comments, let me know what you think but afterward I think we can merge

src/super_gradients/training/pipelines/pipelines.py
Louis-Dupont: @danielafrimi looks like you should undo this change and go back to using `ImageDetectionPrediction` here

src/super_gradients/training/pipelines/pipelines.py
Louis-Dupont: `model: BaseClassifier` instead of `model: SgModule`, no?

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -377,3 +381,61 @@ def _combine_image_prediction_to_video(
...
+            images_predictions = [next(iter(images_predictions))]
+        else:
+            images_predictions = [image_predictions for image_predictions in tqdm(images_predictions, total=n_images, desc="Predicting Images")]
+
+        return ImagesClassificationPrediction(_images_prediction_lst=images_predictions)
+
+    def _combine_image_prediction_to_video(
+        self, images_predictions: Iterable[ImageDetectionPrediction], fps: float, n_images: Optional[int] = None
+    ) -> VideoDetectionPrediction:
+        raise NotImplementedError("This feature is not available for Classification task")
```
Louis-Dupont, line 441: Just realized that the typing is not correct here

Louis-Dupont (approved):
LGTM

Louis-Dupont:
Merged in https://github.com/Deci-AI/super-gradients/pull/1220 because no signed commits in this branch

PR #1178, opened 2023-06-15
Fix collections.Iterable -> typing.Iterable to fix crash in python 3.10

BloodAxe:
In Colab notebooks Python 3.10 is already a thing. And importing collections.Iterable does not work anymore. This fixes broken import


ofrimasad (approved):
LGTM

PR #1181, opened 2023-06-16
Feature/sg 708 time units

BloodAxe:
This PR adds the possibility to log scalar values with explicit time units associated with it.
It enables us to log loss value per step, for instance. The motivation of having this feature is three-fold:
1) It allow us to log losses per step (Which can be quite useful to spot instabilities and high loss spikes during training)
2) It also allow us to use TimerCallback that measures and log batch time (including forward & backward passes) 
3) Some loggers (W&B in particular) has a technical limitation that force you to provice monotonicaly increasing time steps when logging. E.g you cannot log losses for batch 0,1,2,3,4...90 and then log averaged loss per epoch=0. With this PR is is now possible.

torch.compile PR also depends on this PR (We are using TimerCallback to measure training speedup).


shaydeci (approved):
LGTM

Issue #1163, opened 2023-06-12
Missing keys when loading from checkpoint


ikin5050:
### 💡 Your Question

I have finetuned a model based on yolonas_l. When I attempt to load this model using the following code:

 `model = models.get(
    model_name='yolo_nas_l',
    checkpoint_path='checkpoints/yolo_nas_l/Person_ckpt_best.pth',
    num_classes=5
).to(device)`

I get the error message:

`RuntimeError: Error(s) in loading state_dict for YoloNAS_L:
	Missing key(s) in state_dict: "backbone.stem.conv.branch_3x3.conv.weight", "backbone.stem.conv.branch_3x3.bn.weight", "backbone.stem.conv.branch_3x3.bn.bias", "backbone.stem.conv.branch_3x3.bn.running_mean", "backbone.stem.conv.branch_3x3.bn.running_var", "backbone.stem.conv.branch_1x1.weight", ...`

Could anyone explain what I am doing wrong in loading the checkpoints here?

whittenator:
@ikin5050 I am also getting the same issue!

nrepesh:
@ikin5050 I am also getting the same issue.

BloodAxe:
Try updating SG to latest release, where we fixed the issue of loading checkpoints and retrain your model. This should go away

Issue #1153, opened 2023-06-09
How to convert YOLO NAS .pth output into Tflite.


adkbbx:
### 💡 Your Question

I was trying to convert the yolo nas output of my custom object detection model from .pth into .tflite by refering to this [repo](https://github.com/sithu31296/PyTorch-ONNX-TFLite) by converting it into  ONNX -> TF -> Tflite.

```
!pip install super-gradients==3.1.0
!pip install imutils
!pip install pytube --upgrade
!pip install --upgrade pillow==9.2.0

import torch
from super_gradients.training import models

dataset_params = {
    'data_dir':'/content/drive/MyDrive/Numberplate/Yolo/Yolov8/data',
    'train_images_dir':'train/images',
    'train_labels_dir':'train/labels',
    'val_images_dir':'validate/images',
    'val_labels_dir':'validate/labels',
    'test_images_dir':'test/images',
    'test_labels_dir':'test/labels',
    'classes': ['blue','red','green','black','yellow','white']
}


model = models.get('yolo_nas_l', 
                   num_classes=len(dataset_params['classes']), 
                   pretrained_weights="coco"
                   ) 


pt_model_path = "/path/to/ckpt_best.pth" 
model.load_state_dict(torch.load(pt_model_path, map_location='cpu'))
```

 but there seems to be a runtime error when I try to load the model :

```
RuntimeError: Error(s) in loading state_dict for YoloNAS_L:
[...]

PR #1184, opened 2023-06-16
Getting rid of "module." heritage

BloodAxe:
# What this PR does

We do not store state_dict of the model with "module." prefix anymore

# Why this is important

- When training with DDP/DP mode, DistributedDataParallel/DataParallel wrappers adds this "module." prefix unintentionally to `state_dict`. So there goes discrepancy when the model is trained with or without DDP. Previously SG was solving this with a hack of adding artificial "module." prefix to all saved models in single GPU mode. However, this is a broken design

- Some of our checkpoints are saved with and some are saved without "module." which I believe the main cause we have this ugly NO_KEY_MATCHING checkpoint loading hack. The actual reasoning is lost in time, but I'm, pretty sure this is because of this discrepancy of saving models with and without DDP wrapper.

# Solution

1. This PR introduce `get_real_model` method to return the "real" model if it is wrapped with DP/DDP. Whenever there is a need to save state_dict one should use this method to unwrap the model and save state dict of the real one. 
2. Similarly, when loading a checkpoint same logic should be performed: get_real_model(net).load_state_dict(...)
4. Compatibility with existing checkpoints that may or may not has "module." prefix is secured buy checking whether all state_dict keys are starting with "module.". If so - we drop them and load checkpoint as usual. This handles the case when checkpoint was saved with DDP wrapper.

# Testing

Not properly tested yet
Components requiring special attention
- [x] Regular trainer
- [ ] KDTrainer
- [x] EMA
- [x] Pretrained models
- [x] Unit Tests
- [x] Integration Tests
- [x] Shortened Recipe Tests

# Risks of introducing breaking changes

No risk assessment has been done yet

# Related issues

https://github.com/Deci-AI/super-gradients/issues/1163
https://github.com/Deci-AI/super-gradients/issues/1153


Louis-Dupont (commented):
Looks great, I am just not sure about the name `get_real_model`, I think it is pretty confusing. 
Maybe `unwrap_model`? It's also not that good, but maybe slightly more explicit? Any better idea?

BloodAxe:
`unwrap_model` is good, I like this name)

Louis-Dupont (approved):
LGTM

Makefile
ofrimasad, line 1: why was this file changed?

ofrimasad (approved):
LGTM 
not sure if the Makefile was changed by mistake

Makefile
BloodAxe: The change of Makefile was due to wrong executable file. I copied these lines from circle CI, but  on other machines python 3.8 may not be available, so the right excutable is just `python`, which would guarantee we are running against default python from currently active environment.
I used this to run all unit/integration/recipe tests locally on tzgas

ofrimasad (dismissed):
LGTM

Louis-Dupont (approved):
LGTM

PR #1192, opened 2023-06-19
default quantization params set in qat from config

shaydeci:
Before the change, we accidentally rais a warning instead of warn.
Additinoally we did not set quantisation_params in cfg to be the default ones.


ofrimasad (approved):
LGTM

PR #1196, opened 2023-06-20
Make  loading pretrained weights from the platform optional + fix head replacement for unknown datasets weights in platform

shaydeci:
- Made the download of the weights optional (and set to true for backward compatibility), Weights are downlaoded only whwn preatrined_weights-"platform/XXX" (thisi is what @ofrimasad and I agreed upon).
- Pass pretrained_weights longside checkpoint_num_classes to instnaitate the model then load the weights. Head replacement is done with num_classes (As talked with @eran-deci )

@eran-deci please review and verify this solution works for you.


BloodAxe (approved):
LGTM

PR #1202, opened 2023-06-22
Remove casting to uint8 during resize

BloodAxe:
This leads to hard-to-debug bugs in transforms, when image first normalized, then resized and then float image casted to byte.


Louis-Dupont (approved):
LGTM

PR #1218, opened 2023-06-27
Return Dict of loss/metrics instead of Tuple in the Trainer 

Louis-Dupont:
### Motivation
Some time ago (couple of month?) `trainer.train` and `trainer.test` output type was changed from tuple to dict.
Apparently, 2 internal functions (`_validate_epoch`, `_evaluate`) and `ModelWeightAveraging` still work with tuples.

This cannot be scaled when moving to multiple validation loaders so In this PR I update the input/output of these class/methods to work with dict instead of tuples


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -584,25 +584,21 @@ def _save_checkpoint(
...
             self.sg_logger.add_checkpoint(tag="ckpt_latest_weights_only.pth", state_dict={"net": self.net.state_dict()}, global_step=epoch)
             return
 
         # COMPUTE THE CURRENT metric
         # IF idx IS A LIST - SUM ALL THE VALUES STORED IN THE LIST'S INDICES
-        metric = (
-            validation_results_tuple[self.metric_idx_in_results_tuple]
-            if isinstance(self.metric_idx_in_results_tuple, int)
-            else sum([validation_results_tuple[idx] for idx in self.metric_idx_in_results_tuple])
-        )
```
Louis-Dupont, line 605: Not sure about what this condition what supposed to do, because `self.metric_idx_in_results_tuple` had to be an int, it was the fuzzy index of the `metric_to_watch`.
if `metric_to_watch` was not in this list, the code crashes. So from what I understand we always have an int.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -524,12 +523,22 @@ def _init_monitored_items(self):
...
+        # make sure the metric_to_watch is an exact match
+        metric_titles = self.loss_logging_items_names + get_metrics_titles(self.valid_metrics)
+        metric_to_watch_idx = fuzzy_idx_in_list(self.metric_to_watch, metric_titles)
+        metric_to_watch = metric_titles[metric_to_watch_idx]
+        if metric_to_watch != self.metric_to_watch:
+            logger.warning(
+                f"No exact match found for `metric_to_watch={self.metric_to_watch}`. It should be one of {metric_titles}. \n"
+                f"`metric_to_watch={metric_to_watch} will be used instead.`"
+            )
+            self.metric_to_watch = metric_to_watch
```
Louis-Dupont, line 535: Not mandatory, but I thought it is nice to let the user know that he has a typo, because we are not going to use the value of metric_to_watch  that he/she passed as an input and I don't think this should be silent.
But if you think it's too much (or that it would just be lost in the logs) we can take it out

shaydeci: Sure, why not.

BloodAxe (approved):
LGTM

PR #1221, opened 2023-06-27
Feature/sg 000 yolo nas dropout support

BloodAxe:
Adds the ability to specify dropout parameters in the YoloNAS model.
Compatibility with pretrained checkpoints remains intact. 
Default values of dropout probas are set to zero to keep existing training recipies intact.
But if users wants, he can customize these in arch params config.


src/super_gradients/modules/dropout_layers.py
```diff
@@ -0,0 +1,43 @@
...
+    >>> class ResNetBlock(nn.Module):
+    >>>   def __init__(self, ..., drop_path_rate:float):
+    >>>     self.drop_path = DropPath(drop_path_rate)
+    >>>
+    >>>   def forward(self, x):
+    >>>     return self.drop_path(x) + self.conv_bn_act(x)
+    """
+
+    def __init__(self, drop_prob: float = 0.0, scale_by_keep: bool = True):
+        super(DropPath, self).__init__()
```
Louis-Dupont, line 35: If you can just add description of params

BloodAxe: Thanks for the suggestion. While adding docs I realized we in fact already have the DropPath implementation. So I update it a little bit with docs and removed mine. But in terms of numerical output they are the same

Louis-Dupont (approved):
LGTM

PR #1222, opened 2023-06-27
Rewrite `_set_preprocessing_from_valid_loader` to `_get_preprocessing_from_valid_loader`

Louis-Dupont:
Rewriting `_set_preprocessing_from_valid_loader` to be `_get_preprocessing_from_valid_loader`, which that way can be used in 2 places instead of one

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -1349,10 +1349,7 @@ class Trainer:
                 self.sg_logger.close()
 
     def _get_preprocessing_from_valid_loader(self) -> Optional[dict]:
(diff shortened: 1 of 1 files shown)


Louis-Dupont:
@BloodAxe I removed some changes that were in this PR to make it more straightforward.
Eventually, it's about rewriting `_set_preprocessing_from_valid_loader` to be `_get_preprocessing_from_valid_loader`, which that way can be used in 2 places instead of one. 
I did this just because in the following PR I need to add a few more lines of codes to this `_get_preprocessing_from_valid_loader`, so I create the method before hand to make the following PR lighter

BloodAxe (approved):
LGTM

PR #1223, opened 2023-06-27
Allow passing tuple of [low, high] to iou_thresholds parameter in DetectionMetrics.

BloodAxe:
This PR allows to pass user-defined target of IoU thresholds without asking user to edit IoUThresholds enum in SG codebase:

```
valid_metrics_list:
  - DetectionMetrics:
      iou_thres: [0.5, 0.8]
```

Note that this does not solve the problem of having two instances of `DetectionMetrics` in `valid_metrics_list`, but it allow to pass iou_thres as a tuple which we do not allow at the moment


Louis-Dupont (approved):
LGTM

PR #1227, opened 2023-06-27
Fix DEKR's replace_head & improve __repr__ for keypoints transforms

BloodAxe:
Add missing replace_head implementation for DEKR
Add human-friendly __repr__ to keypoint transforms which should improve the readability of transforms in colab when user prints them (For Colab demo)


src/super_gradients/training/transforms/keypoint_transforms.py
```diff
@@ -71,6 +71,14 @@ def get_equivalent_preprocessing(self) -> List:
             preprocessing += t.get_equivalent_preprocessing()
         return preprocessing
 
+    def __repr__(self):
+        format_string = self.__class__.__name__ + "("
+        for t in self.transforms:
+            format_string += "\n"
+            format_string += "    {0}".format(repr(t))
```
Louis-Dupont, line 78: Just curious, why not directly `format_string += f"    {t}"` ?

BloodAxe: I just hit TAB to accept tabnine suggestion 🙈 
But you're right, this can be done better

src/super_gradients/training/transforms/keypoint_transforms.py
```diff
@@ -353,6 +386,14 @@ def __init__(
         self.mask_pad_value = mask_pad_value
         self.prob = prob
 
+    def __repr__(self):
+        return (
+            self.__class__.__name__
+            + "(max_rotation={0}, min_scale={1}, max_scale={2}, max_translate={3}, image_pad_value={4}, mask_pad_value={5}, prob={6})".format(
+                self.max_rotation, self.min_scale, self.max_scale, self.max_translate, self.image_pad_value, self.mask_pad_value, self.prob
+            )
+        )
```
Louis-Dupont, line 395: Similar question as above, I tend to think that `. format` is quite hard to read actually, especially with so many params, and only use it for "templates". Is there any reason for not using f-string ?
To make it clear I'm not asking to change, I think it does not matter here ✌️

src/super_gradients/training/transforms/keypoint_transforms.py
BloodAxe: Fixed 99fd7f6d983b4f151b17c39feb78c704ab89f66d

Louis-Dupont (dismissed):
LGTM

Louis-Dupont (approved):
LGTM

PR #1230, opened 2023-06-29
Feature/sg 946 simpify torch schedulers integration

shaydeci:


documentation/source/LRScheduling.md
```diff
@@ -202,36 +205,89 @@ Note that internally, Trainer unpacks [training_params to the scheduler callback
...
 PyTorch offers a [wide variety of learning rate schedulers](https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate).
-They can all be easily used by wrapping them up with `LRSchedulerCallback` and passing them as phase_callbacks.
+They can all be easily used by passing a Mapping through the lr_mode parameter, following aa simple API.
+From `Trainer.train(...)` docs:
+
+    When Mapping, refers to a torch.optim.lr_scheduler._LRScheduler, following the below API:
+    
+        lr_mode = {LR_SCHEDULER_CLASS_NAME: {**LR_SCHEDULER_KWARGS, "phase": XXX, "metric_name": XXX)
+    
+        Where "phase" (of Phase type) controls when to call torch.optim.lr_scheduler._LRScheduler.step().
```
BloodAxe, line 215: From this line it is not immediately clear what are available options are. In fact, Phase enum contains too many options. 
Usually there are only two options: batch step (E.g updating LR after each gradient update step) and epoch step (Updating LR after the epoch). I can easily see one can get lost with all values Phase enum offers here.

BloodAxe: So maybe would be helpful to add tips:
- I want to update LR on each batch:    Use `phase: Phase.TRAIN_BATCH_END`
- I want to update LR after each epoch: Use `phase: Phase.TRAIN_EPOCH_END`

tests/deci_core_unit_test_suite_runner.py
```diff
@@ -143,6 +144,7 @@ def _add_modules_to_unit_tests_suite(self):
         self.unit_tests_suite.addTest(self.test_loader.loadTestsFromModule(DeprecationsUnitTest))
         self.unit_tests_suite.addTest(self.test_loader.loadTestsFromModule(TestMinSamplesSingleNode))
         self.unit_tests_suite.addTest(self.test_loader.loadTestsFromModule(TestSegmentationMetricsMultipleIgnored))
+        self.unit_tests_suite.addTest(self.test_loader.loadTestsFromModule(TrainWithTorchSchedulerTest))
```
BloodAxe, line 147: This is great we are adding tests cases. 💪 
One test that I really really want to see here - test cases for all pytorch LR schedulers that we declare support. Not only this would give us some peace of mind that all of them are working properly but also can act as an usage examples.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1211,6 +1237,23 @@ def forward(self, inputs, targets):
...
+                net=self.net,
+                training_params=self.training_params,
+                update_param_groups=self.update_param_groups,
+                optimizer=self.optimizer,
+            )
+            self.phase_callbacks.append(lr_scheduler_callback)
+
+            # NEED ACCESS TO THE UNDERLYING TORCH SCHEDULER FOR LOADING/SAVING IT'S STATE_DICT
+            if isinstance(lr_scheduler_callback, LRSchedulerCallback):
+                self._torch_lr_scheduler = lr_scheduler_callback.scheduler
```
BloodAxe, line 1253: What happes to the previous value of the `self._torch_lr_scheduler`?
What if it was None and become a non-None? What side effects could be happening here?

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -863,3 +864,89 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
...
+    :return: a PhaseCallback instance to be used by Trainer for LR scheduling.
+    """
+
+    if isinstance(lr_mode, str) and lr_mode in LR_SCHEDULERS_CLS_DICT:
+        sg_lr_callback_cls = LR_SCHEDULERS_CLS_DICT[lr_mode]
+        sg_lr_callback = sg_lr_callback_cls(
+            train_loader_len=len(train_loader),
+            net=net,
+            training_params=training_params,
+            update_param_groups=update_param_groups,
```
BloodAxe, line 932: Since `update_param_groups` can be easily inferred from the `net` itself via `update_param_groups = hasattr(unwrap_model(self.net), "update_param_groups")` we can safely drop this parameter and infer it's value inside the callback implementation itself.
I think this can simplify the code here.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -863,3 +864,89 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
...
+            **training_params.to_dict(),
+        )
+    elif isinstance(lr_mode, Mapping) and list(lr_mode.keys())[0] in TORCH_LR_SCHEDULERS:
```
BloodAxe, line 935: A) Are Torch LR schedulers compatible with model-specific LR update rules (update_param_groups = True?)
B) If not we should probably emit a warning/exception?

tests/deci_core_unit_test_suite_runner.py
BloodAxe: Second wish that came to my mind after looking through the PR - we should test torch schedules on regular models (subclassing from nn.Module) and on our ones with custom LR per layer (that has this `update_param_groups` thingie)

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: True, but rather keep this out of this PR.
update_param_groups is only used by our SG scheduler callbacks, so I don't feel like this change (although needed) should be here as it requires changing (and perhaps breaking code) of our LR scheduler callbacks.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: It is only initiated with None.
Then, if we are using a torch scheduler it is set to it here.
So, the only time it would not be None, is if we call train() twice, and then we need to point to the right scheduler.

tests/deci_core_unit_test_suite_runner.py
shaydeci: Added testss for all of the torch schedulers in aa6e2347f8afc0793084e1488958510611c67d29

Regarding the second point - I think torch schedulers should not work (Ill add a warning) alongside update_param_groups.

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: nice catch, added warning in e1ad971733d919ed532d4641b69932a66714984a

documentation/source/LRScheduling.md
shaydeci: added your suggestion to docs in 2e9df8ac80feacac1b551dfb3c2961dcc68bd7c4

BloodAxe (approved):
LGTM

PR #1233, opened 2023-06-30
Remove duplicated context

BloodAxe:
- Removes a copy of context created in _train_epoch
- Removes `context_methods`, they are not in use anyway


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2033,36 +2017,6 @@ def set_ema(self, val: bool):
...
-                set_net=self.set_net,
-                set_ckpt_best_name=self.set_ckpt_best_name,
-                reset_best_metric=self._reset_best_metric,
-                validate_epoch=self._validate_epoch,
-                set_ema=self.set_ema,
-            )
-        else:
-            context_methods = ContextSgMethods()
-
-        return context_methods
```
Louis-Dupont, line 2064: What did this do? I mean, why did we need/or at least use it, and we don't anymore

Louis-Dupont (approved):
LGTM

src/super_gradients/training/sg_trainer/sg_trainer.py
BloodAxe: Oh, that was quite an interesting piece of code & logic.
So the idea here was to limit access to certain elements of the context for a certain phase. 
I'm not sure why this was done in the first place, but at the current state of the code it is not used anywhere and anyhow this is the internal API which is not exposed anywhere to public use.

shaydeci: This was done for when we had the QAT implemented as a phase callback.
Back then, we had to get_net, then replace the blocks with QA blocks, call set_net, call set_ckpt_best_name to set it to ckpt_best_qat.pth (to not override the ones from original training), call validate_epoch for calibration and turn off ema.

Don't think there was any other use of it. But I would be careful.

PR #1248, opened 2023-07-06
 num classes specified was wrong

pranoyr:
Num classes was wrong in the doc


BloodAxe:
@spsancti can you please confirm this is indeed the correct number of classes?
@pranoyr please setup signed commits for PR 🙏  https://github.com/Deci-AI/super-gradients/blob/master/CONTRIBUTING.md#signed-commits Thanks!

PR #1260, opened 2023-07-10
moved exceptions from training

shaydeci:
Had a circular import issue when tried to introduce new exception + crash tip.
Exceptions should be in common, not training,


BloodAxe (approved):
LGTM

Issue #1269, opened 2023-07-11
RuntimeError: Input type (unsigned char) and bias type (struct c10::Half) should be the same


DanielaGomez98:
### 💡 Your Question

Hi👋🏻 I am currently working on a project using YOLO NAS to detect diseases in cassava. I am using the *yolo_nas_l* architecture (also tried *yolo_nas_m*), and the task is object detection. The dataset is customized, and the annotations are in the correct format in TXT. The problem is when I make the prediction with new data, after training. I am trying run this code:
```
for image in valid_images_files:
    best_model.to(device).predict(str(image), conf=0.02).show()
```

and I get the following error:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[48], line 2
      1 img_url = 'https://www.mynumi.net/media/catalog/product/cache/2/image/9df78eab33525d08d6e5fb8d27136e95/s/e/serietta_usa_2_1/www.mynumi.net-USASE5AD160-31.jpg'
----> 2 best_model.predict(img_url).show()

File [c:\Users\danielagomez\AppData\Local\anaconda3\envs\yolo-nas\lib\site-packages\super_gradients\training\models\detection_models\customizable_detector.py:177](file:///C:/Users/danielagomez/AppData/Local/anaconda3/envs/yolo-nas/lib/site-packages/super_gradients/training/models/detection_models/customizable_detector.py:177), in CustomizableDetector.predict(self, images, iou, conf, fuse_model)
    168 """Predict an image or a list of images.
    169 
    170 :param images:  Images to predict.
   (...)
    174 :param fuse_model: If True, create a copy of the model, and fuse some of its layers to increase performance. This increases memory usage.
    175 """
    176 pipeline = self._get_pipeline(iou=iou, conf=conf, fuse_model=fuse_model)
--> 177 return pipeline(images)

File [c:\Users\danielagomez\AppData\Local\anaconda3\envs\yolo-nas\lib\site-packages\super_gradients\training\pipelines\pipelines.py:96](file:///C:/Users/danielagomez/AppData/Local/anaconda3/envs/yolo-nas/lib/site-packages/super_gradients/training/pipelines/pipelines.py:96), in Pipeline.__call__(self, inputs, batch_size)
     94     return self.predict_video(inputs, batch_size)
     95 elif check_image_typing(inputs):
---> 96     return self.predict_images(inputs, batch_size)
     97 else:
[...]

PR #1281, opened 2023-07-14
Added explicit casting of input in predict pipeline.

BloodAxe:
This is necessary if for some reason the input's preprocessing pipeline does not have normalization, so input stays in uint8 format.
Example: 
Input image has [0..255] range for uint8 dtype
Validation dataset do not have normalization/standardization.
torch_inputs in this case would have torch.uint8 dtype and that would crash

Solution: Take the model's dtype of first parameter (That is likely to be the first layer taking the input) and cast input's dtype to this type.

Fixes: https://github.com/Deci-AI/super-gradients/issues/1269


shaydeci (approved):
LGTM

PR #1282, opened 2023-07-16
Support for segmentation extreme batch cases

shaydeci:
- Added a base class, for visualizing the worst/best batches according to some metric, loss or loss component.
- Added segmentation support.
- Added some missing pieces that were needed in phase context.
- Full DDP support- aggregation of metrics, the best/worst predictions, and averaging the loss.

Note: images can be resized individually inside the logger.

Screen Shot 2023-07-16 at 14 35 58


src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +953,205 @@ def create_lr_scheduler_callback(
...
+        """
+        if not hasattr(context.valid_metrics, self.metric_name):
+            for metric_name, metric in context.valid_metrics.items():
+                if hasattr(metric, "greater_component_is_better") and self.metric_name in metric.greater_component_is_better.keys():
+                    # WRAP METRIC WITH METRIC COLLECTION TO FILTER ONLY THE NEEDED ARGUMENTS FOR THE METRIC UPDATE
+                    self.metric = MetricCollection(copy.deepcopy(metric))
+                    self.metric.to(device_config.device)
+        else:
+            self.metric = MetricCollection(copy.deepcopy(getattr(context.valid_metrics, self.metric_name)))
+            self.metric.to(device_config.device)
```
Louis-Dupont, line 1020: I think that this should be the responsibility of the Metrics.
This is highly coupled to the Metric implementation and if one day we want to change it then we would also need to update this piece of code (and other places where we do things similarly).

Maybe we should implement some sort of a "SgMetricColletion" which would inherit from `MetricCollection` but on top of that also includes useful methods such as `metrics.get(metric_name=...)`

I know it's not the scope of this task, but at the same time this is way too coupled to the metrics logic and I feel like this will backfire at us later

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +953,205 @@ def create_lr_scheduler_callback(
...
+        self.extreme_targets = None
+
+        self._first_call = True
+        self._idx_loss_tuple = None
+        self._tag = f"max_{self.metric_name}_batch" if self.max else f"min_{self.metric_name}_batch"
+
+        super(ExtremeBatchCaseVisualizationCallback, self).__init__()
+
+    def process_extreme_batch(self) -> np.array:
+        raise NotImplementedError
```
Louis-Dupont, line 1004: I would rather have `@abstractmethod` (along with `ExtremeBatchCaseVisualizationCallback(Callback, ABC)` ) because this will raise an exception on instantiation and not just on call.
It will also help people implement new methods because it is more explicit and IDE detects it

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +953,205 @@ def create_lr_scheduler_callback(
...
+                # FOR LOSS VALUES, GET THE RIGHT COMPONENT, DERRIVE IT ON THE FIRST PASS
+                loss_tuple = context.loss_log_items
+                if self._first_call:
+                    self._init_loss_attributes(context, loss_tuple)
+                score = loss_tuple[self._idx_loss_tuple]
+
+                # IN CONTRARY TO METRICS - LOSS VALUES NEED TO BE REDUCES IN DDP
+                if is_distributed():
+                    device = next(context.net.parameters()).device
+                    score = distributed_all_reduce_tensor_average(tensor=score.to(device), n=torch.distributed.get_world_size())
```
Louis-Dupont, line 1040: Note for later: I think it would be great to have a nice and simple object that would wrap the loss which would abstract away whether it is distributed or not. `loss[idx]` would be responsible to check if it is distributed or not, and if so, it would call`distributed_all_reduce_tensor_average` behind the scene. It would make the code more simple to navigate but also less error-prone.

Louis-Dupont (commented):
LGTM overall; but I think the issue I raised with the metrics is quiet important. I know we currently don't have a clean way to handle the metrics but I think we should honestly think about it

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: Agree, changed in 9bb4c9b0957b9cffb27e10a96450a0c47c5235ec

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +954,206 @@ def create_lr_scheduler_callback(
...
+    def process_extreme_batch(self) -> np.array:
+        raise NotImplementedError
+
+    def on_training_start(self, context: PhaseContext) -> None:
+        """
+        On train start we set the metric (if the metric_name does not corresponf to a loss).
+        :param context: Phase context
+        :return:
+        """
+        if not hasattr(context.valid_metrics, self.metric_name):
```
BloodAxe, line 1014: I feel that logic of juggling with metrics here is unnecessary complicated. SRP and DRY are just crying in corner. I was not able to wrap my head around after reading this code first time. I was not successful second time either.
To me it should be done in a completely different manner which I believe would end up with a much more clearer implementation.
Here's what I suggest.

1) Introduce a concept of "batch metrics", which in the most simple case is a dictionary of `{metric_name1: metric value, metric_name2: metric value2, ...}`.
This concept is to be managed by a Trainer class and available inside the `context`. So all the per-batch metrics (Which is loss values and metrics that we can compute per-batch) should be available through this interface.
Pros: In the callback class now have only the simple access to metric:
```
if _is_more_extreme(context.batch_metrics[self.monitored_value]):
  self.extreme_score = score
  self.extreme_batch = context.inputs
  self.extreme_preds = context.preds
  self.extreme_targets = context.target
```

2) For IoU metric all we have to do is to ensure it is also computed per-batch and stored in `batch metrics` container. And that's it.

I feel this change would greatly improve the responsibility boundaries between classes and reduce the amount of duplicated code.

shaydeci: I did think of it, but I feel a bit worried about interfering with metrics inside the MetricCollection object, therefore I rather have it somehere else.

From what I understand, you suggest passing a dictionary that would suppply the information on what metrics we would like to compute per batch, correct?

There is a drawback here if so, since you are asking the user to pass additional info which IMO sould be coupled with this callback.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +954,206 @@ def create_lr_scheduler_callback(
...
+        self.extreme_targets = None
+
+        self._first_call = True
+        self._idx_loss_tuple = None
+        self._tag = f"max_{self.metric_name}_batch" if self.max else f"min_{self.metric_name}_batch"
+
+        super(ExtremeBatchCaseVisualizationCallback, self).__init__()
+
+    @abstractmethod
+    def process_extreme_batch(self) -> np.array:
```
BloodAxe, line 1005: Just by looking at this method it is unclear what is should do. Maybe a docstring or better name would help. 
Second question - I see this method have no arguments. What advantages you see in this design versus sending arguments explicitly?


PS: Type annotation is wrong, `np.array` is a method. Right type should be `np.ndarray`

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: Passing metric explicitly now.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +954,206 @@ def create_lr_scheduler_callback(
...
+            else:
+
+                # FOR LOSS VALUES, GET THE RIGHT COMPONENT, DERRIVE IT ON THE FIRST PASS
+                loss_tuple = context.loss_log_items
+                if self._first_call:
+                    self._init_loss_attributes(context, loss_tuple)
+                score = loss_tuple[self._idx_loss_tuple]
+
+                # IN CONTRARY TO METRICS - LOSS VALUES NEED TO BE REDUCES IN DDP
+                if is_distributed():
```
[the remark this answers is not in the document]
shaydeci: 6344f834a381dcf0a043151c665816fdf78097ec did similar.

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: Added in 6344f834a381dcf0a043151c665816fdf78097ec

src/super_gradients/training/metrics/segmentation_metrics.py
```diff
@@ -331,6 +331,13 @@ def update(self, preds, target: torch.Tensor):
         super().update(preds=preds, target=target)
 
 
+@register_metric("DIOU")
+class DIOU(IoU):
```
BloodAxe, line 335: What this class does, why is it here and why it has no docs?

BloodAxe (changes requested):
Two important changes that I think this PR is missing:
- Please add a unit tests that verity this callback works as intended.
- Also, adding a docs to show we have this feature would help to increase it's exposure.

src/super_gradients/training/utils/distributed_training_utils.py
```diff
@@ -412,3 +414,36 @@ def __init__(self):
...
+    :param device:
+    :param tensor:
+    :return:
+    """
+    if is_distributed():
+        tensor = distributed_all_reduce_tensor_average(tensor=tensor.to(device), n=torch.distributed.get_world_size())
+    return tensor
+
+
+def maybe_all_gather_np_images(image: np.ndarray) -> np.ndarray:
```
BloodAxe, line 433: I believe this method's signature is a bit confusing. 
From my perspective it should be a List[np.ndarray] and not single np.ndarray. 
And the implementation should be like this:

```
    if is_distributed():
        rank = get_rank()
        output_container = [None for _ in range(get_world_size())]
        all_gather_object(output_container, image)
        return output_container
    return [image]
```

Why:

1) Current implementation behaves differently for master/non-master nodes. Which may be ok for current use case, but may fail in other scenarios.

2) It concatenates arrays along first dim, while this may be undesirable/impossible. Example - I want to gather arrays of predicted boxes which may have different length along axis=0. In this case the code would not work.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +955,247 @@ def create_lr_scheduler_callback(
...
+    :param freq: int, epoch frequency to perform all of the above (default=1).
+
+
+    """
+
+    def __init__(
+        self,
+        metric: Optional[Union[Metric, Mapping, str]] = None,
+        metric_component_name: Optional[str] = None,
+        loss_name: Optional[str] = None,
```
BloodAxe, line 1171: Maybe give a `loss_name` a bit more clear name like `loss_to_monitor`?

src/super_gradients/training/utils/distributed_training_utils.py
shaydeci: Why does it behave differently ?
instead of returning np.array sized BCHW it returns B_sumranksCHW, the concatination will work for different sizes across dim0, so I dont get your point.

That is not the purpose of this method, the name of it and docs are pretty clear on that.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +955,247 @@ def create_lr_scheduler_callback(
...
+
+    True predictions will be marked with green, false ones with red.
+
+    Example usage in training_params definition:
+
+        training_hyperparams ={
+          ...
+          "phase_callbacks":
+            [ExtremeBatchSegVisualizationCallback(
+                metrice=IoU(20, ignore_idx=19)
```
BloodAxe, line 1125: Typo in `metrice`

src/super_gradients/training/metrics/segmentation_metrics.py
shaydeci: Accidentally left it there during development, removed now.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +955,247 @@ def create_lr_scheduler_callback(
...
+
+    :param freq: int, epoch frequency to perform all of the above (default=1).
+
+     Inheritors should implement process_extreme_batch which returns an image, as an np.array (uint8) with shape BCHW.
+    """
+
+    @resolve_param("metric", MetricsFactory())
+    def __init__(
+        self,
+        metric: Optional[Union[Metric, Mapping, str]] = None,
```
BloodAxe, line 997: The `metric` here has type annotation of `Metric` (OK), and `Mapping` (assuming it's Mapping[str, Metric]?) and `str` - ??? Down the road we have this assignment `self.metric = MetricCollection(self.metric)`. Would that still work when metric is str?

src/super_gradients/training/utils/distributed_training_utils.py
```diff
@@ -412,3 +414,36 @@ def __init__(self):
             ">>> setup_device(multi_gpu=MultiGPUMode.DISTRIBUTED_DATA_PARALLEL, num_gpus=...)"
         )
         super().__init__(self.message)
+
+
+def maybe_all_reduce_tensor_average(device: str, tensor: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
shaydeci: agree changed in 2b8674a4f34b48dae6485f8b4d4dcc1aee57135a

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: Yes thats because of the resolve_param.
I can drop it if it causes confusion.

shaydeci: Actually it should't be there logically, removed.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -948,3 +955,247 @@ def create_lr_scheduler_callback(
...
+     the minimum (default=False).
+
+    :param freq: int, epoch frequency to perform all of the above (default=1).
+
+
+    """
+
+    def __init__(
+        self,
+        metric: Optional[Union[Metric, Mapping, str]] = None,
```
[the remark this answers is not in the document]
shaydeci: I like the seperation we have now between loss_to_monitor and metric. To me, its more clear that metric should be passed explicitly and loss_to_monitor actually monitors loss from the context.

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: 0633b5556cee46e5962bcf5bdf6216720293ab91

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: fixed 0633b5556cee46e5962bcf5bdf6216720293ab91

src/super_gradients/training/utils/callbacks/callbacks.py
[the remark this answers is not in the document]
BloodAxe: Up to you )

BloodAxe (approved):
LGTM

PR #1292, opened 2023-07-18
Segformer swi

Yael-Baron:


src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -856,6 +856,22 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
             return total_steps_in_done + train_loader_length + context.batch_idx
 
 
+@register_callback(Callbacks.SLIDING_WINDOW_INFERENCE)
+class ChangeToSWI(Callback):
```
Louis-Dupont, line 860: Same as above, I would prefer a longer name that would be more explicit
`ChangeToSWI` -> `ChangeToSlidingWindowInference`.

To further improve the name, I think we can drop the `ChangeTo` and just name it `SlidingWindowInference` or `SlidingWindowValidation` (I think I prefer the second)
And finally, this is just a convention we have for callbacks, let's add `Callback` at the end of the class name.
`SlidingWindowInferenceCallback` or `SlidingWindowValidationCallback`

These are just ideas, feel free to share if you think you have a better name

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -362,6 +370,12 @@ def init_params(self):
                 nn.init.ones_(m.weight)
                 nn.init.zeros_(m.bias)
 
+    def enable_swi(self):
+        self.is_SWI = True
+
+    def disable_swi(self):
+        self.is_SWI = False
```
Louis-Dupont, line 377: Same as above, I would replace `swi` -> `sliding_window_inference` for more clarity

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -372,12 +386,58 @@ def _remove_auxiliary_heads(self):
...
+        else:
+            return self._forward(x)
+
+    def sliding_window(self, img: torch.Tensor, num_classes: int, stride: Tuple[int, int], crop_size: Tuple[int, int]) -> torch.Tensor:
+        """
+        Inference by sliding-window with overlap.
+
+        If h_crop > h_img or w_crop > w_img, the small patch will be used to
+        decode without padding.
+        """
```
Louis-Dupont, line 407: If you can add params description (:param <name>: <desc>)

Louis-Dupont (commented):
Is the PR for review ?
I had a look anyway and left some comments, but just minor notes on naming/doc

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -372,12 +386,58 @@ def _remove_auxiliary_heads(self):
...
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def _forward(self, x: torch.Tensor) -> torch.Tensor:
         features = self._backbone(x)
         out = self.decode_head(features)
         out = F.interpolate(out, size=x.shape[2:], mode="bilinear", align_corners=False)
         return out
 
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.is_SWI:
+            return self.sliding_window(x, self.num_classes, self.sliding_window_stride, self.sliding_window_size)
```
shaydeci, line 397: Seems like crop_size and stride args are redundant if we always use the attributes.

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -442,6 +504,8 @@ def __init__(self, arch_params: HpmStruct):
     "overlap_patch_pad": [3, 1, 1, 1],
     "eff_self_att_reduction_ratio": [8, 4, 2, 1],
     "eff_self_att_heads": [1, 2, 5, 8],
+    "sliding_window_size": (1024, 1024),
```
shaydeci, line 507: Why are we setting default sliding window parameters ?

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -372,12 +386,58 @@ def _remove_auxiliary_heads(self):
...
+                crop_logits = self._forward(crop_img)
+
+                if isinstance(crop_logits, torch.Tensor):
+                    crop_logits = (crop_logits,)
+
+                crop_logits = crop_logits[0]
+
+                preds += F.pad(crop_logits, pad=(int(x1), int(preds.shape[3] - x2), int(y1), int(preds.shape[2] - y2)))
+
+                count_mat[:, :, y1:y2, x1:x2] += 1
```
shaydeci, line 437: Please add unit tests covering the different scenarios in SWI, making sure we always cover the entire image (i,e count mat should never have a 0 entry)

src/super_gradients/training/models/segmentation_models/segformer.py
shaydeci: Also, make sure to adhere to PEP docs format (add Pycharm plugin for convinience).

shaydeci: Adding to @Louis-Dupont , it would be great to give some example of what goes down under the hood when we use sliding window inference.

src/super_gradients/common/object_names.py
```diff
@@ -148,6 +148,7 @@ class Callbacks:
     DEKR_VISUALIZATION = "DEKRVisualizationCallback"
     ROBOFLOW_RESULT_CALLBACK = "RoboflowResultCallback"
     TIMER = "TimerCallback"
+    SLIDING_WINDOW_INFERENCE = "ChangeToSWI"
```
shaydeci, line 151: Is this the name of the callback?

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -856,6 +856,22 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
             return total_steps_in_done + train_loader_length + context.batch_idx
 
 
+@register_callback(Callbacks.SLIDING_WINDOW_INFERENCE)
```
shaydeci, line 859: Why choose an inconsistent name? This just adds more confusion....

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -856,6 +856,22 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
...
+class ChangeToSWI(Callback):
+    """
+    Performing single-scale sliding window during inference at the end of training on the validation set.
+    """
+
+    def on_validation_loader_start(self, context: PhaseContext) -> None:
```
shaydeci, line 865: Please check if average_best_models=True like we talked, and if so actiavte a newly added phase in average best model validation.

shaydeci (changes requested):
Comments inline, about some we have already had a discussion (so just a reminder).
Still missing:
- Unit tests - see inline.
- Integration tests, pretrained model urls.
- New phase for the final valdiation on the average model.

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -372,12 +386,58 @@ def _remove_auxiliary_heads(self):
...
         out = F.interpolate(out, size=x.shape[2:], mode="bilinear", align_corners=False)
         return out
 
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.is_SWI:
+            return self.sliding_window(x, self.num_classes, self.sliding_window_stride, self.sliding_window_size)
+        else:
+            return self._forward(x)
+
+    def sliding_window(self, img: torch.Tensor, num_classes: int, stride: Tuple[int, int], crop_size: Tuple[int, int]) -> torch.Tensor:
```
BloodAxe, line 401: I have a small comment about naming a function. Usually it's a really good practice to start function name with a verb: `do_work`, `get_something`, `sleep(...)`, `print(...)`.
In this example the function name can be `forward_with_sliding_window`, `predict_tiled`, etc..

src/super_gradients/training/models/segmentation_models/segformer.py
BloodAxe: That may require refactoring the sliding inference logic to an external method to check whether accumulator and count_mat contains expected values, but I agree that should be tested thoroughly.

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -372,12 +386,58 @@ def _remove_auxiliary_heads(self):
...
+                x1 = max(x2 - w_crop, 0)
+                crop_img = img[:, :, y1:y2, x1:x2]
+                crop_logits = self._forward(crop_img)
+
+                if isinstance(crop_logits, torch.Tensor):
+                    crop_logits = (crop_logits,)
+
+                crop_logits = crop_logits[0]
+
+                preds += F.pad(crop_logits, pad=(int(x1), int(preds.shape[3] - x2), int(y1), int(preds.shape[2] - y2)))
```
BloodAxe, line 435: This increment is somewhat suboptimal. Instead of accumulating `preds[roi] += crop_logits` this code padds the single tile to a full resolution and adds it. This is just ineffective.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -856,6 +856,22 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
...
+        if context.training_params.max_epochs - 1 == context.epoch:
+            unwrap_model(context.net).enable_swi()
+            context.valid_loader.dataset.transforms.transforms = []
```
BloodAxe, line 868: I feel that emptying transforms list in this way is VERY risky move. That may be solving immediate goal of adding SWI but if someone who is less familiar with implementation details runs into troubles because of this change - the odds they would curse us are pretty high.
Maybe even now we may have some incompatible scenarios where this would break (Like quantization). I really feel uneasy about this line. Is there really not other way of solving this?

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -315,6 +315,8 @@ def __init__(
         overlap_patch_stride: List[int],
         overlap_patch_pad: List[int],
         in_channels: int = 3,
+        sliding_window_size: Tuple[int, int] = (1024, 1024),
+        sliding_window_stride: Tuple[int, int] = (768, 768),
```
BloodAxe, line 319: 1) If new parameters is added a matching docstring should be also added.
2) (rows, cols) or (width, height) ? That should be reflected in docstring.

src/super_gradients/training/utils/callbacks/callbacks.py
BloodAxe: A callback should check whether model supports this at the start of training and raise an exception if model is not compatible.

src/super_gradients/training/utils/callbacks/callbacks.py
Yael-Baron: I agree. It was a quick fix due to time frame concerns.
@shaydeci how should I proceed?

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -347,6 +349,12 @@ def __init__(
 
         self.init_params()
 
+        self.num_classes = num_classes
+
+        self.is_SWI = False
```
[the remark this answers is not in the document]
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/utils/callbacks/callbacks.py
Yael-Baron: Currently it was implemented for segformer model only.
Fixed on c79d0f2.

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: @shaydeci how should I proceed?

src/super_gradients/common/object_names.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/utils/callbacks/callbacks.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/utils/callbacks/callbacks.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on c79d0f2

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on c79d0f2

Yael-Baron:
Pushed a new commit (no. 1b8a739) with new recipes for each segformer's variant based on a default recipe.

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: Missed this one, completely agree with @BloodAxe here.
How about mentioning explicitly the new transforms we switch to when switching to SWI? Is it always going to be the case that we just empty the transforms ?
It does not make sense to me that transforms that might have nothing to do with resolution should be dropped...

Any way, we should set the transforms taken out after the validation is completed.
I think an appropriate warning/message should be printed as well.

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -326,12 +328,17 @@ def __init__(
         :param overlap_patch_stride:  the patch stride of the overlapping patch embedding in each stage
         :param overlap_patch_pad:  the patch padding of the overlapping patch embedding in each stage
         :param in_channels:  number of input channels
+        :param sliding_window_crop_size:  (width, height) the crop size to take from the image for forward with sliding window
+        :param sliding_window_stride:  (rows, cols) the stride size between crops for forward with sliding window
```
BloodAxe, line 332: I strongly suggest not to mix parameters taking (rows/cols) and (cols/rows) in a single function.
No double, it's helpful to have a docstring, but why we use (rows/cols) in one parameter and (cols/rows) in a other one?

src/super_gradients/training/models/segmentation_models/segformer.py
```diff
@@ -347,6 +354,12 @@ def __init__(
 
         self.init_params()
 
+        self.num_classes = num_classes
+
+        self.use_sliding_window_validation = False
+        self.sliding_window_crop_size = sliding_window_crop_size
```
BloodAxe, line 360: Please convert them to tuples before saving in self, as they my have a type of ListConfig, which is not desirable to be a part of the model: `self.sliding_window_crop_size = tuple(sliding_window_crop_size)`

src/super_gradients/training/utils/callbacks/callbacks.py
BloodAxe: Yes, I know it is intended only for segformer. However, if someone (a random user) switch the model from segformer to something else and forget to remove this callback from the config, he would get a quite uninformative AttributeError at the call `unwrap_model(context.net).enable_sliding_window_validation()`. It may be obvious to us as code owners what is the problem but not for first-time users.

What I'm suggesting to change here is the following:
```
model =unwrap_model(context.net)
if not hasattr(model, "enable_sliding_window_validation") or not hasattr(model, "disable_sliding_window_validation"):
   raise ConfigurationError("You can sliding windown validation callback, but your model does not support sliding window inference. Please either remove the callback or use the model that supports sliding inference: Segformer, XXX, YYY")
```

That's it. The goal is to just show some love to users and help them understand their mistakes.

src/super_gradients/training/models/segmentation_models/segformer.py
BloodAxe: I'm not Shay, but my 5 cents on how to start:

1) Extract `sliding_window_inference(model, input, tile_size, tile_step, ...)` method that would do just the sliding inference of user-provided model. That would also enable us to use this method in other models.

2) Add unit tests to check the correctness of the function for following cases:

* Input size is smaller than tile_size: `input_size=(512,512`, `tile_size=(640,640)`
* Input size is equal to tile size
* Input size is greater than tile size `input_size=(513,513)`, `tile_size=(512,512)`
* For each of these cases you may also want to check: 
* tile_step is equal to tile_size
* tile_step is larger than tile size (This should raise an Exception)
* tile_step is less than tile_size (tile_size=512, tile_step=384)
* Also check that function works correctly with odd-sized inputs (input_size=13, tile_step=3, tile_size=2)

3) How to check correctness
Use dummy model that returns an input. So the `sliding_window_inference` should then reconstruct the same input (torch.randn(input_size))

src/super_gradients/training/utils/callbacks/callbacks.py
BloodAxe: If we drop transforms to keep spatial size, how do we handle normalization transforms then?

src/super_gradients/training/utils/callbacks/callbacks.py
Yael-Baron: Fixed on f400bf3

src/super_gradients/training/models/segmentation_models/segformer.py
shaydeci: I think @BloodAxe covered it pretty much here ^

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on 2821ee.

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on 4e91c8d.

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: I am not 100 percent sure, but think we need to call __iter__ otherwise this won't be updated. Have you checked that images actually pass through this empty pipeline ?

shaydeci: This also makes sense to me as:

```
    def _transform_image_and_mask(self, image, mask) -> tuple:
        """
        :param image:           The input image
        :param mask:            The input mask
        :return:                The transformed image, mask
        """
        transformed = self.transforms({"image": image, "mask": mask})
```

This would'nt have worked...

Yael-Baron: Fixed on 0ec8a9c

src/super_gradients/training/models/segmentation_models/segformer.py
shaydeci: in tests/unit_tests as the other unit tests.
Dont forget to add it to the test suite.

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on 8a178fa

src/super_gradients/training/utils/callbacks/callbacks.py
Louis-Dupont:
``` python
        model = unwrap_model(context.net)
        if not hasattr(model, "enable_sliding_window_validation") or not hasattr(model, "disable_sliding_window_validation"):
            raise ValueError(
                "You can use sliding window validation callback, but your model does not support sliding window "
                "inference. Please either remove the callback or use the model that supports sliding inference: "
                "Segformer"
            )
```
This will be raised even when we dont want to run sliding window. I think this is why your tests failed
Am I right ?

Yael-Baron: Correct. Fixed on ec450ec87e6799ceeb1bc6d25bf3ec75f200a990.

Yael-Baron: Fixed on e2595b598dcb7ace383ba75d31466c470d0e34e7

src/super_gradients/training/models/segmentation_models/segformer.py
Yael-Baron: Fixed on e9a6e9f725e8730efe9130f78b2aa604ed253732, 61ea395063f51661a3f6c8e26e0ce050bc73ffd9, 1475d135331fb8ef7ade9723b4e1a95313ada41a and 0ec8a9cf3c0ae469c9b02018cfa477177caaedb3.

Yael-Baron:
Pushed integration test for segformer's models on commit db42e6c7b94a7e1e193d22f3026fb6f7d8eca015

tests/unit_tests/pretrained_models_unit_test.py
```diff
@@ -29,6 +29,36 @@ def test_pretrained_repvgg_a0_imagenet(self):
         model = models.get(Models.REPVGG_A0, pretrained_weights="imagenet", arch_params={"build_residual_branches": True})
         trainer.test(model=model, test_loader=classification_test_dataloader(), test_metrics_list=[Accuracy()], metrics_progress_verbose=True)
 
+    def test_pretrained_segformer_b0_cityscapes(self):
```
shaydeci, line 32: These tests are a bit meaningless. You can take them out.

shaydeci (changes requested):

shaydeci (approved):
LGTM on my end.

src/super_gradients/training/utils/segmentation_utils.py
```diff
@@ -43,6 +43,10 @@ def to_one_hot(target: torch.Tensor, num_classes: int, ignore_index: int = None)
     :param num_classes: num of classes in datasets excluding ignore label, this is the output channels of the one hot
         result.
     :return: one hot tensor with shape [N, num_classes, H, W]
+
+    Parameters
```
BloodAxe, line 47: Why 
```
Parameters
-----------
```

and not :param ignore index?

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -863,6 +863,52 @@ def _infer_global_step(self, context: PhaseContext, is_train_loader: bool):
...
+    def on_test_loader_start(self, context: PhaseContext) -> None:
+        unwrap_model(context.net).enable_sliding_window_validation()
+        self.test_loader_transforms = context.test_loader.dataset.transforms.transforms
+        context.test_loader.dataset.transforms.transforms = self.transforms_for_sliding_window
+        iter(context.test_loader)
+
+    def on_test_loader_end(self, context: PhaseContext) -> None:
+        unwrap_model(context.net).disable_sliding_window_validation()
+        context.test_loader.dataset.transforms.transforms = self.test_loader_transforms
+        iter(context.test_loader)
```
BloodAxe, line 909: I'm not sure I'm following why we need `iter(context.test_loader)` here.

Yael-Baron:
A signed PR was opened, please follow Segformer swi signed #1361 PR.

PR #1298, opened 2023-07-19
Addiing more logs to let user know when pretrained_weights is being used, and/or downloaded

Louis-Dupont:
The request came from @eran-deci and aims at letting the user know when we are properly loading pretrained weights and whether or not it was downloaded from the platfrom.


shaydeci (approved):
LGTM

PR #1317, opened 2023-07-24
Fix datasetparams not showing

Louis-Dupont:
### Background
When someone defines the dataset name from `dataloader_params`, the parameters are not properly logged

```yaml
val_dataloader_params:
  dataset: Cifar10
  batch_size: 512
  num_workers: 8
  drop_last: False
  pin_memory: True
```

image


BloodAxe (approved):
LGTM

PR #1318, opened 2023-07-24
New Export API

BloodAxe:
- Supports only (some) detection models so far

Supports:
* Export to onnxruntime & tensorrt
* End-to-End export with NMS
* Quantization with calibration

Still remaining to implement:
* Remaining detection models
* Inference of the preprocessing (normalization) params from the model itself
* Docs & Tutorials


src/super_gradients/module_interfaces/exportable_detector.py
```diff
@@ -0,0 +1,396 @@
...
+        an instance of AbstractObjectDetectionDecodingModule that would take raw models' outputs and
+        convert them to a tuple of two tensors (boxes, scores):
+         - boxes: [B, N, 4] - All predicted boxes in (x1, y1, x2, y2) format.
+         - scores: [B, N, C] - All predicted scores ([0..1] range) for each box and class.
+        :return: An instance of AbstractObjectDetectionDecodingModule
+        """
+        raise NotImplementedError()
+
+    def get_preprocessing_callback(self, **kwargs) -> Optional[nn.Module]:
+        raise NotImplementedError(f"get_preprocessing_callback is not implemented for class {self.__class__.__name__}.")
```
Louis-Dupont, line 76: Is there a reason why we go for `raise NotImplementedError()` instead of making it abstract ? To be more precise;

- Do we want to support instantiation of base `ExportableObjectDetectionModel` class ? 
- Is there any situation where we may want to inherit from `ExportableObjectDetectionModel` without implementing `get_decoding_module`  or `get_preprocessing_callback` ?

src/super_gradients/module_interfaces/module_interfaces.py
```diff
@@ -17,8 +19,7 @@ def get_dataset_preprocessing_params(self):
         ...
 
 
-@runtime_checkable
-class HasPredict(Protocol):
+class HasPredict:
```
Louis-Dupont, line 22: What is the issue with `Protocol` ?
Is it because it creates an implicit/hidden contract between classes that implement predict and `HasPredict` ?

Louis-Dupont: But why not making it abstract ? I think that if a class inherits from `HasPredict` without implementing the required classes this should raise an exception on instantiation, not on call.

Louis-Dupont (commented):
Just left some comments, I'll review more in-depth once it's open

Makefile
```diff
@@ -14,3 +14,7 @@ recipe_accuracy_tests:
 	python src/super_gradients/train_from_recipe.py --config-name=cityscapes_regseg48          experiment_name=shortened_cityscapes_regseg48_iou_test   epochs=10 training_hyperparams.average_best_models=False multi_gpu=DDP num_gpus=4
 	python src/super_gradients/examples/convert_recipe_example/convert_recipe_example.py --config-name=cifar10_conversion_params experiment_name=shortened_cifar10_resnet_accuracy_test
 	coverage run --source=super_gradients -m unittest tests/deci_core_recipe_test_suite_runner.py
+
+
+examples_to_docs:
+	jupyter nbconvert --to markdown --output-dir="documentation/source/" --execute src/super_gradients/examples/model_export/models_export.ipynb
```
Louis-Dupont, line 20: That's nice!

documentation/source/models_export.md
```diff
@@ -0,0 +1,900 @@
...
+                    x1=int(x1),
+                    y1=int(y1),
+                    x2=int(x2),
+                    y2=int(y2),
+                    class_id=class_index,
+                    class_names=class_names,
+                    color_mapping=color_mapping,
+                    box_thickness=2,
+                    pred_conf=class_score,
+                )
```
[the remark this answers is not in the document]
Louis-Dupont: Here the example is a function that plots bbox, while for `Flat format` just above, we just iterate over the results. I think it would make more sense to show 2 equivalent pieces of code

Louis-Dupont (commented):
Tutorial looks great! I left some minor comments in-line

src/super_gradients/module_interfaces/module_interfaces.py
BloodAxe: The problem with Protocol is that it was messing up [MRO](https://levelup.gitconnected.com/method-resolution-order-in-python-5afbaecc25e0) so that derived methods `get_decoding_module` were not called.
Empirically I found that removing protocol solves the problem. Other methods (changing order or inheritance and playing with explicit call of c'tors for base classes) did not help

src/super_gradients/module_interfaces/exportable_detector.py
BloodAxe: > Do we want to support instantiation of base ExportableObjectDetectionModel class ?
Nope

> Is there any situation where we may want to inherit from ExportableObjectDetectionModel without implementing get_decoding_module or get_preprocessing_callback ?
As of now - no, I cannot think about such cases.

> Is there a reason why we go for raise NotImplementedError() instead of making it abstract ? To be more precise

You mean subclassing from `abc` and marking methods `abstractmethod`? Maybe, need to test whether this would work fine with inheritance chain.

documentation/source/models_export.md
```diff
@@ -0,0 +1,900 @@
...
+
+### Supported backends
+
+Currently, we support two backends for exporting models:
+
+* ONNX Runtime
+* TensorRT
+
+The only difference between these two backends is what NMS implementation will be used. ONNX Runtime uses NMS implementation from ONNX opset, while TensorRT uses its own NMS implementation which is expected to be faster.
+A disadvantage of TensorRT backend is that you cannot run than ONNX model inside ONNX Runtime. You can, however, run models exported with ONNX Runtime backend inside TensorRT.
```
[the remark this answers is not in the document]
BloodAxe: Fixed in acf2684633c3bef058288467c33b4be6b42c261c

documentation/source/models_export.md
```diff
@@ -0,0 +1,900 @@
...
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+             0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=int64)]
```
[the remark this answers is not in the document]
BloodAxe: Fixed in acf2684633c3bef058288467c33b4be6b42c261c

src/super_gradients/module_interfaces/exportable_detector.py
```diff
@@ -0,0 +1,399 @@
...
+from super_gradients.training.utils.utils import infer_model_device
+from super_gradients.training.utils.quantization.selective_quantization_utils import SelectiveQuantizer
+
+logger = get_logger(__name__)
+
+__all__ = ["ExportableObjectDetectionModel", "AbstractObjectDetectionDecodingModule", "ModelExportResult"]
+
+
+class AbstractObjectDetectionDecodingModule(nn.Module):
+    @abc.abstractmethod
```
shaydeci, line 28: Please add some docs.

src/super_gradients/module_interfaces/exportable_detector.py
```diff
@@ -0,0 +1,399 @@
...
+                    f"Please specify max_predictions_per_image <= {num_pre_nms_predictions}."
+                )
+        else:
+            attach_nms_postprocessing = False
+            postprocessing = None
+
+        if hasattr(model, "prep_model_for_conversion"):
+            model.prep_model_for_conversion(**prep_model_for_conversion_kwargs)
+
+        if quantization_mode == ExportQuantizationMode.INT8:
```
shaydeci, line 267: Let's check if the model has already quantised layers inside it (i.e QuantConv2d etc), and if so skip this part.
Then we can call export right after Trainer's quantization methods.

shaydeci (changes requested):
Amazing work @BloodAxe !

I added some comments we discussed in our convo.
Let's also try and detail what the label format in each stage is/expected as much as possible.

src/super_gradients/module_interfaces/exportable_detector.py
BloodAxe: Fixed in e2d4fece90638c286336fcacf162bb86805da476

src/super_gradients/module_interfaces/exportable_detector.py
```diff
@@ -0,0 +1,399 @@
...
+        calibration_loader: Optional[DataLoader] = None,
+        preprocessing: Union[bool, nn.Module] = True,
+        postprocessing: Union[bool, nn.Module] = True,
+        postprocessing_kwargs: Optional[dict] = None,
+        batch_size: int = 1,
+        input_image_shape: Optional[Tuple[int, int]] = None,
+        input_image_channels: Optional[int] = None,
+        input_image_dtype: Optional[torch.dtype] = None,
+        max_predictions_per_image: Optional[int] = None,
+        onnx_export_kwargs: Optional[dict] = None,
```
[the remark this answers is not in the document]
BloodAxe: Fixed in e2d4fece90638c286336fcacf162bb86805da476

src/super_gradients/module_interfaces/exportable_detector.py
BloodAxe: Fixed in 57890f38b159f9fae4d745c5cf97b90937d16311

tests/unit_tests/export_detection_model_test.py
```diff
@@ -0,0 +1,713 @@
...
+                engine=ExportTargetBackend.TENSORRT,
+                max_predictions_per_image=300,
+                input_image_shape=(640, 640),
+                output_predictions_format=DetectionOutputFormatMode.BATCH_FORMAT,
+                quantization_mode=ExportQuantizationMode.INT8,
+                calibration_loader=dummy_calibration_loader,
+            )
+            assert export_result is not None
+
+    def test_export_yolonas_quantized_with_calibration_to_tensorrt(self):
```
shaydeci, line 370: Is this supposed to work on a machine without gpu ?
Take in account that the unit tests suite does not run on spots....

BloodAxe: Yep, it runs fine - exporting to TRT does not involve TRT itself. We still export through ONNX

tests/unit_tests/export_detection_model_test.py
```diff
@@ -0,0 +1,781 @@
...
+from super_gradients.training import models
+from super_gradients.training.dataloaders import coco2017_val  # noqa
+from super_gradients.training.datasets.datasets_conf import COCO_DETECTION_CLASSES_LIST
+from super_gradients.training.utils.detection_utils import DetectionVisualization
+from super_gradients.training.utils.export_utils import infer_image_shape_from_model, infer_image_input_channels
+from super_gradients.training.utils.media.image import load_image
+import onnx_graphsurgeon as gs
+
+
+class TestDetectionModelExport(unittest.TestCase):
```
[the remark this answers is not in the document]
BloodAxe: Fixed, these tests are part of unit testing suite

PR #1327, opened 2023-07-27
Fix class_inclusion_list in DetectionDataset

Louis-Dupont:
### Motivation
There is currently a bug where subclassing does not work when the input `target_format` is ConcatenatedTensorFormat.
The root cause of this is that the tests were not updated when `ConcatenatedTensorFormat` was introduced to this dataset, so the code still pass with old deprecated format, but we did not detect that it doesnt work with the new format.


BloodAxe:
Can you please leave a few words what this fix fixes and why it is done the way it is done?

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -232,9 +232,9 @@ def _sub_class_annotation(self, annotation: dict) -> Union[dict, None]:
         :param annotation: Dict representing the annotation of a specific image
         :return:           Subclassed annotation if non empty after subclassing, otherwise None
         """
-        cls_posx = get_cls_posx_in_target(self.original_target_format)
+        cls_index = _get_cls_index_in_target(target_format=self.original_target_format)
```
BloodAxe, line 235: Lt's nt cmprsd wrds in vrbl nd fnctn nms pls

Louis-Dupont: [8f5324f](https://github.com/Deci-AI/super-gradients/pull/1327/commits/8f5324f6ae7f39f1836f5cef7630f5c8cf80c417)

BloodAxe (approved):
LGTM

PR #1350, opened 2023-08-07
Feature/sg 1059 fix ci

BloodAxe:
# What this PR does

Ensures that we test againts a checkout copy of SG. 
It first uninstall the (maybe) existing SG installation and do a development-install a checkouted copy.
This PR also adds an option to have `--extra-index-url` in requirements.txt (Would be necessary for export API PR)
Since we install the SG we also remove the unnecessary `export PYTHONPATH=/home/circleci/super_gradients`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,7 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             python3 -m pip install pip==23.1.2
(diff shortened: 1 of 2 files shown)


Louis-Dupont (commented):
LGTM

shaydeci (approved):
LGTM

PR #1353, opened 2023-08-08
Feature/sg 1027 checkpoint directory refacto

Louis-Dupont:
### Motivation
Being able to run new training in same experiment without overriding latest run...

### Implemented Logic
By default: `Trainer.train()` creates a unique run directory every time it is called (based on call timestamp). 
⚠️ It is NOT one run directory per `Trainer` instance (on purpose!)

- Calling `Trainer.train()` twice will create one one directory per run -> 2 directories in total
- Calling `Trainer.train()` and then  `Trainer.train()`  with `resume=True` will work like before, i.e. resume latest run and overwrite the old checkpoint with new one - all in the same run directory -> 1 directory in total
- Calling `Trainer.train()` and then  `Trainer.train()`  with `run_id=...` will work like resume but on a specific run, i.e. resume that specific run and overwrite the old checkpoint with new ones - all in the same run directory -> 1 directory in total
- Calling `Trainer.train()` and then  `Trainer.train()`  with `resume_path=...` will work like before, i.e. load the checkpoint based on `resume_path` (full path to `.pth` file), create a new run directory and then save the checkpoints in this directory. -> 2 directories in total

Or simplified:
- `resume=True` will `CONTINUE` latest run
- `run_id=...` will `CONTINUE` a specific run
- `resume_path=...` will `BRANCH` a specific checkpoint and create a new run out of it

### Backward compatibility
1. **What I did**

I added backward compatibility by supporting `resume` when no **run directory** is found. In this case, it will assume that the  run directory is actually the experiment directory (what we did until now). This only affect resuming on old experiment directories and will just work like before.

3. **What is not fully covered**

(Kind of edge case)
Previously, the checkpoint directory only depended on the `ckpt_root_dir` and `experiment_dir`, so on instantiation the `Trainer.checkpoints_dir_path` would be set. Now it depends on the "run", i.e. it is set on `trainer.train()` call.  This can theoretically create issue if someone
- execute `trainer.checkpoints_dir_path` BEFORE `trainer.train` (it will just return `None`, before it used to give the path to the experiment directory)
- assumes that trainer.checkpoints_dir_path remaines the same between the two calls of `trainer.train()`

I dont see a clean solution here, and eventually it seems pretty edge case (and I don't see a real motivation for people to do that) + the test successfully passed. But I think its worth mentioning.

**Extra note**
We have lots of attributes in the SgTrainer that depend on the run. I know we talk a lot about it and don't do much, but if/when we refacto the Trainer, I think we should add the concept of Run to avoid storied variables that depend on the run inside the trainer.


BloodAxe:
Great change!

If I may, there one thing I'd like to be explicitly tested - hydra sweep syntax:

```python -m super_gradients.train_from_recipe -m --config-name=segformer_b2 training_hyperparams.initial_lr=0.001,0.01,0.0001 training_hyperparams.ema=True,False`

**Comment by Louis-Dupont** [comment]
> Great change!
> 
> If I may, there one thing I'd like to be explicitly tested - hydra sweep syntax:
> 
> ```python
> python -m super_gradients.train_from_recipe -m --config-name=segformer_b2 training_hyperparams.initial_lr=0.001,0.01,0.0001 training_hyperparams.ema=True,False
> ```

Good idea, I'll add that

**Comment by Louis-Dupont** [comment]
@BloodAxe 
```
python -m super_gradients.train_from_recipe -m --config-name=cifar10_resnet experiment_name=TEST_SWEEP_cifar10 training_hyperparams.initial_lr=0.001,0.01,0.03,0.0001 training_hyperparams.ema=True,False
```
Leads to this
```
(louis-dev) louis.dupont@deci-tzag8-desktop:~/PycharmProjects/super-gradients/checkpoints/TEST_SWEEP_cifar10$ ls -al
total 40
drwxrwxr-x 10 louis.dupont louis.dupont 4096 Aug  9 13:19 .
drwxr-xr-x 75 louis.dupont louis.dupont 4096 Aug  9 13:15 ..
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:15 RUN_20230809_131515_515021
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:16 RUN_20230809_131555_908275
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:17 RUN_20230809_131633_522811
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:17 RUN_20230809_131711_213721
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:18 RUN_20230809_131748_323576
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:19 RUN_20230809_131826_842790
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:19 RUN_20230809_131904_885102
drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:20 RUN_20230809_131943_476859
```

```
(louis-dev) louis.dupont@deci-tzag8-desktop:~/PycharmProjects/super-gradients/checkpoints/TEST_SWEEP_cifar10$ ls RUN_20230809_131515_515021
average_model.pth  ckpt_latest.pth             events.out.tfevents.1691576116.deci-tzag8-desktop.105548.0  experiment_logs_Aug09_13_15_16.txt
ckpt_best.pth      console_Aug09_13_15_16.txt  events.out.tfevents.1691576154.deci-tzag8-desktop.105548.1  logs_Aug09_13_15_16.txt

```

It works all great. 
To tests, I only see two options:
1. launching sweeper with subprocess, but I dont think we should go for that
2. launching in the CI directly, but I generally prefer to avoid that because I think tests should belong to test folder.
I couldnt find a better way, do you have an idean of a clean way to do that?

File: src/super_gradients/training/kd_trainer/kd_trainer.py [hunk 1]
```diff
@@ -211,6 +211,7 @@ def _load_checkpoint_to_model(self):
         Initializes teacher weights with teacher_checkpoint_path if needed, then handles checkpoint loading for
          the entire KD network following the same logic as in Trainer.
         """
+        super(KDTrainer, self)._load_checkpoint_to_model()
```
shaydeci, line 214: Is there a specific reason you move the call up ?

tests/unit_tests/resume_training_test.py
```diff
@@ -65,6 +66,62 @@ def test_resume_training(self):
...
+        net = LeNet()
+        trainer = Trainer(ckpt_root_dir=ckpt_root_dir, experiment_name=experiment_name)
+        trainer.train(model=net, training_params=train_params, train_loader=classification_test_dataloader(), valid_loader=classification_test_dataloader())
+
+        # Check directory size
+        self.assertEqual(original_dir_count + 1, len(os.listdir(experiment_dir)), "You should have 1 run folder created only after calling `Trainer.train`.")
+
+        # TRAIN FOR 1 MORE EPOCH AND COMPARE THE NET AT THE BEGINNING OF EPOCH 3 AND THE END OF EPOCH NUMBER 2
+        first_epoch_cb = FirstEpochInfoCollector()
+        train_params["run_id"] = get_latest_run_id(checkpoints_root_dir=ckpt_root_dir, experiment_name=experiment_name)
```
shaydeci, line 103: Maybe also add a test that resumes from a run id that is not the latest one?
This would get us more coverage.

src/super_gradients/recipes/training_hyperparams/default_train_params.yaml
```diff
@@ -1,5 +1,6 @@
-resume: False # whether to continue training from ckpt with the same experiment name.
-resume_path: # Explicit checkpoint path (.pth file) to use to resume training.
+resume: False # Whether to continue training from ckpt from the latest run, within the same experiment name.
+run_id:       # ID of run to resume from the same experiment.
```
shaydeci, line 2: Please also add this parameter documentation in the Trainer.train(..) doctoring.
Also, I think there are multiple .md documentation files that need to know about this feature.

shaydeci (changes requested):
Very nice work!
I have some minor comments inline.
Besides them, I have 2 more:
1. I think we should also print the run_id at the start of the training.
2. We have a few documentation .md files that should be updated and should detail how to work with the run_id feature.

src/super_gradients/training/kd_trainer/kd_trainer.py
Louis-Dupont: The base `_load_checkpoint_to_model` loads checkpoints, but also sets some checkpoints attributes which are used by `KDTrainer._load_checkpoint_to_model`.

src/super_gradients/recipes/training_hyperparams/default_train_params.yaml
Louis-Dupont: .md doc added ced84409863d301a0b70f0454377e9bc612b2696
Trainer doc added https://github.com/Deci-AI/super-gradients/pull/1353/commits/e46007d5140e85f7abd98db89e1e8d50946776e7

shaydeci:
> @BloodAxe
> 
> ```
> python -m super_gradients.train_from_recipe -m --config-name=cifar10_resnet experiment_name=TEST_SWEEP_cifar10 training_hyperparams.initial_lr=0.001,0.01,0.03,0.0001 training_hyperparams.ema=True,False
> ```
> 
> Leads to this
> 
> ```
> (louis-dev) louis.dupont@deci-tzag8-desktop:~/PycharmProjects/super-gradients/checkpoints/TEST_SWEEP_cifar10$ ls -al
> total 40
> drwxrwxr-x 10 louis.dupont louis.dupont 4096 Aug  9 13:19 .
> drwxr-xr-x 75 louis.dupont louis.dupont 4096 Aug  9 13:15 ..
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:15 RUN_20230809_131515_515021
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:16 RUN_20230809_131555_908275
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:17 RUN_20230809_131633_522811
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:17 RUN_20230809_131711_213721
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:18 RUN_20230809_131748_323576
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:19 RUN_20230809_131826_842790
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:19 RUN_20230809_131904_885102
> drwxrwxr-x  2 louis.dupont louis.dupont 4096 Aug  9 13:20 RUN_20230809_131943_476859
> ```
> 
>  ```
> (louis-dev) louis.dupont@deci-tzag8-desktop:~/PycharmProjects/super-gradients/checkpoints/TEST_SWEEP_cifar10$ ls RUN_20230809_131515_515021
> average_model.pth  ckpt_latest.pth             events.out.tfevents.1691576116.deci-tzag8-desktop.105548.0  experiment_logs_Aug09_13_15_16.txt
> ckpt_best.pth      console_Aug09_13_15_16.txt  events.out.tfevents.1691576154.deci-tzag8-desktop.105548.1  logs_Aug09_13_15_16.txt
> ```
> 
> It works all great. To tests, I only see two options:
> 
> 1. launching sweeper with subprocess, but I dont think we should go for that
> 2. launching in the CI directly, but I generally prefer to avoid that because I think tests should belong to test folder.
>    I couldnt find a better way, do you have an idean of a clean way to do that?

We can add it for now as a part of one of the recipe tests in the CI.
Nice work, after we add that it LGTM on my end.

.circleci/config.yml
```diff
@@ -448,6 +447,72 @@ jobs:
...
+              echo "Error: Directory $experiment_dir was not created."
+              exit 1
+            fi
+
+            # Step 4: Make sure that experiment_dir includes $expected_num_dir subdirectories. If not, fail
+            subdir_count=$(find "$experiment_dir" -mindepth 1 -maxdepth 1 -type d | wc -l)
+            if [ "$subdir_count" -ne  $expected_num_dir ]; then
+              echo "Error: $experiment_dir should include $expected_num_dir subdirectories but includes $subdir_count."
+              exit 1
+            fi
```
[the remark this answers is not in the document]
Louis-Dupont: Actually, why don't we:
- Write all tests in the `Makefile` - currently partially done
- If the script is too big, save it to `/scripts` folder
- Call `Make` command in the CI to make sure that the test are the same

WDYT?

BloodAxe:
@Louis-Dupont regarding sweep test - I think it is enough to just test and ensure it is working as expected and call it a day. IMO there is no big need in dedicated test for that.

.circleci/config.yml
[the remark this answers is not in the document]
Louis-Dupont: @BloodAxe I moved the test to Makefile and simplified it

2e31ae402351d12da75c2de1bc3b0ef65e236351
(This commit was dirty, I cleaned it in the next commit)

BloodAxe (approved):
LGTM

Louis-Dupont:
This PR was merged from another branch https://github.com/Deci-AI/super-gradients/pull/1401

PR #1354, opened 2023-08-08
Summarize models, losses & metrics for segmentation

BloodAxe:


Louis-Dupont (dismissed):
LGTM, but I think ideally it would be great to get a quick feedback from people it is meant to (DLE)

eran-deci (commented):
Link to pretrained weights (line 132) is broken

PR #1365, opened 2023-08-10
Feature/sg 1047 predict od with labels

shaydeci:
Screen Shot 2023-08-13 at 19 01 02

This PR also contains a small fix for prep_mode_for_conversion of PPYOLOE (extraction of h,w at some point where some references pass the entire image shape, while others only the last 2 channels, made the vuild fail with the new tests for predict).


src/super_gradients/examples/predict/detection_predict_with_labels.py
```diff
@@ -0,0 +1,26 @@
+from super_gradients.common.object_names import Models
+from super_gradients.training import models
+from pathlib import Path
+
+# Note that currently only YoloX, PPYoloE and YOLO-NAS are supported.
```
ofrimasad, line 5: this comment should be in line 5 (before the model instantiation)

src/super_gradients/examples/predict/detection_predict_with_labels.py
```diff
@@ -0,0 +1,26 @@
...
+
+model = models.get(Models.YOLO_NAS_L, pretrained_weights="coco")
+mini_coco_data_dir = str(Path(__file__).parent.parent.parent.parent.parent / "tests" / "data" / "tinycoco")
+
+dataset = COCODetectionDataset(
+    data_dir=mini_coco_data_dir, subdir="images/val2017", json_file="instances_val2017.json", input_dim=None, transforms=[], cache_annotations=False
+)
+
+# x's are np.ndarrays images of shape (H,W,3)
+# y's are np.ndarrays of shape (num_boxes,x1,y1,x2,y2,class_id)
```
ofrimasad, line 16: This is so confusing...
what are x and y? I thought they are coordinates. 
don't call images x and labels y....

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -182,18 +183,39 @@ def predict(
         conf: Optional[float] = None,
         batch_size: int = 32,
         fuse_model: bool = True,
+        target_bboxes: Optional[List[np.ndarray]] = None,
```
ofrimasad, line 186: should be `Optional[Union[np.ndarry, List[np.ndarray]]] = None`

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -182,18 +183,39 @@ def predict(
...
         :param batch_size:  Maximum number of images to process at the same time.
+
         :param fuse_model:  If True, create a copy of the model, and fuse some of its layers to increase performance. This increases memory usage.
+
+        :param target_bboxes: Optional[List[np.ndarray]], ground truth bounding boxes. Can either be an np.ndarray of shape
+         (image_i_object_count, 4) when predicting a single image, or a list of length len(target_bboxes), containing such arrays.
+         When not None, will plot the predictions and the ground truth bounding boxes side by side (i.e 2 images stitched as one).
+
+        :param target_class_ids: Optional[List[np.ndarray]], ground truth target class indices. Can either be an np.ndarray of shape
+         (image_i_target_count) when predicting a single image, or a list of length len(target_bboxes), containing such arrays (default=None).
```
ofrimasad, line 207: shouldn't this be "list of length len(images)" ?

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
```diff
@@ -165,6 +166,9 @@ def predict(
         conf: Optional[float] = None,
         batch_size: int = 32,
         fuse_model: bool = True,
+        target_bboxes: Optional[List[np.ndarray]] = None,
```
ofrimasad, line 169: same here about the type (can be a single ndarray as well)

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
```diff
@@ -174,9 +178,22 @@ def predict(
                             If None, the default value associated to the training is used.
         :param batch_size:  Maximum number of images to process at the same time.
         :param fuse_model:  If True, create a copy of the model, and fuse some of its layers to increase performance. This increases memory usage.
+
+        :param target_bboxes: Optional[List[np.ndarray]], ground truth bounding boxes. Can either be an np.ndarray of shape
+         (image_i_object_count, 4) when predicting a single image, or a list of length len(target_bboxes), containing such arrays.
+         When not None, will plot the predictions and the ground truth bounding boxes side by side (i.e 2 images stitched as one).
+
+        :param target_class_ids: Optional[List[np.ndarray]], ground truth target class indices. Can either be an np.ndarray of shape
+         (image_i_target_count) when predicting a single image, or a list of length len(target_bboxes), containing such arrays (default=None).
```
ofrimasad, line 187: same comment

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -266,31 +266,83 @@ def __init__(
...
         """Decode the model output, by applying post prediction callback. This includes NMS.
 
         :param model_output:    Direct output of the model, without any post-processing.
         :param model_input:     Model input (i.e. images after preprocessing).
+
+        :param target_bboxes: Optional[List[np.ndarray]], ground truth bounding boxes. Can either be an np.ndarray of shape
+         (image_i_object_count, 4) when predicting a single image, or a list of length len(target_bboxes), containing such arrays.
+         When not None, will plot the predictions and the ground truth bounding boxes side by side (i.e 2 images stitched as one).
+
+        :param target_class_ids: Optional[List[np.ndarray]], ground truth target class indices. Can either be an np.ndarray of shape
```
ofrimasad, line 287: wrong type here only in the documentation

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -266,31 +266,83 @@ def __init__(
...
+            or (target_class_ids is not None and (target_bboxes is None or target_bboxes_format is None))
+        ):
+            raise ValueError("target_bboxes, target_bboxes_format, and target_class_ids should either all be None or all not None.")
+
+        if isinstance(target_bboxes, np.ndarray):
+            target_bboxes = [target_bboxes]
+        if isinstance(target_class_ids, np.ndarray):
+            target_class_ids = [target_class_ids]
+
+        return target_bboxes, target_class_ids
```
ofrimasad, line 344: please check that all list are of same length

src/super_gradients/examples/predict/detection_predict_with_labels.py
```diff
@@ -0,0 +1,26 @@
...
+x1, y1, _ = dataset[0]
+x2, y2, _ = dataset[1]
+
+# images from COCODetectionDataset are RGB and images as np.ndarrays are expected to be BGR
+x2 = x2[:, :, ::-1]
+x1 = x1[:, :, ::-1]
+
+predictions = model.predict([x1, x2], target_bboxes=[y1[:, :4], y2[:, :4]], target_class_ids=[y1[:, 4], y2[:, 4]], target_bboxes_format="xyxy")
+predictions.show()
```
Louis-Dupont, line 25: Why do we add these params to `predict` instead of `show` ? Eventually, it's just a visualization adding, and does not affect the predictions.
One practical downside of adding to predict is that this has to propagate to the DetectionPipeline, and it adds complexity to many classes/functions. Another thing is that if we run a prediction, it would make sense to have option to show or not show the GT without rerunning the prediction.

Would it be complex to move the logic to the `predictions.show()` function? From what it looks, I think the code would actually me much simpler, but maybe I'm wrong

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -266,31 +266,83 @@ def __init__(
...
+        if (
+            (target_bboxes is None and target_bboxes_format is not None)
+            or (target_bboxes is not None and target_bboxes_format is None)
+            or (target_class_ids is None and (target_bboxes is not None or target_bboxes_format is not None))
+            or (target_class_ids is not None and (target_bboxes is None or target_bboxes_format is None))
+        ):
+            raise ValueError("target_bboxes, target_bboxes_format, and target_class_ids should either all be None or all not None.")
+
+        if isinstance(target_bboxes, np.ndarray):
+            target_bboxes = [target_bboxes]
```
ofrimasad, line 340: but if all 3 are None, you will get lists of Nones....

src/super_gradients/training/models/detection_models/customizable_detector.py
Louis-Dupont, line 207: I guess `image_i_object_count` and `image_i_target_count` refer to the same thing

src/super_gradients/training/utils/predict/prediction_results.py
```diff
@@ -129,6 +129,22 @@ def draw(self, box_thickness: int = 2, show_confidence: bool = True, color_mappi
                 y2=int(self.prediction.bboxes_xyxy[pred_i, 3]),
             )
 
+        if plot_targets:
+            target_image = self.image.copy()
+            for target_idx in range(len(self.prediction.target_bboxes_xyxy)):
+                class_id = int(self.prediction.target_labels[target_idx])
+                target_image = draw_bbox(
+                    image=target_image,
+                    title=f"{self.class_names[class_id]}_GT",
```
Louis-Dupont, line 138: I think it would look better without the `_GT`, and instead write below or on the image "Ground Truth".
It's just that seeing `_GT` once is enough because they are all GT and it adds noise to the image

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -266,31 +266,83 @@ def __init__(
...
         return predictions
 
+    @staticmethod
+    def _check_target_args(
+        target_bboxes: Optional[Union[np.ndarray, List[np.ndarray]]] = None,
+        target_bboxes_format: Optional[str] = None,
+        target_class_ids: Optional[Union[np.ndarray, List[np.ndarray]]] = None,
+    ):
+        if (
+            (target_bboxes is None and target_bboxes_format is not None)
```
ofrimasad, line 332: too complex...
what is the logic? target ids are optional? 
```
if target_bboxes is None and target_bboxes_format is None and target_class_ids is None:
    return None, None, None
if target_bboxes is None or target_bboxes_format is None or target_class_ids is None:
    raise ValueError(..)
```

src/super_gradients/training/utils/predict/prediction_results.py
```diff
@@ -129,6 +129,22 @@ def draw(self, box_thickness: int = 2, show_confidence: bool = True, color_mappi
...
+                    color=color_mapping[class_id],
+                    box_thickness=box_thickness * 3,
```
Louis-Dupont, line 140: Why `*3` compared to predictions ?

Louis-Dupont (commented):
Left some comments inline

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -266,31 +266,83 @@ def __init__(
...
+        :param target_bboxes_format: Optional[str], bounding box format of target_bboxes, one of ['xyxy','xywh',
+        'yxyx' 'cxcywh' 'normalized_xyxy' 'normalized_xywh', 'normalized_yxyx', 'normalized_cxcywh']. Will raise an
+        error if not None and target_bboxes is None.
+
         :return:                Predicted Bboxes.
         """
+        target_bboxes, target_class_ids = self._check_target_args(target_bboxes, target_bboxes_format, target_class_ids)
+
         post_nms_predictions = self.post_prediction_callback(model_output, device=self.device)
+        if target_bboxes is None:
```
ofrimasad, line 299: it will not be None, it will be `[None,]` because of _check_target_args returning always two lists

src/super_gradients/training/utils/predict/prediction_results.py
ofrimasad, line 140: no need to make it thicker if it is side by side

src/super_gradients/training/utils/predict/predictions.py
```diff
@@ -20,14 +20,37 @@ class DetectionPrediction(Prediction):
...
+        target_bbox_format: str,
+    ):
         """
         :param bboxes:      BBoxes in the format specified by bbox_format
         :param bbox_format: BBoxes format that can be a string ("xyxy", "cxywh", ...)
         :param confidence:  Confidence scores for each bounding box
         :param labels:      Labels for each bounding box.
         :param image_shape: Shape of the image the prediction is made on, (H, W). This is used to convert bboxes to xyxy format
+
+        :param target_bboxes: Optional[List[np.ndarray]], ground truth bounding boxes. Can either be an np.ndarray of shape
```
ofrimasad, line 44: all the types in the documentation and all shapes are wrong

ofrimasad (commented):
generally looks ok, bun many comments on static typing and documentation. 
naming in the example should be fixed.

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: There is a downside to that - saving images will not consider the targets if we go that way.
I argue that from some threshold of the image count, users will tend to use save rather then show().
Also, sometimes using show() is not an option.

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: Which is exactly where it is.

src/super_gradients/training/models/detection_models/customizable_detector.py
shaydeci: True, renamed bb076b10e7fea6a7d2f5d10e917b8c0121a040ee

src/super_gradients/training/utils/predict/prediction_results.py
shaydeci: changed 07664cee9cbeb9eef9ca6dff7113907a3292f9bb

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: 4ec8187f3b19bb9da202c961f5ed2401327f7505 changed

src/super_gradients/training/models/detection_models/customizable_detector.py
shaydeci: It's the same length, but I will change it for clarity.

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: fixed

src/super_gradients/training/pipelines/pipelines.py
shaydeci: it just means what the error says, I changed the statement to a more clear one in 16bd98c744f224f04057a99af2dfc48ee77acc96

src/super_gradients/training/pipelines/pipelines.py
shaydeci: No, because the isinstance will not hold.

src/super_gradients/training/pipelines/pipelines.py
shaydeci: not true, see previous comment.

src/super_gradients/training/utils/predict/prediction_results.py
shaydeci: dropped.

src/super_gradients/training/pipelines/pipelines.py
shaydeci: added in 16bd98c744f224f04057a99af2dfc48ee77acc96

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
shaydeci: bb076b10e7fea6a7d2f5d10e917b8c0121a040ee

src/super_gradients/training/models/detection_models/customizable_detector.py
shaydeci: fixed 0798978413fab364fd7d89ea23baeab266c80e77

src/super_gradients/training/models/detection_models/pp_yolo_e/pp_yolo_e.py
shaydeci: 0798978413fab364fd7d89ea23baeab266c80e77

src/super_gradients/training/pipelines/pipelines.py
shaydeci: 9cdd670b08fe05f2e66e10aa8fc9605374e4ab5a fixed

src/super_gradients/training/utils/predict/predictions.py
shaydeci: 0798978413fab364fd7d89ea23baeab266c80e77 fixed

src/super_gradients/training/utils/predict/prediction_results.py
shaydeci: Done in latest commit, see updated image in description.

src/super_gradients/examples/predict/detection_predict_with_labels.py
```diff
@@ -0,0 +1,27 @@
+from super_gradients.common.object_names import Models
```
BloodAxe, line 1: Do you want to convert this script to ipynb fine and generate a documenation from it automatically similary to what I did for model export?

src/super_gradients/examples/predict/detection_predict_with_labels.py
BloodAxe: Overall the PR looks great to me.

I agree with @Louis-Dupont - to me the advantage of putting `targets` to predict method signature is ill-justified decision. It violates SRP, makes signature of the method more complicated, I frankly I see no real advantage of solving it this way.

To address your counter-arguments:

> There is a downside to that - saving images will not consider the targets if we go that way.
True, unless you add `targets` to `save(...)` method.

In conclusion, I feel that having `def save(..., target_boxes, target_labels` and `def show(..., target_boxes, target_labels)` would be much cleaner API design.

Under-the-hood implementation looks already LGTM, my only concert is user-facing API.

src/super_gradients/examples/predict/detection_predict_with_labels.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Note that currently only YoloX, PPYoloE and YOLO-NAS are supported.
+model = models.get(Models.YOLO_NAS_L, pretrained_weights="coco")
+mini_coco_data_dir = str(Path(__file__).parent.parent.parent.parent.parent / "tests" / "data" / "tinycoco")
+
+dataset = COCODetectionDataset(
+    data_dir=mini_coco_data_dir, subdir="images/val2017", json_file="instances_val2017.json", input_dim=None, transforms=[], cache_annotations=False
+)
+
+# x's are np.ndarrays images of shape (H,W,3)
```
ofrimasad, line 14: drop the comments please

ofrimasad (commented):
LGTM (just remove some comments in the example)

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: Alright you guys win (:
de413502d01b617481b509f27da06fe092ec1fa1

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: updated in de413502d01b617481b509f27da06fe092ec1fa1

src/super_gradients/examples/predict/detection_predict_with_labels.py
shaydeci: sure, how does one do that ?

shaydeci: Actually thinking it over again, perhaps what in here is a bit dull for a notebook.
Maybe we should unite our predict examples to one notebook, WDYT ?

BloodAxe: I think having a single predict notebook would be a great way to show how one can utilize predict() for different tasks.
What you want to have to achieve this is a notebook file somewhere in `super_gradients/examples` and a modified Makefile target:

```Makefile
examples_to_docs:
	jupyter nbconvert --to markdown --output-dir="documentation/source/" --execute src/super_gradients/examples/model_export/models_export.ipynb
	jupyter nbconvert --to markdown --output-dir="documentation/source/" --execute src/super_gradients/examples/using_models_predict/using_models_predict.ipynb
```

This way each time one executes `examples_to_docs` target a documentation is generated from the output of the notebook.

BloodAxe (approved):
LGTM

PR #1367, opened 2023-08-10
fix

Louis-Dupont:
- Fix display
- Rename `labels` -> `label` before it is used by people! I think it is important because, well, it's single class classification and i really got myself confused thinking it may be multiclass)

### Before
image
image
image

### After
image
image
image

### DIFF
--- a/src/super_gradients/training/utils/visualization/classification.py
+++ b/src/super_gradients/training/utils/visualization/classification.py
@@ -16,13 +16,16 @@ def draw_label(image: np.ndarray, label: str, confidence: float) -> np.ndarray:
     fontScale = 0.8
     thickness = 1
 
+    # Define additional spacing between the two lines
+    line_spacing = 5
+
     # Determine the size of the label and confidence text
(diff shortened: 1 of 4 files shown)


src/super_gradients/training/utils/visualization/classification.py
```diff
@@ -1,43 +1,64 @@
...
+        image,
+        label,
+        (start_x + (text_width - label_size[0]) // 2, start_y + label_size[1]),
+        cv2.FONT_HERSHEY_SIMPLEX,
+        fontScale,
+        text_color,
+        thickness,
+        lineType=cv2.LINE_AA,
+    )
+    cv2.putText(
```
BloodAxe, line 53: I would add a little bit of a spacing between two lines
They looks quite tight IMO

Louis-Dupont: Like this ? 0e9153a342ba831a4d171689660bff878411d20e
Before (Left) » After (Right)
image

BloodAxe (approved):
LGTM

PR #1373, opened 2023-08-15
Feature/sg 1057 add breaking change tests

Louis-Dupont:
# Descritption
Adding automatic test that would detect breaking change, and adding it to CI.

This PR focused on the following breaking changes:
- Import Removal (in __init__.py)
- Class Removal
- Function Removal (includes moving from file A to B)
- Function Parameter Removal
- Function Required Parameter Addition

Let me know if you think another important type of breaking change is not covered and that we should add it

---
# Breaking Change test
I created a branch to make sure the test indeed breaks the CI when a breaking change is introduced. 
image
image


src/super_gradients/common/code_analysis/breaking_change.py
```diff
@@ -0,0 +1,268 @@
...
+
+def analyze_breaking_changes(verbose: bool = 1) -> List[Dict[str, Union[str, List]]]:
+    """Analyze changes between the current branch (HEAD) and the master branch.
+    :param verbose: If True, print the summary of breaking changes in a nicely formatted way
+    :return:        List of changes, where each change is a dictionary listing each type of change for each module.
+    """
+    # GitHelper requires `git` library which should NOT be required for the other functions
+    from super_gradients.common.code_analysis.git_utils import GitHelper
+
+    root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))))
```
ofrimasad, line 223: Not very clean... 
could you use 
```
from setuptools import find_packages
find_packages(where="./src")

```
maybe

src/super_gradients/common/code_analysis/breaking_change.py
```diff
@@ -0,0 +1,268 @@
...
+    git_explorer = GitHelper(git_path=root_dir)
+
+    summary = ""
+    breaking_changes_list = []
+    for module_path in git_explorer.diff_files(source_branch="master", current_branch="HEAD"):
+
+        master_code = git_explorer.load_branch_file(branch="master", file_path=module_path)
```
ofrimasad, line 230: will this catch a file with a changed name?
i.e. if the file was 
myClass.py
and it is now 
myNewClass.py

src/super_gradients/common/code_analysis/breaking_change.py
```diff
@@ -0,0 +1,268 @@
...
+            print(colored("NO BREAKING CHANGE DETECTED!", "green"))
+
+    return breaking_changes_list
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Example script using flags")
+    parser.add_argument("--verbose", action="store_true", default=True, help="Enable verbose mode")
+    parser.add_argument("--output-file", default=None, type=str, help="Output file name")
+    parser.add_argument("--fail-on-error", action="store_true", default=True, help="Fail on error")
```
ofrimasad, line 253: i am not sure this works... 
action="store_true" is always default=False

ofrimasad: and if that is true, the default setting of this script is: do nothing. (no print, no file, no fail)

ofrimasad (commented):
This looks great!!
very nice.
minor comments and questions

.circleci/config.yml
```diff
@@ -214,6 +214,15 @@ jobs:
...
             coverage report
             coverage html  # open htmlcov/index.html in a browser
+
+      - run:
+          name: run breaking change test
+          no_output_timeout: 30m
+          command: |
+            . venv/bin/activate
+            python -m pip install gitpython==3.1.0
+            python src/super_gradients/common/code_analysis/breaking_change.py --verbose --fail-on-error
```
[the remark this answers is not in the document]
Louis-Dupont: The whole `code_analysis` module, or just the `main` function ?

src/super_gradients/common/code_analysis/breaking_change.py
Louis-Dupont: The issue is that this function can be called from anywhere, not just from the project root , so it's dangerous to work with the working directory path, which depends on where the script is called from. Using `find_packages` would not resolve it.
To make it robust to that, I chose ot work with path relatively to the file itself (__file__). 

But I agree it looks dirty.

Do you have an idea how to make it cleaner ?
We can maybe go for `root_dir = Path(__file__).resolve().parents[5]` which may be slightly better ? Not sure

src/super_gradients/common/code_analysis/breaking_change.py
Louis-Dupont: You're right, I took it out 06f5db250bc210e19ed1e1bf0ea853cba27c16a3
Default value should not matter too much anyway because we set it explicitly in the CI

src/super_gradients/common/code_analysis/breaking_change.py
Louis-Dupont: The way it is currently detected is that when a file was renamed, the old file is seen as deleted and a new one as created.
Therefore, any function/class that was in that renamed file will be considered as deleted.

It's not as clean as simply having "module renamed", but eventually we get what we want; old functions/classes are not available in the module they used to.
Do you think we should catch it explicitly with a `ModuleRenamed` case?

src/super_gradients/common/code_analysis/breaking_change.py
ofrimasad: yes. even that is better

.circleci/config.yml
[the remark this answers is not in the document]
ofrimasad: all of it. 
it is not part of our product. it is a way to TEST our product

src/super_gradients/common/code_analysis/breaking_change.py
Louis-Dupont: 0612555efc4a5b117f38d7ac8eb061bfa07e6b0e

src/super_gradients/common/code_analysis/breaking_change.py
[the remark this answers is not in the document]
Louis-Dupont: done

Louis-Dupont:
I moved the test to a new CI "check". As you can see in the screenshot, the breaking change test is separate from the regular build3.8
image

.circleci/config.yml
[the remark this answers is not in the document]
Louis-Dupont: Changed in dcd031afa5e1582beacc7cb141cf53854fd8cb71

ofrimasad (approved):
LGTM!
great work @Louis-Dupont

PR #1378, opened 2023-08-16
closing inactive issues

bit-scientist:
Feel free to modify `days-before-issue-stale` and `days-before-issue-close`


23pointsNorth:
External opinion so feel free to ignore: This is generally not looked very favorably. This repo has 50-60 issues. One can manually comment "Needs more info from author." or "Would not be acted upon right now." and close them when that's the case. If an issue is inactive for 8 days, but the error/documentation/tutorial is still not fixed, what good does it make that the issue is closed, but the problem persists?

BloodAxe:
That's a valid point. However it also often the case where users ask a question, we ask for details but users never come back (Imagine a bugreport without a minimal case to reproduce the problem, we request it but never get a reply). Mainly this bot aimed to close such dangling issues automatically.
I imagine the well-formulated issues will be forwarded through a fast-track lane and addressed in a way more timely manner and non a subject of closing by bot.
That being said, I feel current config is a bit too aggressive/fast. A 7 days for marking issue as stale and additional 14 days after that to close the issue should be a more reasonable settings to my task @bit-scientist

PR #1386, opened 2023-08-17
Bugfix: Fix loading DeciDet models that has anchors keys in state dict.

BloodAxe:
Explicitly remove the stride keys from the checkpoint if they are present which should fix the issue with DeciDet checkpoints
Since these models are also utilizing Anchors class previously, those must be deleted explicitly


Louis-Dupont (approved):
LGTM

PR #1393, opened 2023-08-20
PLFM-3938 Create a code-loading utility in DeciClient

roikoren755:


src/super_gradients/common/plugins/deci_client.py
```diff
@@ -149,43 +149,53 @@ def get_model_weights(self, model_name: str) -> Optional[str]:
         :return:            model_weights path. None if weights were not found for this specific model on this SG version."""
         return self._get_file(model_name=model_name, file_name=AutoNACFileName.WEIGHTS_PTH)
 
+    @staticmethod
+    def load_code_from_zipfile(*, file: str, target_path: str, success_message: str, package_name: str = "deci_model_code") -> None:
+        """
+        load additional code files.
+        the zip file is extracted, and code files will be placed in the target_path/package_name and imported dynamically
+        """
```
Louis-Dupont, line 157: If you can add params description (`:param <name>: <desc>`)

src/super_gradients/common/plugins/deci_client.py
```diff
@@ -149,43 +149,53 @@ def get_model_weights(self, model_name: str) -> Optional[str]:
...
-                f"you can override this by passing models.get(... download_required_code=False) and importing the files yourself"
-            )
+        self.load_code_from_zipfile(
+            file=file,
+            target_path=target_path,
+            package_name=package_name,
+            success_message=f"*** IMPORTANT ***: files required for the model {model_name} were downloaded and added to your code in:\n{package_path}\n"
+            f"These files will be downloaded to the same location each time the model is fetched from the deci-client.\n"
+            f"you can override this by passing models.get(... download_required_code=False) and importing the files yourself",
+        )
```
Louis-Dupont, line 198: I think we still want `if file is not None` because this can still happen if there is an error in the downloading of the data, and this would lead to a cryptic error

Louis-Dupont (commented):
Minor notes, but other than that looks good.
I am just wondering, what was the motivation?

roikoren755:
> Minor notes, but other than that looks good. I am just wondering, what was the motivation?

We're working on an MVP that involves instantiating SG models, possibly with additional code, but one that will come from a different location than what is currently used in the client. I want to make this refactor, to have a re-usable piece of code to be able to use the additional code in other places too.

Louis-Dupont (approved):
LGTM

PR #1394, opened 2023-08-20
Move device only when needed in predict

Louis-Dupont:


src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -68,8 +68,9 @@ def __init__(
         fuse_model: bool = True,
         dtype: Optional[torch.dtype] = None,
     ):
-        self.device = device or next(model.parameters()).device
-        self.model = model.to(self.device)
+        model_device: str = next(model.parameters()).device.type
```
BloodAxe, line 71: FYI: We now (3.2) have nice `infer_model_device(m:nn.Module)` and `infer_model_dtype(m:nn.Module)` methods that does `next(model.parameters()).device.type` and a bit more checks under the hood. So you may want to use these instead

Louis-Dupont: Good point, I changed it here 797e50685f43ad51d0c0f17b7b96d79edfc3eee8

src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -68,10 +68,13 @@ def __init__(
         fuse_model: bool = True,
         dtype: Optional[torch.dtype] = None,
     ):
-        model_device: str = infer_model_device(model=model).type
-        self.model = model.to(device) if device and device != model_device else model
-        self.device = device or model_device
+        model_device: torch.device = infer_model_device(model=model)
+        if device:
+            device: torch.device = resolve_torch_device(device=device)
```
Louis-Dupont, line 73: I thought it would be more clear to add typing `torch.device` because the type is different to the input type `str`

PR #1398, opened 2023-08-21
closing inactive issues

bit-scientist:
The following worklows does the following:

- Non-active issues are marked as stale after 8 days of no activity
- Stale issues are closed after 14 days from the time the issue is marked as stale


BloodAxe (approved):
LGTM

PR #1411, opened 2023-08-24
Hotfix/sg 000 fix yolox replace head

BloodAxe:
Fixes bug in yolox's replace head method and added test that checks that
Removed removal of checkpoints cache dir since other tests may reuse the data and not re-download them (saving $$$ on S3 transfer)

### DIFF
--- a/src/super_gradients/training/models/detection_models/yolo_base.py
+++ b/src/super_gradients/training/models/detection_models/yolo_base.py
@@ -205,7 +205,12 @@ class DetectX(nn.Module):
         self.n_anchors = 1
         self.grid = [torch.zeros(1)] * self.detection_layers_num  # init grid
 
-        self.register_buffer("stride", torch.tensor(stride), persistent=False)
+        if torch.is_tensor(stride):
+            stride = stride.clone().detach()
+        else:
+            stride = torch.tensor(stride)
+
+        self.register_buffer("stride", stride, persistent=False)
 
         self.cls_convs = nn.ModuleList()
         self.reg_convs = nn.ModuleList()
@@ -691,7 +696,7 @@ class YoloBase(SgModule, ExportableObjectDetectionModel, HasPredict):
 
             new_last_layer = DetectX(
                 num_classes=new_num_classes,
-                stride=self._head.anchors.stride,
+                stride=self.strides,
                 activation_func_type=activation_type,
                 channels=[width_mult(v) for v in (256, 512, 1024)],
                 depthwise=isinstance(old_detectx.cls_convs[0][0], GroupedConvBlock),
--- a/tests/unit_tests/replace_head_test.py
+++ b/tests/unit_tests/replace_head_test.py
@@ -1,5 +1,3 @@
-import os
-import shutil
 import unittest
 
 import torch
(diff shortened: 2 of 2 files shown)


tests/unit_tests/replace_head_test.py
```diff
@@ -37,10 +46,6 @@ def test_dekr_replace_head(self):
         self.assertEqual(heatmap.size(1), 20 + 1)
         self.assertEqual(offsets.size(1), 20 * 2)
 
-    def tearDown(self) -> None:
-        if os.path.exists("~/.cache/torch/hub/"):
-            shutil.rmtree("~/.cache/torch/hub/")
-
```
Louis-Dupont, line 43: Any reason to remove this ?

src/super_gradients/training/models/detection_models/yolo_base.py
```diff
@@ -691,7 +696,7 @@ def replace_head(self, new_num_classes=None, new_head=None):
 
             new_last_layer = DetectX(
                 num_classes=new_num_classes,
-                stride=self._head.anchors.stride,
+                stride=self.strides,
```
Louis-Dupont, line 699: Not sure if I understand, this single line was the bug right? It was just forgotten in the previous PR right?

src/super_gradients/training/models/detection_models/yolo_base.py
```diff
@@ -205,7 +205,12 @@ def __init__(
         self.n_anchors = 1
         self.grid = [torch.zeros(1)] * self.detection_layers_num  # init grid
 
-        self.register_buffer("stride", torch.tensor(stride), persistent=False)
+        if torch.is_tensor(stride):
+            stride = stride.clone().detach()
+        else:
+            stride = torch.tensor(stride)
+
+        self.register_buffer("stride", stride, persistent=False)
```
Louis-Dupont, line 213: Is this just optimization, or it also meant to fix the bug/another bug?
I think that the only noticeable difference is if stride is on GPU, `stride.clone().detach()` will create a copy on GPU, while `torch.tensor(stride)` will create a copy on CPU. Was that the motivation?

Louis-Dupont (approved):
LGTM, I just asked minor clarification

tests/unit_tests/replace_head_test.py
BloodAxe: Yes, as stated in the PR description above: 
> Removed removal of checkpoints cache dir since other tests may reuse the data and not re-download them (saving $$$ on S3 transfer)

:)

src/super_gradients/training/models/detection_models/yolo_base.py
BloodAxe: Exactly. Removed the self._head.anchors.stride, but did forget to test replace_head still works

src/super_gradients/training/models/detection_models/yolo_base.py
BloodAxe: Not exactly - this is no avoid a warning message if input is already a torch tensor.
Otherwise torch complains and suggest we use `x.clone().detach()` instead of `torch.tensor(x)`

tests/unit_tests/replace_head_test.py
Louis-Dupont: Haha indeed, sorry 😅

PR #1413, opened 2023-08-24
Export API fix: Raise meaningful exception if model has no preprocessing metadata but preprocessing=True

BloodAxe:


Louis-Dupont (approved):
LGTM

PR #1415, opened 2023-08-24
Fix DDP checkpoint

Louis-Dupont:


src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,61 @@
+import warnings
+from typing import Optional
+from functools import wraps
+
+
+def make_function_deprecated(version: str, new_function_name: Optional[str] = None, new_module_name: Optional[str] = None):
```
ofrimasad, line 6: For a long time, we have been using "Deprecated" library (which is not that good)
If we are adding our own depreciation mechanism, let's do it correctly:
1. Call the decorator @deprecate - that is the convention 
2. move this change to its ow PR
3. Params should be
  - deprecated_in_v: str   (i.e. 3.2.0)
  - removed_in_v: str        (i.e. 3.5.0)
  - target: runable               (i.e. a function) can be None
  - reason: str
4. add assertion: assert removed_in_v < __version__   (so we will remember to remove this function when releasing the right version)
5. note that the new function is not a name - it is the function itself. this is you don't need the module name. you can infer it.

src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,61 @@
...
+                    f"Please update your code to import it as follows:\n"
+                    f"  [-] from {old_func.__module__} import {old_func.__name__}\n"
+                    f"  [+] from {new_module} import {new_name}\n."
+                )
+                warnings.warn(reason, DeprecationWarning, stacklevel=2)
+                wrapper._warned = True
+
+            return old_func(*args, **kwargs)
+
+        # Each decorated object will have its own _warned state
```
ofrimasad, line 45: good idea!

src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,61 @@
...
+            if not wrapper._warned:
+                new_name = new_function_name or old_func.__name__
+                new_module = new_module_name or old_func.__module__
+                reason = (
+                    f"You are using `{old_func.__module__}.{old_func.__name__}` which is deprecated and will be removed in {version}.\n"
+                    f"Please update your code to import it as follows:\n"
+                    f"  [-] from {old_func.__module__} import {old_func.__name__}\n"
+                    f"  [+] from {new_module} import {new_name}\n."
+                )
+                warnings.warn(reason, DeprecationWarning, stacklevel=2)
```
ofrimasad, line 40: What is this warnings? why not use our logger like always?

ofrimasad (commented):
Reviewed only the deprecation part.
lets split this PR please

src/super_gradients/common/environment/ddp_utils.py
```diff
@@ -77,3 +82,159 @@ def find_free_port() -> int:
...
+        payload = torch.empty((local_size,), dtype=torch.uint8, device="cuda")
+
+    dist.broadcast(payload, 0)
+    buffer = payload.cpu().numpy().tobytes()
+    return pickle.loads(buffer)
+
+
+def all_gather(data: Any) -> List[Any]:
+    """
+    Run all_gather on arbitrary picklable data (not necessarily tensors)
```
shaydeci, line 201: Please see torch.distributed.all_gather_object.
It can simplify alot of the code here.

shaydeci (changes requested):
Please use torch.distributed.all_gather_object or torch.distributed.broadcast_object_list for gathering pickleable objects (then if you just want to broadcast it you can access the rank0 item...

src/super_gradients/common/environment/ddp_utils.py
Louis-Dupont: done [36abc20](https://github.com/Deci-AI/super-gradients/pull/1415/commits/36abc20222d1cc94e0bd50c0b503037e83af1f05)

PR #1416, opened 2023-08-24
Add deprecate module

Louis-Dupont:
Breaking change fail will be fixed in https://github.com/Deci-AI/super-gradients/pull/1420
Please review the PR above first, to fix breacking change test 🙏 

## Motivation
Have a nicer decorator to deprecate functions. The idea is to easily deprecate any function, with nice logs

### I. Usage of `warning` vs `logger.warning`
With warnings, we get
1. a nice `DeprecationWarning` built-in, which is standard in python, and which is also easy to catch (see tests)
2. Option to show the line that raised the warning `  fully_configured_deprecated_func()` in the example below. This is thanks to `stacklevel` in `warnings.warn(message, DeprecationWarning, stacklevel=2)`
```
/Users/Louis.Dupont/PycharmProjects/super-gradients/tests/unit_tests/deprecate.py:19: DeprecationWarning: Function `__main__.fully_configured_deprecated_func` is deprecated since version `3.2.0` and will be removed in version `4.0.0`.
Reason: Replaced for optimization.
Please update your code:
  [-] from `module` import `fully_configured_deprecated_func`
  [+] from `module` import `new_func`
  fully_configured_deprecated_func()
```

I think using this is cleaner than `logger.warning`, what do you think ?

### II. Making sure that `removed_in_v` is enforced
After discussion with @shaydeci and @ofrimasad, we decided to enforce `removed_in_v`. Basically, using a function that is not supported (version > removed_in_v) will lead to an ImportError with a clear error message explaining how to fix it.

Should be something like
```
ImportError('Function `some_module.get_local_rank` was deprecated and has been removed. Deprecated since version `3.2.0`
        and will be removed in version `4.0.0`. Reason: `Replaced for optimization`.
        Please update your code:
          [-] from `some_module` import `get_local_rank`
          [+] from `new.module.path` import `new_get_local_rank`.')
```
I think it is eventually a nice way to remove a function and force people to move to the new one, while giving them a clear tip on how to do


src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,61 @@
...
+
+    return decorator
+
+
+def make_deprecated(func, reason):
+    def inner(*args, **kwargs):
+        with warnings.catch_warnings():
+            warnings.simplefilter("once", DeprecationWarning)
+            warnings.warn(reason, category=DeprecationWarning, stacklevel=2)
+        warnings.warn(reason, DeprecationWarning)
```
Louis-Dupont, line 58: I tried this, in `make_function_deprecated` but it doesnt change anything.
This is why I am using `wrapper._warned`...

src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,79 @@
+import warnings
+from functools import wraps
+from typing import Optional
+
+
+def deprecate_call(deprecated_in_v: str, removed_in_v: str, target: Optional[callable] = None, reason: str = ""):
```
ofrimasad, line 6: why not "deprecated"?
not removed_in_v, remove_in_v

ofrimasad (commented):
about the warning - ok
about the assert on version, that is incorrect. 
before version release, we bump the `__version__`, at wich point the tests will fail.

src/super_gradients/common/deprecate.py
Louis-Dupont: This was designed for **callable**. We may want to add a deprecated decorator for **parameters** (that's a common use-case as well)
So I thought it would be better to distinguish between both with a clear name. But I'm also ok with `deprecate`, we can go for it if you still think it's better

Louis-Dupont:
Added
- Test to make sure it works well with classes
- Check the version + test
- rename to deprecated

src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Each decorated object will have its own _warned state
+        # This state ensures that the warning will appear only once, to avoid polluting the console in case the function is called too often.
+        wrapper._warned = False
+        return wrapper
+
+    return decorator
+
+
+def make_deprecated(func, reason):
```
shaydeci, line 82: Why do we have this ?
Why not just use our new deprecate ?

shaydeci (changes requested):

src/super_gradients/__init__.py
```diff
@@ -1,3 +1,5 @@
+__version__ = "3.2.0"
+
```
Louis-Dupont, line 2: Version needs to be at the beginning because of circular loops.
This would not be required if we don't automatically check the deprecated versions in runtime (see other comments), but I also don't think it should be our criteria to chose or not to test on runtime.

src/super_gradients/common/deprecate.py
Louis-Dupont: Yeah you're right 9f0bcae7ae1ab34278972f421fccdf74cab0db71

src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,90 @@
...
+        Please update your code:
+          [-] from `some_module` import `get_local_rank`
+          [+] from `new.module.path` import `new_get_local_rank`.
+    """
+
+    def decorator(old_func: callable) -> callable:
+
+        if parse_version(super_gradients.__version__) >= parse_version(remove_in_v):
+            raise EnvironmentError(
+                f"`super_gradients.__version__={super_gradients.__version__}` >= `remove_in_v={remove_in_v}`. "
```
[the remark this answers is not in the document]
shaydeci: I am just talking about the message itself, I do tend to agree its better to crash the code.

src/super_gradients/__init__.py
shaydeci: I experienced this as well!

src/super_gradients/common/deprecate.py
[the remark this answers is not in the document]
Louis-Dupont: Like that [b7d47a7](https://github.com/Deci-AI/super-gradients/pull/1416/commits/b7d47a7436195dda7b54d73f23c58a0421a2477c) ?

src/super_gradients/common/deprecate.py
```diff
@@ -0,0 +1,90 @@
...
+        and will be removed in version `4.0.0`. Reason: `Replaced for optimization`.
+        Please update your code:
+          [-] from `some_module` import `get_local_rank`
+          [+] from `new.module.path` import `new_get_local_rank`.
+    """
+
+    def decorator(old_func: callable) -> callable:
+
+        if parse_version(super_gradients.__version__) >= parse_version(remove_in_v):
+            raise EnvironmentError(
```
shaydeci: Lets change this to import error

Louis-Dupont: [bd2d58c](https://github.com/Deci-AI/super-gradients/pull/1416/commits/bd2d58c50f1906bf38f563d4c2ae46c1bcb7f8e4)

src/super_gradients/training/models/__init__.py
```diff
@@ -135,51 +135,28 @@
...
 
 
-def make_deprecated(func, reason):
-    def inner(*args, **kwargs):
-        with warnings.catch_warnings():
-            warnings.simplefilter("once", DeprecationWarning)
-            warnings.warn(reason, category=DeprecationWarning, stacklevel=2)
-        warnings.warn(reason, DeprecationWarning)
-        return func(*args, **kwargs)
+@deprecated(deprecated_in_v="3.1.0", remove_in_v="3.4.0", target=_make_divisible_current_version)
```
BloodAxe, line 138: Minor complain, but why `deprecated_in_v` and not `deprecated_in_version` or `deprecated_since` ?
Similarly - `remove_in_v` -> `removed_in_version`/`removed_from` maybe?

BloodAxe (approved):
LGTM
Left a minor comment on argument names. Not a big deal tho
Up to you whether to address or ignore if you disagree ;)

ofrimasad (approved):
LGTM
consider @BloodAxe comments about param names

src/super_gradients/training/models/__init__.py
Louis-Dupont: Updated in [65fc918](https://github.com/Deci-AI/super-gradients/pull/1416/commits/65fc91897a13e5f55abd9408f312c5ab4aefc8e7)

BloodAxe (approved):
LGTM

PR #1419, opened 2023-08-27
Hotfix/sg 000 fix predict loading from np torch

Louis-Dupont:
Issue raised: `model.predict` implicitely expected (H, W, C) images.


BloodAxe (changes requested):
This PR seems to be extending scope of supported input types that predict function supports. However no documentation reflecting this pretty serious change is updated. I think it must be explicitly stated in docs what is the expected (or supported) dtypes for predict() function can be and **what will happen** if one pass HWC/CHW/BHWC/BCHW inputs there. 

Also see my inline comment for `load_image`.

src/super_gradients/training/utils/media/image.py
```diff
@@ -77,12 +83,17 @@ def load_image(image: ImageSource) -> np.ndarray:
...
-    :return:      Image as numpy arrays. If loaded from string, the image will be returned as RGB.
+    :return:      Image as numpy arrays (H, W, C). If loaded from string, the image will be returned as RGB.
     """
     if isinstance(image, np.ndarray):
+        if image.shape[0] == 3:
+            image = image.transpose((1, 2, 0))
         return image
     elif isinstance(image, torch.Tensor):
-        return image.numpy()
+        image = image.cpu().numpy()
```
BloodAxe, line 93: What if image is already on cuda? That would hurt the performance quite a lot. Yes the predict() was not meant for high-speed inference, but I also think we should not make it intentionally slower.

src/super_gradients/training/utils/media/image.py
```diff
@@ -77,12 +83,17 @@ def load_image(image: ImageSource) -> np.ndarray:
...
+        if image.shape[0] == 3:
```
BloodAxe, line 94: What about BCHW / BHWC formats?

src/super_gradients/training/utils/media/image.py
```diff
@@ -77,12 +83,17 @@ def load_image(image: ImageSource) -> np.ndarray:
         - str:              A string representing either a local file path or a URL to an image
 
     :param image: Single image of supported types.
-    :return:      Image as numpy arrays. If loaded from string, the image will be returned as RGB.
+    :return:      Image as numpy arrays (H, W, C). If loaded from string, the image will be returned as RGB.
     """
     if isinstance(image, np.ndarray):
+        if image.shape[0] == 3:
```
BloodAxe, line 89: Hard-coded number of input channels of 3 should go away IMO. Instead we should _somehow_ get the number of input channels from the model itself.

src/super_gradients/training/utils/media/image.py
BloodAxe: Second comment regarding torch tensors as input - let think for a second when and why this image type may come in.
To me, the `torch.from_numpy(cv2.imread(...))` is the least common scenario. I would guess the most common use case would be an attempt to run predict on the sample from a dataset:

```
dataset = ..... # User got some dataset
sample = dataset[0]
image = sample[0]

model.predict(image)
```

And here lies the problem, as most likely a normalization was applied to this image. 
There is only one indirect clue that indicates whether or not image was normalized, which is dtype of image tensor. However I'm not sure we should be checking this.

In general, I have a strong concern how exactly we should support torch tensor types to avoid confusion.

Louis-Dupont: I agree, but we run the processings on numpy, so we don't really have a choice.

src/super_gradients/training/utils/media/image.py
Louis-Dupont: `load_image` expects a single image. In the `predict`, we use `load_images` which itself calls `load_image` per image.

We could add a check to make sure a single image is passed.
Or, we could instead move this `if image.shape[0] == 3:` logic into `load_images` and that way "optimize" the code, by not doing it per image, and not "spliting" the batch tensor into multiple single-image array.
I'll consider the second option and see if it all works easily

src/super_gradients/training/utils/media/image.py
Louis-Dupont: Cf my other comment

>  `load_image` expects a single image. In the `predict`, we use load_images which itself calls `load_image` per image.
> 
> We could add a check to make sure a single image is passed.

src/super_gradients/training/utils/media/image.py
Louis-Dupont: > Second comment regarding torch tensors as input - let think for a second when and why this image type may come in.
> To me, the torch.from_numpy(cv2.imread(...)) is the least common scenario. I would guess the most common use case would be an attempt to run predict on the sample from a dataset:

That's the paradox of the `predict`; it was built with the idea of a quickstart (not really optimized) function to get a taste of YoloNAS, when in practice it is often seen as THE way to run predictions (and benchmark) on super-gradients.
So, supporting dataset output was never the target but I agree that some users may want or even already try to do so.

My approach to this would be: lets keep as it is now until we get enough feedback from people wanting/complaining about that, and then we can discuss about it more in-depth

shaydeci:
don;'t forget to PR master320 branch too

src/super_gradients/training/utils/media/image.py
```diff
@@ -50,17 +51,29 @@ def generate_image_loader(images: Union[List[ImageSource], ImageSource]) -> Iter
         images_paths = list_images_in_folder(images)
         for image_path in images_paths:
             yield load_image(image=image_path)
-    elif _is_batch_of_images(images=images):
+    elif _is_4d_array(images):
+        warnings.warn(
+            "It seems you are using predict() with 4D array as input. "
+            "Please note we cannot track whether the input was already normalized or not. "
+            "You will get incorrect results if you feeding batches from train/validation dataloader."
```
shaydeci, line 58: typo: if you are*

src/super_gradients/training/utils/media/image.py
```diff
@@ -44,31 +45,49 @@ def generate_image_loader(images: Union[List[ImageSource], ImageSource]) -> Iter
...
-    :return:        Generator of images as numpy arrays. If loaded from string, the image will be returned as RGB.
+    :return:        Generator of images as numpy arrays (H, W, C). If loaded from string, the image will be returned as RGB.
     """
     if isinstance(images, str) and os.path.isdir(images):
         images_paths = list_images_in_folder(images)
         for image_path in images_paths:
             yield load_image(image=image_path)
-    elif isinstance(images, (list, Iterator)):
+    elif _is_4d_array(images):
+        warnings.warn(
```
shaydeci, line 55: suggestion: just warn once (could be the case that people would iterate through some data loader and call predict on 4d tensors multiple times leaving a mess here).

src/super_gradients/training/utils/media/image.py
```diff
@@ -77,12 +96,26 @@ def load_image(image: ImageSource) -> np.ndarray:
         - str:              A string representing either a local file path or a URL to an image
 
     :param image: Single image of supported types.
-    :return:      Image as numpy arrays. If loaded from string, the image will be returned as RGB.
+    :param input_image_channels: Number of channels that model expects as input.
+                                 This value helps to infer the layout of the input image array.
+                                 As of now this argument has default value of 3, but in future it will become mandatory.
```
shaydeci, line 101: Why? 3 is a pretty good default choice IMO.

src/super_gradients/training/utils/media/image.py
```diff
@@ -44,31 +45,49 @@ def generate_image_loader(images: Union[List[ImageSource], ImageSource]) -> Iter
...
             yield load_image(image=image)
     else:
         yield load_image(image=images)
 
 
+def _is_4d_array(images: ImageSource) -> bool:
+    return isinstance(images, (np.ndarray, torch.Tensor)) and images.ndim == 4
+
+
+def _is_list_of_images(images: ImageSource) -> bool:
```
shaydeci, line 75: I think typing here for images is wrong.

shaydeci (changes requested):
A few minor comments.

src/super_gradients/training/utils/media/image.py
```diff
@@ -44,31 +45,49 @@ def generate_image_loader(images: Union[List[ImageSource], ImageSource]) -> Iter
...
+    return isinstance(images, (list, Iterator))
```
shaydeci, line 76: I think typing for images here is wrong.
Or we should updated ImageSource  to include Iterator[SingleImageSource]?or what am I missing here?

src/super_gradients/training/utils/media/image.py
BloodAxe: Actually, warnings.warn works exactly that.
See the docs at http://docs.python.org/2/library/warnings.html:

Repetitions of a particular warning for the same source location are typically suppressed.

src/super_gradients/training/utils/media/image.py
BloodAxe: Changed type alias for ImageSource. good catch!

src/super_gradients/training/utils/media/image.py
BloodAxe: This is a good default value, however I'm convinced that load_image should not hold any assumption on number of channels. In fact I'd rather require data layout to be passed explicitly, instead of trying to infer it based on number of channels. This is big design change, so I can live without it for now.

Anyhow, my point is Pipeline class should be knowing what input channels should be (It has all the information needed to infer this from model instance) and propagate this value to `load_image`.
So the idea from removing the default value then is to enforce this parameter is explicitly passed.

src/super_gradients/training/utils/media/image.py
BloodAxe: Fixed in 985eff2ce59370bdf8b7e9fbe12a0e40a14c62e5

src/super_gradients/training/utils/media/image.py
BloodAxe: Fixed in 1337b83b832831704bdcdc23b651c39c1af81027

shaydeci (approved):
LGTM

BloodAxe (approved):

PR #1420, opened 2023-08-27
Hotfix/sg 000 improve breaking change detection with imports

Louis-Dupont:
Issue
Currently the breaking change tests is sensitive to 
- import removal in __init__
- Function removal in any module

BUT
- Removing an import should be OK if it's not a SG dependency (e.g. `import warnings`)
- Removing a function from a module should be OK if it is instead imported in the same module


BloodAxe (approved):
LGTM

PR #1421, opened 2023-08-27
Adding a deprecation option into registeries

Louis-Dupont:
### Context
Use the class names instead of custom snake_case format to name objects

### What this PR includes
This PR introduces the logic to deprecate old snake_case names.
Check the unittest to better understand the logic.
Changing the registration names will be done in followup PRs for more clarity (you check out this PR https://github.com/Deci-AI/super-gradients/pull/1446 which will introduce the first object renaming)


BloodAxe (approved):
LGTM

PR #1442, opened 2023-09-01
Change use case to catch detected object

Pslydhh:
This PR fix NoneType error from:
`prediction = model_predictions[0].prediction`
Maybe some objects' structure changed?


BloodAxe:
Thanks for your contribution. Can you please elaborate the change you've introduced and motivation behind?
Anyhow `model_predictions._images_prediction_lst[0]` is accessing a private property and this is a bad practice and should be avoided.

manoj-samal:
did someone find the solution on this i encountered the same error..
any findings on this please help

PR #1445, opened 2023-09-03
Fix checkpoint doc

Louis-Dupont:
1. Some `.md` were not properly updated since the change on how checkpoints are structured. 
2. The quickstart `.md` were introducing checkpoints but with way less depth than the `checkpoints.md`. I added some extra information which I believe will bring more clarity, on top of adding a link the the main `checkpoints.md`.


BloodAxe (approved):
LGTM

PR #1449, opened 2023-09-04
Updating OD example by adding additional details on customizing metri…

BloodAxe:
…cs and adding visualization

### DIFF
--- a/documentation/source/ObjectDetection.md
+++ b/documentation/source/ObjectDetection.md
@@ -40,6 +40,65 @@ val_dataloader = dataloaders.get(name='coco2017_val',
                                  )
 ```
 
+### Loss functions
+
+Generally speaking, in object detection task the loss function is tightly coupled with the model and cannot be used interchangeably. E.g. you cannot use YoloX loss with YoloNAS model and vice versa.
+This is different from classification or segmentation task where model output is "standard" and usually does not change. 
+In Object Detection task, the model output format may vary greatly and also training objective is often tailored for a specific model architecture.
+
+To indicate compatibility between a model and a loss function, we use the convention of model name and loss function starting from the same prefix name. For example: `SSDLiteMobileNetV2` model & `SSDLoss`, `YoloX_S` and `YoloXFastDetectionLoss`, etc.
+
+Of course, you are free to adjust hyperparameters of the loss function to your liking. Let's check a `PPYoloELoss` loss class as an example:
+It has the following constructor:
+
+```python
+@register_loss(Losses.PPYOLOE_LOSS)
+class PPYoloELoss(nn.Module):
+    def __init__(
+        self,
+        num_classes: int,
+        use_varifocal_loss: bool = True,
+        use_static_assigner: bool = True,
+        reg_max: int = 16,
+        classification_loss_weight: float = 1.0,
+        iou_loss_weight: float = 2.5,
+        dfl_loss_weight: float = 0.5,
+    ):
+        ...
+```
+
+In your recipe you can pass the desired values for each parameter. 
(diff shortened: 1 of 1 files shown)


documentation/source/ObjectDetection.md
```diff
@@ -76,12 +175,58 @@ Box coordinates are in absolute (pixel) units.
...
 By visualizing the predicted boxes, developers and researchers can identify errors or inaccuracies in the model's output and adjust the model's architecture or training data accordingly.
 
-SuperGradients provide an implementation of [DetectionVisualizationCallback](https://docs.deci.ai/super-gradients/docstring/training/utils.html#training.utils.callbacks.callbacks.DetectionVisualizationCallback). 
-You can use this callback in your training pipeline to visualize predictions during training. For this, just add it to `training_hyperparams.phase_callbacks` in your yaml.
-During training, the callback will generate a visualization of the model predictions and save it to the TensorBoard or Weights & Biases depending on which logger you
-are using (Default is Tensorboard). 
+#### Extreme Batch Visualization during training
+
+SuperGradients provide an implementation of [ExtremeBatchDetectionVisualizationCallback](https://docs.deci.ai/super-gradients/docstring/training/utils.html#src.super_gradients.training.utils.callbacks.callbacks.ExtremeBatchDetectionVisualizationCallback).
```
Louis-Dupont, line 180: `provide` -> `provides`

documentation/source/ObjectDetection.md
```diff
@@ -112,6 +257,16 @@ The saved train image for a dataset with a mosaic transform should look somethin
...
 
+#### Visualization of predictions after training using predict()
+
+If you would like to do the visualization outside of training you can use `predict()` method that is implemented for most of our detection models.
+```python
+model = models.get("yolox_s", pretrained_weights="coco", num_classes=80)
+model.predict("https://deci-pretrained-models.s3.amazonaws.com/sample_images/beatles-abbeyroad.jpg").show()
+```
+
+See for more details on using [Predict API](https://docs.deci.ai/super-gradients/V3_1/documentation/source/ModelPredictions.html).
```
Louis-Dupont, line 268: IMO it's better to use local path because
1. More robust if the url changes; an example is that in your case you have `V3_1` in your name. But it can also happen if the structure of the documentation hub is modified.
2. When going through `.md` in Pycharm or Github, I think it is nicer to not be redirected to another place. I clearly prefer to stay in the same "environment".

The script that generates the documentation platform converts the local `.md` path to the full URL, so we don't need to worry about it. I would just write `[Predict API](ModelPredictions.md)`

Louis-Dupont (commented):
Nice adding! I left a small note concerning the path of documentation pages, let me know if you want to change it or not and then we can merge

documentation/source/ObjectDetection.md
BloodAxe: Thanks, will address that.
PS:I think we need docs on how to contribute to docs explaining these useful tips & tricks )

documentation/source/ObjectDetection.md
BloodAxe: Fixed in 4ad8fdf3843bc9b6882c329055ea664ba31ddbe1

documentation/source/ObjectDetection.md
BloodAxe: Fixed in 4ad8fdf3843bc9b6882c329055ea664ba31ddbe1

Louis-Dupont: I added a ticket to write a guide on how to write guides

Louis-Dupont (approved):
LGTM

PR #1458, opened 2023-09-12
Fix Breaking Change Detection on Inheritence

Louis-Dupont:
### Context
Let's imagine we change the code

FROM
```python
class MyClass:
    def method(self): pass
```

TO
```python
class MyNewClass:
    def method(self): pass

class MyClass(MyNewClass): pass
```

This is **NOT** a breaking change, because `MyClass` inherits from `MyNewClass` which still includes `method`.

**Issue**: current class signature extractor (implemented as part of the breaking change detection) currently misses this and will raise that the new `MyClass` does not include `method`.


BloodAxe (dismissed):
LGTM

BloodAxe (approved):
LGTM

PR #1460, opened 2023-09-12
Update transforms.py cv2.wrapAffine dsize should be (xshape,yshape))

aromalma:
a

### DIFF
--- a/src/super_gradients/training/transforms/transforms.py
+++ b/src/super_gradients/training/transforms/transforms.py
@@ -1284,7 +1284,7 @@ def random_affine(
     targets_seg = np.zeros((targets.shape[0], 0)) if targets_seg is None else targets_seg
     M = get_affine_matrix(img.shape[:2], target_size, degrees, translate, scales, shear)
 
-    img = cv2.warpAffine(img, M, dsize=target_size, borderValue=(border_value, border_value, border_value))
+    img = cv2.warpAffine(img, M, dsize=target_size[::-1], borderValue=(border_value, border_value, border_value))
 
     # Transform label coordinates
     if len(targets) > 0:


BloodAxe:
Thanks for your PR. 
Can you please share the portion of dataset config you are using?

aromalma:
> Thanks for your PR. Can you please share the portion of dataset config you are using?
"dataset_params": {
        "train_dataset_params": "{'data_dir': '/home/groot/Aromal/yolonas-train-infer/nn/yolonas_train/', 'images_dir': 'train/', 'labels_dir': 'train/', 'classes': ['Signboard_Caution_Board', 'Signboard_Chevron_Board_Left', 'Signboard_Gantry_Board', 'Signboard_Gantry_Board_2', 'Signboard_Hazard_Board', 'Signboard_Information_Board_B', 'Signboard_Information_Board_G', 'Signboard_Information_Board_N', 'Signboard_Mandatory_Board', 'VMS_Gantry', 'Delineator', 'Bus_Shelter', 'ECB_(SOS)', 'Encroachment', 'Hectometer_stone', 'Kilometer_Stone', 'Solar_Blinker', 'Double_Arm_Street_Light', 'Street_Light_Slanding', 'Pot_Holes', 'Signboard_Mandatory_Board_Stop', 'Single_Arm_Light_Slanting', 'High_Mast_Light', 'Expansion_Joint', 'Hoardings', 'Flex_Banner', 'Shop_Board', 'Water_Stagnation', 'Excess_Mud', 'ATCC', 'CCTV', 'DB_box', 'Wall_poster', 'Vegetation_Growth', 'Cattle_movement', 'Car', 'Motorbike', 'Bus', 'Truck', 'Signboard_Mandatory_Board_White'], 'input_dim': [352, 640], 'cache_dir': None, 'cache': False, 'cache_annotations': True, 'ignore_empty_annotations': True, 'transforms': [{'DetectionMosaic': {'input_dim': [352, 640], 'prob': 0.5, 'enable_mosaic': False, 'border_value': 114}}, {'DetectionRandomAffine': {'degrees': 6, 'translate': 0.05, 'scales': [0.7, 1.2], 'shear': 0.5, 'target_size': [352, 640], 'filter_box_candidates': True, 'wh_thr': 2, 'area_thr': 0.1, 'ar_thr': 20}}, {'DetectionHSV': {'prob': 0.3, 'hgain': 5, 'sgain': 6, 'vgain': 6}}, {'DetectionPaddedRescale': {'input_dim': [352, 640], 'pad_value': 114}}, {'DetectionTargetsFormatTransform': {'input_dim': [352, 640], 'output_format': 'LABEL_CXCYWH'}}], 'class_inclusion_list': None, 'max_num_samples': None}",


these are the dataset parameters i used and the aspect ratio is changing during affine transformation.

aromalma:
a
This the output after affine transform after this modification

BloodAxe:
After carefully reviewing the existing transformations and existing dataset configs it seems the problem is deeper than it looked at the first glance. Seems our transformations are not using consistent convention whether image size defined as (H,W) or (W,H). 
Some transformations assumes (H,W) while other - (W,H) and this is the first place we should fix in the first place. 

Merging your PR as is considered a breaking change. Since there could be other projects that expect current behavior (Eventually what you need to do is flip values for `target_size` argument when creating `DetectionRandomAffine` in your local config) and that would cause existing configs to work incorrectly.

The right way to solve this would be:
1) We decide how we approach error-proof way of defining output size of images in our transformations
2) Implement it, making existing properties deprecated. 
3) Document thoroughly the convention
3) Make a new release

Clearly this is one a one-day effort, so as a temporary solution I suggest you to fix it locally as follows:

```
DetectionRandomAffine:
   ```
   target_size: [640, 352] # Instead of [352, 640]
```

Again, thanks for bringing this issue up!

aromalma:
> Again, thanks for bringing this issue up!
it is my pleasure

PR #1464, opened 2023-09-13
Update Segmentation.md

donatoaz:
Updates missing link to dataset instructions. I assume`Data.md` is the markdown it is related to.


Louis-Dupont (approved):
Good catch!

PR #1471, opened 2023-09-19
Feature/sg 000 add adapter

Louis-Dupont:
# Introduction of Adapter logic

### Why Do We Need This?
We aim to allow seamless integration with SG for users who are already using DG. 

### Solution (updated)

If the user doesnt want to set anything in advance (in such case he will be asked questions) OR if he already has a cache (from DG or previous run)
```python
val_loader = DetectionDataloaderAdapterFactory.from_dataset(
    dataset=train_set,
    config_path="<path>",
    batch_size=10,
    drop_last=True,
)
```

If the user wants to set in advnace some parameters, he can still use `DetectionDataConfig`
```python

adapter_config = DetectionDataConfig(images_extractor=lambda sample: sample[0], labels_extractor=voc_format_to_bbox, cache_path=analyzer.data_config.cache_path)
train_loader = DetectionDataloaderAdapterFactory.from_dataset(
    dataset=train_set,
    adapter_config=adapter_config,
    batch_size=10,
    drop_last=True,
)
```

Now, the same is possible if the user already has a dataloader.
```python
val_loader = DetectionDataloaderAdapterFactory.from_dataloader(
    dataloader=train_loader,
    config_path="...",
)
```
---
Notebook
https://colab.research.google.com/drive/1clgfrdCrg5cWSl7nrjY0zVndz06SIC0P?usp=sharing


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1136,6 +1136,14 @@ def forward(self, inputs, targets):
...
             raise ValueError("`test_loaders` must be a dictionary mapping dataset names to DataLoaders")
 
+        from super_gradients.training.datasets.collate_fn import maybe_calibrate_dataset_adapter
+
+        # Not mandatory but ensures that - if the dataloader use DatasetAdapterCollateFN - the adapter is properly calibrated before iterating over it.
+        maybe_calibrate_dataset_adapter(dataloader=self.train_loader)
+        maybe_calibrate_dataset_adapter(dataloader=self.valid_loader)
+        for test_loader in self.test_loaders.values():
+            maybe_calibrate_dataset_adapter(dataloader=test_loader)
+
```
Louis-Dupont, line 1146: I don't like this too much, I feel like it should not be called here, it should not be the responsability of the Trainer.
But, it smoothens the user experience - if someone uses the `DatasetAdapterCollateFN` and did not calibrate the adapter, we do it for him.
I'm open to remove it, or to move to somewhere else

src/super_gradients/recipes/cityscapes_stdc_seg50.yaml
```diff
@@ -55,7 +55,7 @@ arch_params:
 
 checkpoint_params:
   checkpoint_path:
-  load_backbone: True
+  load_backbone: False
```
BloodAxe, line 58: Accidental editing?

src/super_gradients/training/datasets/collate_fn.py
```diff
@@ -0,0 +1,421 @@
...
+from super_gradients.common.factories.collate_functions_factory import CollateFunctionsFactory
+from super_gradients.training.utils.detection_utils import xyxy2cxcywh
+
+from super_gradients.common.environment.device_utils import device_config
+from super_gradients.common.environment.ddp_utils import get_local_rank
+from super_gradients.training.utils.distributed_training_utils import wait_for_the_master
+from super_gradients.training.datasets.segmentation_datasets.segmentation_dataset import SegmentationDataSet
+
+
+class DatasetItemsException(Exception):
```
BloodAxe, line 28: Let's move expection classes to some subpackage?  E.g `super_gradients.exceptions` or `super_gradients.common.exceptions` ? Of course those exception classes should not have imports of any other classes, but I feel having some standardized place for all exceptions is a good thing.

src/super_gradients/training/datasets/collate_fn.py
```diff
@@ -0,0 +1,421 @@
...
+        :param collate_type: type of the collate that caused the exception
+        :param expected_item_names: tuple of names of items that are expected by the collate to be returned from the dataset
+        """
+        collate_type_name = collate_type.__name__
+        num_sample_items = len(data_sample) if isinstance(data_sample, tuple) else 1
+        error_msg = f"`{collate_type_name}` only supports Datasets that return a tuple {expected_item_names}, but got a tuple of len={num_sample_items}"
+        super().__init__(error_msg)
+
+
+class BaseDatasetAdapterCollateFN(ABC):
```
BloodAxe, line 41: Let's split this file and have one file per class.
So instead of `src/super_gradients/training/datasets/collate_fn.py` there would be
```
src/super_gradients/training/datasets/collate_fn/__init__.py
src/super_gradients/training/datasets/collate_fn/ppyolo_collate_fn.py
src/super_gradients/training/datasets/collate_fn/ppyolo_crowd_collate_fn.py
...
```

Reason: 
One file easier to digest
Easier to do diff between two files and see their difference.

src/super_gradients/training/datasets/collate_fn.py
```diff
@@ -0,0 +1,421 @@
...
+        :param adapter:             Dataset adapter to use
+        """
+        self._adapter = adapter
+        self._collate_fn = collate_fn or default_collate
+        self._adapt_on_batch = adapter.data_config.is_batch or adapter.data_config.is_batch is None
+
+        if isinstance(self._collate_fn, type(self)):
+            raise RuntimeError(f"You just tried to instantiate {self.__class__.__name__} with a `collate_fn` of the same type, which is not supported.")
+
+        self._require_calibration = not self._adapter.data_config.is_completely_initialized
```
BloodAxe, line 63: I don't get notion of 'calibration' in this context. Since we are using calibration in the context of quantization I feed this may confuse users (as it confuses me). 
What does "calibration" mean? Is there a better term to use?

src/super_gradients/training/datasets/collate_fn.py
Louis-Dupont: Fair enough, I'll do that

src/super_gradients/training/datasets/collate_fn.py
Louis-Dupont: I think the main advantage is when someone builds a new class/function and wants to raise an exception, it's easier to find whether or not such exception already exists.
I'll open a PR for that

src/super_gradients/training/datasets/collate_fn.py
Louis-Dupont: I agree, just kind of hoped that the person spotting this would have a better naming idea 🤣

Louis-Dupont: maybe `setup_adapter` and `require_setup` could actually do the job

src/super_gradients/training/datasets/collate_fn.py
Louis-Dupont: There was already a dataset_exception apparently. I moved the exception there (among other) in this PR
https://github.com/Deci-AI/super-gradients/pull/1470

src/super_gradients/training/datasets/collate_fn.py
Louis-Dupont: I did it here https://github.com/Deci-AI/super-gradients/pull/1470
Eventually, I gave up on moving it in the `dataset` folder, I just lost more than 1h of trying to fix circular import loops and I think it is just too deep...

src/super_gradients/training/datasets/collate_fn.py
Louis-Dupont: I now use `maybe_setup_dataloader_adapter` and `_maybe_setup_adapter` (which are functions)

src/super_gradients/training/sg_trainer/sg_trainer.py
Louis-Dupont: removed

src/super_gradients/training/__init__.py
```diff
@@ -1,11 +1,11 @@
 # PACKAGE IMPORTS FOR EXTERNAL USAGE
 import super_gradients.training.utils.distributed_training_utils as distributed_training_utils
+from super_gradients.common import MultiGPUMode, StrictLoad, EvaluationType
 from super_gradients.training.datasets import datasets_utils, DataAugmentation
+from super_gradients.training.pre_launch_callbacks import modify_params_for_qat
+from super_gradients.training.qat_trainer import QATTrainer
```
BloodAxe, line 6: I feel we need to start using `isort` as a pre-commit hook to ensure order of import statements remains unchanged :) 
I see no imports added/removed, just order changed. If possible - revert since the change not necessary critical to this PR.

src/super_gradients/training/dataloaders/adapters.py
```diff
@@ -0,0 +1,154 @@
...
+        dataset: torch.utils.data.Dataset,
+        adapter_config: Optional[DataConfig] = None,
+        adapter_cache_path: Optional[str] = None,
+        collate_fn: Optional[callable] = None,
+        **dataloader_kwargs,
+    ) -> torch.utils.data.DataLoader:
+        """Wrap a DataLoader to adapt its output to fit SuperGradients format for the specific task.
+
+        :param dataset:                 Dataset to adapt.
+        :param adapter_config:          Adapter configuration. Use this if you want to hard code some specificities about your dataset.
```
BloodAxe, line 34: `hardcode` => explicitly set
`specificities` => specific params
about your dataset => of your dataset

tests/integration_tests/data_adapter/data_adapter_on_other.py
```diff
@@ -0,0 +1,329 @@
...
+    DetectionDataloaderAdapter,
+    SegmentationDataloaderAdapter,
+    ClassificationDataloaderAdapter,
+    DetectionDatasetAdapterCollateFN,
+    SegmentationDatasetAdapterCollateFN,
+)
+
+
+def datagradients_detection():
+    data_dir = "/Users/Louis.Dupont/Desktop/DataGradients/data-examples/solar_panels"  # ...  # Fill with the <path-to>/data_dir
```
BloodAxe, line 20: Absolute path

tests/integration_tests/data_adapter/test_data_adapter_on_sg.py
```diff
@@ -0,0 +1,169 @@
...
+from super_gradients import setup_device
+from super_gradients.training.dataloaders.dataloaders import coco2017_val, cityscapes_stdc_seg50_val, cifar10_val
+from super_gradients.training.utils.collate_fn import (
+    DetectionDataloaderAdapter,
+    SegmentationDataloaderAdapter,
+    ClassificationDataloaderAdapter,
+)
+
+
+def test_python_detection():
```
BloodAxe, line 17: Tests are not membest or TestCase class. Was it intentional?

requirements.txt
```diff
@@ -36,3 +36,4 @@ numpy<=1.23
 rapidfuzz
 json-tricks==3.16.1
 onnx-simplifier>=0.3.6,<1.0
+data-gradients>=0.1.5
```
BloodAxe, line 39: I feel we should have a documentation page in SG describing this feature

BloodAxe: Maybe in a form of Notebook that will be exported to markdown 😉

src/super_gradients/training/__init__.py
Louis-Dupont: The original purpose of this was circular import loops. Not sure if it is stil an issue, I'll try to revert and see.

About `isort` I didnt know about it. It looks great and I would be up for it BUT we need to be careful with circular import loops. Currently, some of the code is sensitive to the import order. 
That's bad and should be fixed by improving the design of the code, but yeah it should be fixed

BloodAxe:
Thanks for the Colab notebook. It really helpful to see how the intended usage.
I have few comments regarding the notebook, nothing critical or wrong, more a suggestsions:

* DetectionDataloaderAdapterFactory / SegmentationDataloaderAdapterFactory - may I suggest DataAdapterForDetection / DataAdapterForSegmentation ? 

* `adapter_cache_path` argument name may be a bit confusing to users non familiar with internals and one may assume we will store some data samples in that location. Perhaps if we rename this argument as "config_path" that would be clearer to the user.

* > *As* you can see, the labels are normalized (0-1). This is all right, but it is not the format expected by SuperGradients. 
Actually this was ❓❓❓ moment for me, why that dataset also normalize masks and how SG denormalize them and why. Lost the track why we modify targets. Maybe should have been explained a bit better. Usually in segmentation datasets we don't normalize targets so maybe in notebook we should show a more common use case?

* For the questions that has fixed number of options to choose from, perhaps we can show some love to user and heuristicaly pick a default option (Which will be chosen if user just hit Enter):

```
--------------------------------------------------------------------------------
Which dimension corresponds the image channel? 
--------------------------------------------------------------------------------
Image shape: torch.Size([1, 3, 512, 512])
Options:
[0] | 0
[1] | 1 [Suggested value, Enter to accept]
[2] | 2
[3] | 3
```
Sure, heuristics may not handle ALL use cases but would certainly save some time in 90% of the cases.

* `yolonas_recipe = load_recipe(config_name="coco2017_yolo_nas_s", overrides=[f"arch_params.num_classes={20}", "training_hyperparams.max_epochs=1", "training_hyperparams.mixed_precision=False"])`

As we've already asked user about number of classes when creating adapter data loader, we should be able to pull this information from that instance:

`yolonas_recipe = load_recipe(config_name="coco2017_yolo_nas_s", overrides=[f"arch_params.num_classes={train_loader.adapter_config.num_classes}", "training_hyperparams.max_epochs=1", "training_hyperparams.mixed_precision=False"])`

src/super_gradients/training/__init__.py
Louis-Dupont: From what I recall, what often creates nightmare circular loops are
- DDP utils (we have one file in common, one in training and they rely on other modules, while being used by many other modules)
- Datasets - Don't recall why though but I think it's related to both `register` and `Transforms`
- `__init__.py` - We have some modules in `__init__.py` that should not be included I believe. For instance why `is_distributed` is in the root of the project ? I never even used it outside of SG.

requirements.txt
Louis-Dupont: Yeah, that was the plan, notebook to markdown

src/super_gradients/training/dataloaders/dataloaders.py
```diff
@@ -81,6 +82,8 @@ def get_data_loader(config_name: str, dataset_cls: object, train: bool, dataset_
 
     dataloader = DataLoader(dataset=dataset, **dataloader_params)
     dataloader.dataloader_params = dataloader_params
+
+    maybe_setup_dataloader_adapter(dataloader=dataloader)
```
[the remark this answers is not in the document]
Louis-Dupont: I agree with you 99%, but it is technically possible that the user would have a custom dataset (not SG at all), register it and want to use the adapter with recipe.

It's not the most probable use case, and we may prefer to not support such option (at least until it's requested). It's usually a good rule of thumb to avoid unnecessary complexity.
But the solution is here and takes 1 line, and that's why I thought maybe we should just support it already.

No hard opinion on that, we can remove if you still think so

Louis-Dupont:
> DetectionDataloaderAdapterFactory / SegmentationDataloaderAdapterFactory - may I suggest DataAdapterForDetection / DataAdapterForSegmentation ?

Not fundamentally against the change, but I have a small concern; `DataAdapterForDetection.from_dataset(dataset=...)` It's not clear that this generates a dataloader. `DetectionDataloaderAdapterFactory` was not 100% explicit but with the name it's already a bit more intuitive.

We could change the method to convey this idea, for instance `DataAdapterForDetection.build_dataloader_from_dataset`, but I can't find a good name. here `build_dataloader_from_dataset` sounds bad, especially when we think about the `from_dataloader` version which would be `build_dataloader_from_dataloader`.

Eventually, my point is I would like the name (of class and method combined) to include the info that this builds a "dataloader which adapts X for a task Y" with X (dataset/dataloader).
Any idea of better naming? Or do you think it's fine to not refer to dataloader?

Louis-Dupont:
> adapter_cache_path argument name may be a bit confusing to users non familiar with internals and one may assume we will store some data samples in that location. Perhaps if we rename this argument as "config_path" that would be clearer to the user.

Yeah I wasnt too sure about whether to simplify or to be more explicit. But I think you're right, with the context it is (relatively) clear that the config_path is that of the current object (i.e. adapter)

Louis-Dupont:
> For the questions that has fixed number of options to choose from, perhaps we can show some love to user and heuristicaly pick a default option (Which will be chosen if user just hit Enter):

That's an interesting idea. I will create a task for this, and maybe to include this idea to other questions (if possible/useful).

Louis-Dupont:
> As we've already asked user about number of classes when creating adapter data loader, we should be able to pull this information from that instance:
> 
> `yolonas_recipe = load_recipe(config_name="coco2017_yolo_nas_s", overrides=[f"arch_params.num_classes={train_loader.adapter_config.num_classes}", "training_hyperparams.max_epochs=1", "training_hyperparams.mixed_precision=False"])`

Good point, i'll change that

src/super_gradients/training/dataloaders/adapters.py
Louis-Dupont: [975528b](https://github.com/Deci-AI/super-gradients/pull/1471/commits/975528bdc716f36cf4f3c472c887f95b6e7d2fd8)

tests/integration_tests/data_adapter/test_data_adapter_on_sg.py
Louis-Dupont: Not sure what to do with these tests.
I am testing Roboflow datasets, for the sake of using using adatper on non-sg datasets, but this requires external datasets.
Downloading the datasets would slow down the tests by a lot, I really don't think we want that.

So until now I only used it manually, but it's also very sad to not add these tests to TestCase.
What do you think should be the right way ? Just replacing these datasets by dummy datasets ?

BloodAxe: I think using tinycoco dataset that we already have and implementing a few custom classes that return samples in a variety of formats would be great way to test this.

src/super_gradients/training/dataloaders/dataloaders.py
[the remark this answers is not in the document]
BloodAxe: Okay okay, let's keep it then)

Louis-Dupont:
Note: the build fails because we did not release yet the latest DG version.
We need to first release DG, and then update the requirements.txt

BloodAxe (approved):
LGTM

BloodAxe (approved):
LGTM

PR #1473, opened 2023-09-19
Fixed loading preprocessing params from pretrained weights

BloodAxe:
If pretrained weights checkpoint contains preprocessing params ensures that it is properly loaded.


src/super_gradients/training/utils/checkpoint_utils.py
```diff
@@ -1598,3 +1582,41 @@ def load_pretrained_weights_local(model: torch.nn.Module, architecture: str, pre
...
+def _maybe_load_preprocessing_params(model: Union[nn.Module, HasPredict], checkpoint: Mapping[str, Tensor]) -> bool:
+    """
+    Tries to load preprocessing params from the checkpoint to the model.
+    The function does not crash, and raises a warning if the loading fails.
+    :param model: Instance of nn.Module
+    :param checkpoint: Entire checkpoint dict (not state_dict with model weights)
+    :return: True if the loading was successful, False otherwise.
+    """
+    model = unwrap_model(model)
+    preprocessing_params_in_checkpoint = "processing_params" in checkpoint.keys()
```
Louis-Dupont, line 1605: `preprocessing_params_in_checkpoint` make it feel like it represents the preprocessing params, when it is in fact a bool.
`checkpoint_has_preprocessing_params` or `is_checkpoint_with_preprocessing_params` could be clearer

src/super_gradients/training/utils/checkpoint_utils.py
```diff
@@ -1598,3 +1582,41 @@ def load_pretrained_weights_local(model: torch.nn.Module, architecture: str, pre
...
+    logger.info(f"Successfully loaded pretrained weights for architecture {architecture}")
+
+
+def _maybe_load_preprocessing_params(model: Union[nn.Module, HasPredict], checkpoint: Mapping[str, Tensor]) -> bool:
+    """
+    Tries to load preprocessing params from the checkpoint to the model.
+    The function does not crash, and raises a warning if the loading fails.
+    :param model: Instance of nn.Module
+    :param checkpoint: Entire checkpoint dict (not state_dict with model weights)
+    :return: True if the loading was successful, False otherwise.
```
Louis-Dupont, line 1602: Really not blocking, and I dont remember if we really took a decision, but a couple of weeks ago we talked about aligning docstrings like this
```
:param model:      Instance of nn.Module
:param checkpoint: Entire checkpoint dict (not state_dict with model weights)
...
```

Louis-Dupont (approved):
Looks great, just left 2 non-blocking comments, you can always ignore them.

About the tests, is it possible to set `pretrained_weights` to be a local path ? I never tried and need to dig into the code to check, but maybe it would work ?
In that case we can just quickly train a model locally and use its checkpoint path

BloodAxe:
@Louis-Dupont  I've added the test that saves a temporary checkpoint with preprocessing params and then ensures that predict() works as expected.
Had to patch a pretrained dict loading method though a bit :)

Louis-Dupont (approved):
LGTM

PR #1492, opened 2023-10-02
Introduce sample-centric keypoint transforms

BloodAxe:
What this PR does:

* Introduce `PoseEstimationSample` that holds all information about a single dataset sample
* Refactor all existing transforms (and add a bunch of new ones) for pose estimation that enables then to work with new `PoseEstimationSample` as well as existing API that pass a tuple (`image`,`mask`,`boxes`,...)
* Move each transform to a single file (Just to make editing / navigating a little easier). Old import 
 paths are still working fine.

New transforms:
* Random 90-degree rotation
* Mixup
* Mosaic
* Brightness/Contrast
* Rescale to fixed size (Non-preserving aspect ratio)

Sanity checks:
* Existing DEKR recipe still works as expected. No API change on the user level
* New & existing transforms tested extensively during YoloNASPose training 

We can merge this PR as is to master or we can 'fake approve' this PR (If everyone is good with the concept) and I will keep adding yolo-nas pose stuff on top of it (Through PR to this branch)


src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
```diff
@@ -0,0 +1,193 @@
...
+        )
+
+    def _get_affine_matrix(self, img, angle, scale, dx, dy):
+        """
+
+        :param center: (x,y)
+        :param scale:
+        :param output_size: (rows, cols)
+        :param rot:
+        :return:
```
Louis-Dupont, line 75: docstrings + typing

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def apply_to_sample(self, sample: PoseEstimationSample) -> PoseEstimationSample:
+        """
+
+        :param image: (np.ndarray) Image of shape [H,W,3]
+        :param mask: Single-element array with mask of [H,W] shape.
+        :param joints: Single-element array of joints of [Num instances, Num Joints, 3] shape. Semantics of last channel is: x, y, joint index (?)
+        :param area: Area each instance occipy: [Num instances, 1]
+        :return:
+        """
```
Louis-Dupont, line 91: wrong docstring

src/super_gradients/training/transforms/keypoints/keypoints_pad_if_needed.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@register_transform(Transforms.KeypointsPadIfNeeded)
+class KeypointsPadIfNeeded(AbstractKeypointTransform):
+    """
+    Pad image and mask to ensure that resulting image size is not less than `output_size` (rows, cols).
+    Image and mask padded from right and bottom, thus joints remains unchanged.
+    """
+
+    def __init__(self, min_height: int, min_width: int, image_pad_value: int, mask_pad_value: float, padding_mode: str = "bottom_right"):
```
Louis-Dupont, line 19: I just learned recently that python includes `Literal` typing which may be appropriate here

```python
from typing import Literal


padding_mode: Literal["bottom_right", "center"] = "bottom_right"
```

I dont know if you ever used it or have an opinion on its usage, but to me it sounds like a good fit for this situation, it would add explicitness + type checking

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
```diff
@@ -0,0 +1,193 @@
...
+    def apply_to_areas(cls, areas: np.ndarray, mat):
+        det = np.linalg.det(mat[:2, :2])
+        return (areas * abs(det)).astype(areas.dtype)
+
+    @classmethod
+    def apply_to_bboxes(cls, bboxes_xywh: np.ndarray, mat: np.ndarray):
+        """
+
+        :param bboxes: (N,4) array of bboxes in XYWH format
+        :param mat:
```
Louis-Dupont, line 133: same, docstring for mat

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
```diff
@@ -0,0 +1,193 @@
...
+
+            if sample.areas is not None:
+                sample.areas = self.apply_to_areas(sample.areas, mat_output)
+
+            sample = sample.sanitize_sample()
+
+        return sample
+
+    @classmethod
+    def apply_to_areas(cls, areas: np.ndarray, mat):
```
Louis-Dupont, line 124: `mat` not typing & missing docstring (would be nice to know the shape at least)

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
```diff
@@ -0,0 +1,193 @@
...
+        outside_bottom = keypoints[:, :, 1] >= image_height
+
+        joints_outside_image = outside_left | outside_top | outside_right | outside_bottom
+
+        keypoints_with_visibility[:, :, 0:2] = keypoints
+        keypoints_with_visibility[joints_outside_image, 2] = 0
+        return keypoints_with_visibility.astype(dtype, copy=False)
+
+    @classmethod
+    def apply_to_image(cls, image, mat, interpolation, padding_value, padding_mode=cv2.BORDER_CONSTANT):
```
Louis-Dupont, line 182: typing

src/super_gradients/training/transforms/keypoints/keypoints_random_horisontal_flip.py
```diff
@@ -0,0 +1,70 @@
...
+    def apply_to_image(self, image):
+        return np.ascontiguousarray(np.fliplr(image))
+
+    def apply_to_keypoints(self, keypoints, cols):
+        keypoints = keypoints.copy()
+        keypoints = keypoints[:, self.flip_index]
+        keypoints[:, :, 0] = cols - keypoints[:, :, 0] - 1
+        return keypoints
+
+    def apply_to_bboxes(self, bboxes, cols):
```
Louis-Dupont, line 57: typing

src/super_gradients/training/transforms/keypoints/keypoints_rescale.py
```diff
@@ -0,0 +1,72 @@
...
+        keypoints[:, :, 0] *= sx
+        keypoints[:, :, 1] *= sy
+        return keypoints
+
+    @classmethod
+    def apply_to_bboxes(cls, bboxes, sx, sy):
+        bboxes = bboxes.astype(np.float32, copy=True)
+        bboxes[:, 0::2] *= sx
+        bboxes[:, 1::2] *= sy
+        return bboxes
```
Louis-Dupont, line 66: Typing + docstring.
(Also, same question as above, was there a motivation for using `@ classmethod` instead of `@staticmethod` ?)

Louis-Dupont (commented):
I had a quick overview of the code. It looks great, for now I only have notes on documentation.

I will go over it later, now that I better understand the changes.

src/super_gradients/training/transforms/keypoints/keypoints_pad_if_needed.py
Louis-Dupont: Actually, it only exists since `python==3.8`, and if I'm not wrong we support `python==3.7`, so that's a bad idea

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
```diff
@@ -0,0 +1,132 @@
...
+        :param min_bbox_area:  Min bbox area of the pose instances to keep them
+        :return:               Transformed sample
+        """
+        applied_transforms_so_far = []
+        for t in transforms:
+            if t.additional_samples_count == 0:
+                sample = t.apply_to_sample(sample)
+                applied_transforms_so_far.append(t)
+            else:
+                additional_samples = [load_sample_fn() for _ in range(t.additional_samples_count)]
```
shaydeci, line 106: From what I understand load_sample_fn should point to the dataset's load_sample method, correct?
In that case load_sample expects an independent as an argument (from what I saw at super_gradients.training.datasets.pose_estimation_datasets.base_keypoints.BaseKeypointsDataset.load_sample).
If I am right, does it work like this?
Otherwise maybe lets generate an index?

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
```diff
@@ -0,0 +1,132 @@
...
+                additional_samples = [
+                    cls._apply_transforms(
```
shaydeci, line 108: In object detection we don't apply the transforms we did on the original sample (currently), but I actually like this.

Perhaps, if we want to preserve the behaviour and re-use this code, can we add a flag, something like "apply_transforms_on_additional_samples" which can be set to true ?

Also, if we are already applying these on the additinal samples, we can just load them in the beginning and it would perhaps be a bit more clear (this is just a suggestion, your call).

src/super_gradients/training/transforms/keypoints/keypoints_image_standardize.py
```diff
@@ -0,0 +1,31 @@
...
+class KeypointsImageStandardize(AbstractKeypointTransform):
+    """
+    Standardize image pixel values with img/max_value formula.
+
+    :attr max_value: Current maximum value of the image pixels. (usually 255)
+    """
+
+    def __init__(self, max_value: float = 255.0):
+        super().__init__()
+        self.max_value = max_value
```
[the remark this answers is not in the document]
shaydeci: We are not alligned here, but IMO attr: is not instead of Args:, which should be in the classdocstring

src/super_gradients/training/transforms/keypoints/keypoints_mixup.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def apply_to_sample(self, sample: PoseEstimationSample) -> PoseEstimationSample:
+        if random.random() < self.prob:
+            sample = self.apply_mixup(sample, sample.additional_samples[0])
+        return sample
+
+    def apply_mixup(self, sample: PoseEstimationSample, other: PoseEstimationSample) -> PoseEstimationSample:
+        sample.image = (sample.image * 0.5 + other.image * 0.5).astype(sample.image)
+        sample.mask = np.logical_or(sample.mask, other.mask).astype(sample.mask.dtype)
+        sample.joints = np.concatenate([sample.joints, other.joints], axis=0)
```
shaydeci, line 36: If just one of them is None, is the padding really necessary ?
same for L40-L41

src/super_gradients/training/transforms/keypoints/keypoints_mosaic.py
```diff
@@ -0,0 +1,132 @@
...
+from super_gradients.training.samples import PoseEstimationSample
+from .abstract_keypoints_transform import AbstractKeypointTransform
+
+
+@register_transform()
+class KeypointsMosaic(AbstractKeypointTransform):
+    """
+    Mix two samples together.
+
+    :attr prob:            Probability to apply the transform.
```
shaydeci, line 17: Docs are wrong here.

src/super_gradients/training/transforms/keypoints/keypoints_mosaic.py
```diff
@@ -0,0 +1,132 @@
...
+
+        bottom_joints_offset = np.array([0, top_sample_height, 0], dtype=bottom.joints.dtype).reshape((1, 1, 3))
+        bottom_bboxes_offset = np.array([0, top_sample_height, 0, 0], dtype=bottom_bboxes.dtype).reshape((1, 4))
+
+        joints = np.concatenate([top.joints, bottom.joints + bottom_joints_offset], axis=0)
+        bboxes = self._concatenate_arrays(top.bboxes, bottom_bboxes + bottom_bboxes_offset, shape_if_empty=(0, 4))
+
+        is_crowd = np.concatenate([top.is_crowd, bottom.is_crowd], axis=0)
+        areas = self._concatenate_arrays(top.areas, bottom.areas, shape_if_empty=(0,))
+        return PoseEstimationSample(image=image, mask=mask, joints=joints, is_crowd=is_crowd, bboxes=bboxes, areas=areas, additional_samples=None)
```
shaydeci, line 122: So this transform is not deterministic in output size?
I think this should be documented.

Just another note:
I am not gonna be a party pooper, because it is nice work (:
I do think there is value to having deterministic output size while having having the mosaic center being randomly sampled as we do in DetectionMosaic.

tests/unit_tests/transforms_test.py
```diff
@@ -375,6 +380,93 @@ def test_rescale_and_pad_to_size(self):
         padded_area = rescaled_padded_image[:, :, resized_image_shape[1] :]  # Bottom padding area
         self.assertTrue((padded_area == pad_val).all())
 
+    def test_keypoints_brightness_contrast(self):
```
shaydeci, line 383: I think its also important to test them as part of a keypoints compose.
Perhpas we should also check that it works when used within a dataset in an additional test.

shaydeci (changes requested):

src/super_gradients/training/transforms/keypoints/abstract_keypoints_transform.py
```diff
@@ -0,0 +1,63 @@
...
+            image=image,
+            mask=mask,
+            joints=joints,
+            areas=areas,
+            bboxes=bboxes,
+            is_crowd=np.zeros(len(joints)),  # Old style API does not pass is_crowd parameter, so we set it to zeros
+            additional_samples=None,
+        )
+        sample = self.apply_to_sample(sample)
+        return sample.image, sample.mask, sample.joints, sample.areas, sample.bboxes
```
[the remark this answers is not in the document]
BloodAxe: Indeed. This way we allow existing pose estimation dataset to work without issues.

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
```diff
@@ -0,0 +1,132 @@
...
+        :param transforms:         List of keypoint-based transformations
+        :param min_bbox_area:      Minimal bbox area of the pose instances to keep them.
+                                   Default value is 0, which means that all pose instances will be kept.
+        :param min_visible_joints: Minimal number of visible joints to keep the pose instance.
+                                   Default value is 0, which means that all pose instances will be kept.
+        :param load_sample_fn:     A method to load additional samples if needed (for mixup & mosaic augmentations).
+                                   Default value is None, which would raise an error if additional samples are needed.
+        """
+        if load_sample_fn is None and (KeypointsMixup in transforms or KeypointsMosaic in transforms):
+            raise RuntimeError("KeyointsMixup & KeypointsMosaic augmentations require load_sample_fn to be passed")
```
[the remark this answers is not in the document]
BloodAxe: Agree. After some though process I found quite more robust way to check that. 
Fixed  in fa1f79d5750ef37400851cf5833a37450a8d646d

src/super_gradients/training/transforms/keypoints/keypoints_image_standardize.py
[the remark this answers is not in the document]
BloodAxe: Actually it looks like we should not be using :attr and use :param in classdocstring - https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#an-example-class-with-docstrings
Interesing, for some reason I was sure there is such thing as `attr` but apparently there is not.

src/super_gradients/training/transforms/keypoints/keypoints_pad_if_needed.py
BloodAxe: Yeah, unfortunately 3.7 is still there :/

BloodAxe: Well, string enums is still a thing

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        sample = sample.sanitize_sample()
+        sample = self._apply_transforms(
+            sample, transforms=self.transforms, load_sample_fn=self.load_sample_fn, min_bbox_area=self.min_bbox_area, min_visible_joints=self.min_visible_joints
+        )
+        return sample
+
+    @classmethod
+    def _apply_transforms(
+        cls, sample: PoseEstimationSample, transforms: List[AbstractKeypointTransform], load_sample_fn, min_bbox_area, min_visible_joints
```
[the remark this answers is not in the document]
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_mosaic.py
```diff
@@ -0,0 +1,132 @@
...
+
+        sample.joints[:, :, 0] += pad_left
+        sample.joints[:, :, 1] += pad_top
+
+        sample.bboxes[:, 0] += pad_left
+        sample.bboxes[:, 1] += pad_top
+
+        return sample
+
+    def stack_samples_horisontally(self, left, right, pad_from_top):
```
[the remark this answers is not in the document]
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_random_affine.py
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
BloodAxe: In this case we can move Mixup/Mosaic to the top of transforms. I will update the docs to state this more clearly

src/super_gradients/training/transforms/keypoints/keypoints_mosaic.py
BloodAxe: Fixed in c7971002071494f217d38258a9466266aa1ba38f

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
BloodAxe: Fixed in c7971002071494f217d38258a9466266aa1ba38f

src/super_gradients/training/samples/pose_estimation_sample.py
```diff
@@ -0,0 +1,158 @@
...
+    @classmethod
+    def compute_area_of_joints_bounding_box(cls, joints) -> np.ndarray:
+        """
+        Compute area of a bounding box enclosing visible joints for each pose instance.
+        :param joints: np.ndarray of [Num Instances, Num Joints, 3] shape (x,y,visibility)
+        :return:       np.ndarray of [Num Instances] shape with box area of the visible joints
+                       (zero if all joints are not visible or only one joint is visible)
+        """
+        visible_joints = joints[:, :, 2] > 0
+        xmax = np.max(joints[:, :, 0], axis=-1, where=visible_joints, initial=joints[:, :, 0].min())
```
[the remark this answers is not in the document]
BloodAxe: It is needed. When computing the bounding box of visible joints with respect to mask (visible_joints) we have to pass initial value. One can pass some -1_000_000_000 but I though that it would be just a bit nicer not to make any assumptions how big/small values there can be and take the smallest available coordinate.

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
BloodAxe: Initially I did it via `load_sample_fn(index)` but then I though this is not great since we have to pass a dataset size to KeypointsCompose to figure out the range of indexes we can sample from.
Ended up with a argument-free method instead because decided this is too much knowledge exposed to Compose. WDYT?

tests/unit_tests/transforms_test.py
BloodAxe: Fixed in 0d006683020824b6b2e85be7f147019ad9057d4d

BloodAxe: Fixed in 0d006683020824b6b2e85be7f147019ad9057d4d

src/super_gradients/training/transforms/keypoints/keypoints_random_horisontal_flip.py
BloodAxe: Fixed in 88cef73f0a3c26f9e158589c0d36ef541787eaee

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
BloodAxe: So the existing dataset classes for pose estimation will not support the loading of additional samples (This is also helps gently push users to migrate to new sample-centric datasets).
And for new dataset what I did is this:
```
    @abc.abstractmethod
    def load_sample(self, index: int) -> PoseEstimationSample:
        """
        Read a sample from the disk and return a PoseEstimationSample
        :param index: Sample index
        :return: Returns an instance of PoseEstimationSample that holds complete sample (image and annotations)
        """
        raise NotImplementedError()

    def load_random_sample(self):
        num_samples = len(self)
        random_index = random.randrange(0, num_samples)
        return self.load_sample(random_index)

```
To me it's more clear to have two methods other than one with optional parameter, since in latter case we are asking user to not only implement dataset-specific logic of loading data but also a sampling logic which IMO should be common for all dataset classes.

BloodAxe: Marking as resolved after discussing with @shaydeci privately ^_^

shaydeci: How will we pass load_random_sample in Yaml config though ?

shaydeci: Any answer on this? How did you handle this in the yaml file when running ?
It's my last open comment...

BloodAxe: Pardon, completely missed your question ^_^
We don't need to pass KeypointsCompose explicitly in YAML. In fact, KeypointsCompose not even in the registry so it is technically impossible to instantiate it in a YAML. 
It is created internally inside base dataset class and thus able to obtain a reference to `load_random_sample`.

PR #1493, opened 2023-10-03
Feature/sg 1147 notebooks in repo

BloodAxe:
This PR introduce CI job & workflow to run notebooks during integration tests.

CI Config now contains `run_and_convert_notebooks_to_docs` parameter to trigger a workflow manually and it will be executed as part of sanity test & release.

To add new notebooks to be tested, add new notebooks in `Makefile` target `run_and_convert_notebooks_to_docs`

Here is an example of the failing job (Added intentionally outdated notebook) - https://app.circleci.com/pipelines/github/Deci-AI/super-gradients/9479/workflows/6499349e-27f7-47a1-a2eb-ce40578905c2/jobs/28496

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -593,7 +593,7 @@ jobs:
             python -m pip install torch torchvision torchaudio
             python -m pip install pytorch-quantization==2.1.2 --extra-index-url https://pypi.ngc.nvidia.com
             python -m pip install onnx_graphsurgeon==0.3.27   --extra-index-url https://pypi.ngc.nvidia.com
-            
+            python -m ipykernel install --name << parameters.sg_new_env_name >>
             make run_and_convert_notebooks_to_docs
 
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -561,6 +561,49 @@ jobs:
           event: fail
           template: basic_fail_1 # see https://github.com/CircleCI-Public/slack-orb/wiki#templates.
 
+  run_and_convert_notebooks_to_docs:
+    docker:
+      - image: 307629990626.dkr.ecr.us-east-1.amazonaws.com/deci/infra/circleci/runner/sg-gpu:<< pipeline.parameters.sg_docker_version >>
+    resource_class: deci-ai/sg-gpu-on-premise
+    parameters:
+      sg_existing_env_path:
+        type: string
+        default: "/env/persistent_env"
+      sg_new_env_name:
+        type: string
+        default: "${CIRCLE_BUILD_NUM}"
+      sg_new_env_python_version:
+        type: string
+        default: "python3.8"
+    steps:
+      - checkout
+      - run:
+          name: install requirements and execute notebooks
+          command: |
+            << parameters.sg_new_env_python_version >> -m venv << parameters.sg_new_env_name >>
+            source << parameters.sg_new_env_name >>/bin/activate
+            python -m pip install --upgrade setuptools pip wheel
+            python -m pip install -r requirements.txt
+            python -m pip install -r requirements.dev.txt
+            python -m pip install .
+            python -m pip install torch torchvision torchaudio
+            python -m pip install pytorch-quantization==2.1.2 --extra-index-url https://pypi.ngc.nvidia.com
+            python -m pip install onnx_graphsurgeon==0.3.27   --extra-index-url https://pypi.ngc.nvidia.com
+            
+            make run_and_convert_notebooks_to_docs
+
+      - run:
+          name: Remove new environment when failed
+          command: "rm -r << parameters.sg_new_env_name >>"
+          when: on_fail
+
+      - slack/notify:
+          channel: "sg-integration-tests"
+          event: fail
+          template: basic_fail_1 # see https://github.com/CircleCI-Public/slack-orb/wiki#templates.
+
+
+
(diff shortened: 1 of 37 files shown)


Makefile
```diff
@@ -30,6 +30,6 @@ sweeper_test:
 	  exit 1; \
 	fi
 
-
-examples_to_docs:
+# This Makefile target runs notebooks listed below and converts them to markdown files in documentation/source/
+run_and_convert_notebooks_to_docs:
 	jupyter nbconvert --to markdown --output-dir="documentation/source/" --execute src/super_gradients/examples/model_export/models_export.ipynb
```
Louis-Dupont, line 35: There is no other notebook we want to add already ?

Louis-Dupont: Also, I cannot explain why but I got really confused by the name `run_and_convert_notebooks_to_docs`, it took me some time to understand what it means. Maybe my brain saw `run_and_convert` and thought about something like `notebook about running and converting a model`. 

So maybe we could rename it. A few ideas:
- `run_notebooks_and_export_as_md`
- `run_and_save_notebooks_as_md`

BloodAxe: All other notebooks in our repo are outdated 😢 & broken 😞 
And must be updated which I believe should be a separate effort (One PR per notebook)

It is probably a good idea to write some docs on how to write these notebooks and rethink the external data usage. In the nutshell:
* Notebooks MUST NOT require any user action to run them (User should not manually obtain API keys to download dataset, or authorize google drive access). Notebook must be self-contained, download datasets if necessary
* Unless demonstrating a feature that is hardware specific (CUDA AMP for example), notebook must not require a presence of GPU.
* Notebook runtime must not exceed 1h of execution time (The goal of notebooks to demonstrate concepts and not make users wait a day while long-term training finishes)

.circleci/config.yml
```diff
@@ -561,6 +565,49 @@ jobs:
           event: fail
           template: basic_fail_1 # see https://github.com/CircleCI-Public/slack-orb/wiki#templates.
 
+  run_and_convert_notebooks_to_docs:
+    docker:
+      - image: 307629990626.dkr.ecr.us-east-1.amazonaws.com/deci/infra/circleci/runner/sg-gpu:<< pipeline.parameters.sg_docker_version >>
```
shaydeci, line 570: SG docker version would point to something older...
Not sure there is something to do at the moment, and I guess we are sinners for doing it elsewehere as well.
I am just a bit worried that we would actually be using the installed version on our image rather then the updated one since notebooks usually just pip install.

.circleci/config.yml
```diff
@@ -561,6 +565,49 @@ jobs:
...
+          name: install requirements and execute notebooks
+          command: |
+            << parameters.sg_new_env_python_version >> -m venv << parameters.sg_new_env_name >>
+            source << parameters.sg_new_env_name >>/bin/activate
+            python3.8 -m pip install --upgrade setuptools pip wheel
+            python3.8 -m pip install -r requirements.txt
+            python3.8 -m pip install -r requirements.dev.txt
+            python3.8 -m pip install .
+            python3.8 -m pip install torch torchvision torchaudio
+            python3.8 -m pip install pytorch-quantization==2.1.2 --extra-index-url https://pypi.ngc.nvidia.com
```
shaydeci, line 594: Do you think these should already be there?
The notebooks have specifc installations executed in them.

shaydeci (changes requested):

.circleci/config.yml
BloodAxe: Good point. I think we can uninstall the SG and indeed rely on SG installation in the notebook itself. Or use the generic pytorch container image (But I'd prefer not to because of maintenance reasons)

.circleci/config.yml
BloodAxe: That's good point. Probably we should not install anything here and really rely on notebook itself to install what it needs. This way we will ensure it is self-contained.
Ideally we would have to have fresh env per notebook, but that would be quite hard to achieve.

BloodAxe: Ok I think I found the solution.

1) Added a new Makefile target `check_notebooks_version_match` that would go over notebooks we want to export and check whether version of SG installed in them matches with the version of SG defined in __version__ variable (See https://github.com/Deci-AI/super-gradients/blob/b03485c01a071aa0561ea68f09578487ddc83776/tests/verify_notebook_version.py). This check guarantees that we ALWAYS have matching versions between SG.__version__ and notebooks

2) We want to keep `python3.8 -m pip install .` here, since it would allow us to install any feature/tag version of SG. 

3) Now the magic part. So if we have SG.__version__==3.2.1 and our notebook has any other version it will immediately quit with non-zero error code and build will be marked as failed. This would prevent us from releasing a version if there is version mismatch

4) If SG.__version__ matches the one in notebook then notebook executed as usual.

So I think this pretty much solves the problem of outdated notebooks during release. This however does not solve the problem of notebooks execution result not being comited to git but that's a different story.

.circleci/config.yml
BloodAxe: See my comment below, I think this is should kept as is https://github.com/Deci-AI/super-gradients/pull/1493#discussion_r1343988540

.circleci/config.yml
shaydeci: Nice one!

shaydeci (approved):
LGTM

PR #1501, opened 2023-10-05
A call to _process_collate_fn_params was missing in dataloaders.get.

BloodAxe:
Otherwise string collate FN would not be instantiated properly IF dataloader name is None and dataset name is present in dataloader params


Louis-Dupont (approved):
LGTM

PR #1502, opened 2023-10-05
A call to _process_collate_fn_params was missing in dataloaders.get. …

BloodAxe:
…(#1501)

Otherwise string collate FN would not be instantiated properly IF dataloader name is None and dataset name is present in dataloader params

(cherry picked from commit 8637ee7e34e0c98a65b6eace2b302f624d10630a)


Louis-Dupont (approved):
LGTM

PR #1503, opened 2023-10-05
Hotfix/sg 000 fix sigmas

BloodAxe:
Though it is a right thing to do as a standalone PR.

1) Fixes OKS sigmas for COCO dataset in all places
2) Simplify YAML files (Avoid duplication of skeleton definition across multiple files)

Tested that integration accuracy test for DEKR is still passing (If the fixes sigma did affect the score the change was less than 0.05)


Louis-Dupont (approved):
LGTM

PR #1511, opened 2023-10-09
Feature/sg 1144 new export in quantization

BloodAxe:
This PR adds support to export YoloNAS/PPYoloE/YoloX models (That implements ExportableObjectDetectionModel) while using PTQ/QAT models.
Tested both on YoloNAS-S

Side note:
Currently we export model with `preprocessing=False` (Because we calibrated model using validation dataloader that already performed normalization) and `postprocessing=True` because users interested in end-to-end model.
However, maybe it make sense to introduce `export_params` section to give users mode control on export process (Choose max number of detections, batch/flat format, TRT or ONNX nms)


shaydeci (approved):
LGTM

PR #1517, opened 2023-10-10
ExtremeBatchCaseVisualizationCallback now has access to additional_batch_items

BloodAxe:
# What

ExtremeBatchCaseVisualizationCallback will also store additional_batch_items for extreme batch.

# Why

For pose estimation we cannot rely on `targets` to visualize groundtruth (DEKR & YoloNAS has completely different targets format). In contrast we can easily get groundtruth data directly from `additional_batch_items`. 

# How

Since we also don't want to rely on specific keys from this dictionary, I thought it make sense to store `additional_batch_items` as is in the `context`. This way we can refer to the `additional_batch_items` as a whole and also not break existing metrics that rely on upacking this dict into context.

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -474,7 +474,9 @@ class Trainer:
                 if self.pre_prediction_callback is not None:
                     inputs, targets = self.pre_prediction_callback(inputs, targets, batch_idx)
 
-                context.update_context(batch_idx=batch_idx, inputs=inputs, target=targets, **additional_batch_items)
+                context.update_context(
+                    batch_idx=batch_idx, inputs=inputs, target=targets, additional_batch_items=additional_batch_items, **additional_batch_items
+                )
                 self.phase_callback_handler.on_train_batch_start(context)
 
                 # AUTOCAST IS ENABLED ONLY IF self.training_params.mixed_precision - IF enabled=False AUTOCAST HAS NO EFFECT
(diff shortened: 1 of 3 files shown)


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -474,7 +474,9 @@ def _train_epoch(self, context: PhaseContext, silent_mode: bool = False) -> tupl
                 if self.pre_prediction_callback is not None:
                     inputs, targets = self.pre_prediction_callback(inputs, targets, batch_idx)
 
-                context.update_context(batch_idx=batch_idx, inputs=inputs, target=targets, **additional_batch_items)
+                context.update_context(
+                    batch_idx=batch_idx, inputs=inputs, target=targets, additional_batch_items=additional_batch_items, **additional_batch_items
+                )
                 self.phase_callback_handler.on_train_batch_start(context)
```
Louis-Dupont, line 480: I'm not 100% in love with the idea of having a magical dict (`additional_batch_items `) that is passed throughout the context but I do agree it is much cleaner than the current logic where the `**additional_batch_items` are all considered like class attributes, without any explicitness.
And technically, this is what we already do with the `sg_trainer_utils.unpack_batch_items` and the collate functions, so it's not really an issue.

So I'm good with what you propose, and I do think we should stick to it, but an alternative could be to introduce a dataclass `AdditionalBatchItems` which would be different for different collate function. In this case, we would make the contract much more explicit. But that's extra work and not really in the scope of this PR

Louis-Dupont (approved):
LGTM, I just left some thought but I think we should just keep as you did

src/super_gradients/training/sg_trainer/sg_trainer.py
BloodAxe: I generally agree. Maybe at SG 4.0 we rewrite that part completely 💅

PR #1530, opened 2023-10-16
Fixed a bug in YoloNASPose.export() that prevented to export model for BS>1

BloodAxe:


src/super_gradients/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "3.3.0"
+__version__ = "3.3.1"
```
Louis-Dupont, line 1: Version change because you want to release a fix already right ?

Louis-Dupont (commented):
LGTM

src/super_gradients/__init__.py
BloodAxe: Up!

Louis-Dupont (approved):
LGTM

PR #1531, opened 2023-10-16
Deprecated sg_logger_params

shaydeci:


README.md
```diff
@@ -503,15 +503,18 @@ trainer = Trainer("experiment_name")
...
-                   "sg_logger": "dagshub_sg_logger",  # DagsHub Logger, see class super_gradients.common.sg_loggers.dagshub_sg_logger.DagsHubSGLogger for details
-                   "sg_logger_params":  # Params that will be passes to __init__ of the logger super_gradients.common.sg_loggers.dagshub_sg_logger.DagsHubSGLogger
-                     {
-                       "dagshub_repository": "<REPO_OWNER>/<REPO_NAME>", # Optional: Your DagsHub project name, consisting of the owner name, followed by '/', and the repo name. If this is left empty, you'll be prompted in your run to fill it in manually.
-                       "log_mlflow_only": False, # Optional: Change to true to bypass logging to DVC, and log all artifacts only to MLflow  
-                       "save_checkpoints_remote": True,
-                       "save_tensorboard_remote": True,
-                       "save_logs_remote": True,
-                     }
+                   "sg_logger": {"dagshub_sg_logger":
```
BloodAxe, line 506: Should we also introduce logger names consistent with class names and deprecate old snake_case names? E.g "DagsHubSGLogger" instead of "dagshub_sg_logger"?

shaydeci: I was thinking about that too actually.
Should I? Then since we don't want to break (not actually breaking since this is a new format) we would also deprecate the old way?
Or should we just leave both options ?

src/super_gradients/recipes/training_hyperparams/default_train_params.yaml
```diff
@@ -109,15 +109,15 @@ max_train_batches:  # For debug- when not None- will break out of inner train lo
...
-sg_logger_params:
-  tb_files_user_prompt: False # Asks User for Tensorboard Deletion Prompt
-  launch_tensorboard: False
-  tensorboard_port:
-  save_checkpoints_remote: False  # upload checkpoint files to s3
-  save_tensorboard_remote: False  # upload tensorboard files to s3
-  save_logs_remote: False  # upload log files to s3
-  monitor_system: True  # Monitor and write to tensorboard the system statistics, such as CPU usage, GPU, ...
+sg_logger:
+  base_sg_logger:
```
BloodAxe, line 113: How would one "unset" the `base_sg_logger` once it is already in defaults?

I feel this may be the problem if someone would like to use non-default logger:

```yaml
defaults:
  training_hyperparams: default_train_params

training_hyperparams:
  sg_logger:
    wandb_sg_logger:
      ...
```

In this case the resolved config would include:
```
training_hyperparams:
  sg_logger:
    base_sg_logger:
      ...
    wandb_sg_logger:
      ...
```

And that would lead to an exception in factory.

Second use case that would not work is using W&B via Makefile (To keep YAML files intact):

```
WANDB_PARAMS = training_hyperparams.sg_logger=wandb_sg_logger +training_hyperparams.sg_logger_params.api_server=https://wandb.research.deci.ai +training_hyperparams.sg_logger_params.entity=super-gradients training_hyperparams.sg_logger_params.launch_tensorboard=false training_hyperparams.sg_logger_params.monitor_system=true +training_hyperparams.sg_logger_params.project_name=PoseEstimation training_hyperparams.sg_logger_params.save_checkpoints_remote=true training_hyperparams.sg_logger_params.save_logs_remote=true training_hyperparams.sg_logger_params.save_tensorboard_remote=false training_hyperparams.sg_logger_params.tb_files_user_prompt=false

coco2017_yolo_nas_s:
	python src/super_gradients/train_from_recipe.py --config-name=coco2017_yolo_nas_s $(WANDB_PARAMS)

```

README.md
Louis-Dupont: I think it would be great, add the class names, and deprecate the old names.

src/super_gradients/recipes/training_hyperparams/default_train_params.yaml
shaydeci: Strong point right there...
I will think of something.

PR #1533, opened 2023-10-16
Ensure that `checkpoint_num_classes` is propagated from YAML to model

BloodAxe:
Fixes a bug that `checkpoint_num_classes` cannot be set through YAML file:

Use case: Downloaded pre-trained checkpoint from sghub locally and want to train a model with 1 class instead of 80. Since it's local checkpoint, not using `pretrain_weights` but `checkpoint_path`. In this case must specify `checkpoint_num_classes` but it is not propagated properly.

```
checkpoint_params:
 checkpoint_num_classes: 80
 num_classes: 1
 checkpoint_path: /yolo_nas_m_coco.pth
 strict_load: key_matching
```

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -380,6 +380,7 @@ class Trainer:
             pretrained_weights=cfg.checkpoint_params.pretrained_weights,
             checkpoint_path=cfg.checkpoint_params.checkpoint_path,
             load_backbone=cfg.checkpoint_params.load_backbone,
+            checkpoint_num_classes=cfg.checkpoint_params.checkpoint_num_classes,
         )
 
         # TEST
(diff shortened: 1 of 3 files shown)


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -380,6 +380,7 @@ def evaluate_from_recipe(cls, cfg: DictConfig) -> Tuple[nn.Module, Tuple]:
             pretrained_weights=cfg.checkpoint_params.pretrained_weights,
             checkpoint_path=cfg.checkpoint_params.checkpoint_path,
             load_backbone=cfg.checkpoint_params.load_backbone,
+            checkpoint_num_classes=cfg.checkpoint_params.checkpoint_num_classes,
```
Louis-Dupont, line 383: I know this would work on our recipes, but what if someone uses an older recipe that doesnt inherit from `default_checkpoint_params` ?
I think this would break, and it that case we should use `get_params` for more safety.

Louis-Dupont (commented):
See comment

src/super_gradients/training/sg_trainer/sg_trainer.py
BloodAxe: Fair point!

BloodAxe: Fixed in 71ab980650db223e428f8cf91a07d5d1a645502d

Louis-Dupont (approved):
LGTM

PR #1536, opened 2023-10-16
Check class_id validity in DetectionDataset

Louis-Dupont:
2 checks were added to make sure class_id is in the range of (0, num_classes-1)
- `DetectionDataset`: added a check when loading annotations and raise an exception if this happens
- `YoloDarknetDataset`: add a check when parsing.


src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -293,6 +295,18 @@ def _load_all_annotations(self, n_samples: int) -> Tuple[Dict[int, Dict[str, Any
 
         return non_empty_annotations, empty_annotations
 
+    def _validate_annotation(self, annotation: dict):
```
BloodAxe, line 298: The benefit of having a method in base class is not immediately obvious to me. 
To start, this method inside relies on another abstract method `_get_class_index_in_target` to get the class from untyped `annotation` dict. To me it looks unnecessary complicated. Yes, it may be used in child classes, but currently this PR covers only one class (YoloFormatDetection).

If you put all the logic inside YoloFormatDetection class it would make `_validate_annotation` completely redundant:

```
label_id, cx, cw, w, h = int(label_id), float(cx), float(cw), float(w), float(h)

if label_id not in all_classes_list:
  raise ValueError(f"`class_id={label_id}` invalid. It should be in {all_classes_list}.")
```

src/super_gradients/training/datasets/detection_datasets/yolo_format_detection.py
```diff
@@ -184,6 +184,7 @@ def _load_annotation(self, sample_id: int) -> dict:
 
         yolo_format_target, invalid_labels = self._parse_yolo_label_file(
             label_file_path=label_path,
+            num_classes=len(self.all_classes_list),
```
BloodAxe, line 187: Here we assume all_classes_list is dense and sequential (E.g there are no gaps: 0, 2, 3, 4).
Is it true for all use cases? Even when we specify class_inclusion_list?

Louis-Dupont: `self.all_classes_list` is the dense list of classes.
If you set `class_inclusion_list`, it will not affect `self.all_classes_list` (it will affect `self.classes` instead)

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
Louis-Dupont: For `YoloDarknetFormatDetectionDataset` we have a logic of ignoring invalid labels when parsing the labels. We don't have it for the base class (we don't parse in the base class).
So, I thought it's fine to have a check in 2 places; when parsing for `YoloDarknetFormatDetectionDataset` and after loading the annotations in `DetectionDataset`.

You think we should only do it for `YoloDarknetFormatDetectionDataset` ?

BloodAxe: I think so. Where's why:
1) In base class you are calling `_validate_annotation` from the `_load_sample_annotation` which on itself called when iterating over the dataset. We want to fail as fast as possible during class creation, and not while iterating over. If we were able to instantiate the dataset it should be already valid. 

2) When you are parsing dataset in child class you have the full context and can raise the most informative warning if some data is incorrect. In case of base class you are dealing with pre-digested `annotation` structure that already lost some bits of information

3) I don't think that validating the correctness of the data format is the responsibility of base class. If we are parsing COCO annotation, the specific class should say whether it is valid or not. Same for Yolo and other formats.

Maybe it is just that our base class for detection it already too complicated and polluted with lot of functions so it is hard to see where the 'base' class ends and other stuff that just 'happen to be there' starts.

Louis-Dupont: Ok, it makes sense!

> Maybe it is just that our base class for detection it already too complicated and polluted with lot of functions so it is hard to see where the 'base' class ends and other stuff that just 'happen to be there' starts.

Yep

PR #1542, opened 2023-10-17
Fix training_params deprecation

Louis-Dupont:
### Reproduce

```python
from omegaconf import OmegaConf
from hydra.utils import instantiate

from super_gradients import Trainer
from super_gradients.common.environment.cfg_utils import load_recipe
from super_gradients.training.dataloaders import coco2017_train, coco2017_val

train_loader = coco2017_train(dataset_params={"max_num_samples": 400})
val_loader = coco2017_val(dataset_params={"max_num_samples": 400})

yolonas_recipe = load_recipe(config_name="coco2017_yolo_nas_s", overrides=[f"arch_params.num_classes=80", "training_hyperparams.max_epochs=1", "training_hyperparams.mixed_precision=False"])
yolonas_recipe = OmegaConf.to_container(instantiate(yolonas_recipe))
training_params = yolonas_recipe['training_hyperparams']

trainer = Trainer(experiment_name="yolonas_training_with_adapter", ckpt_root_dir="./")
model = models.get(model_name=Models.YOLO_NAS_S, num_classes=80, pretrained_weights="coco")

trainer.train(model=model, training_params=training_params, train_loader=train_loader, valid_loader=val_loader)
```

### Explanation
It looks like `criterion_params` still exists with this code, but is empty. this may be due to default in yamls.
Instead of checking if `criterion_params` exists, I changed the code to check if `criterion_params` is empty


src/super_gradients/common/deprecate.py
```diff
@@ -112,7 +112,7 @@ def override(self, **entries):
 
     def decorator(func):
         def wrapper(*args, **training_params):
-            if deprecated_tparam_name in training_params:
+            if training_params.get(deprecated_tparam_name):
```
BloodAxe, line 115: Should we also check whether result of get is empty dict (`{}`)? or this should not happen?

Louis-Dupont: Empty dict can happen but it will also be ignored (`{}` is seen as `False` in `if` statements).
This is precisely what we want to cover. 
if you have `training_params = {..., "loss": {"MyLoss": {"param": "value}}, "criterion_params": {}}` Then `criterion_params` is meaningless.

BloodAxe (approved):
LGTM

PR #1559, opened 2023-10-22
Feature/sg 1053 add explanations on experiment managing

Louis-Dupont:
Adding a tutorial to explain more in-depth how experiments are handled.
This information was already available here and there but we did not have a centralized tutorial focused on it


BloodAxe (approved):
LGTM

PR #1560, opened 2023-10-23
Added recipe unifying script

shaydeci:
Usage:
    python merge_recipe --config_name=cifar10_resnet -> saves cifar10_resnet_complete.yaml in path/to/super_gradients/recipes

    python merge_recipe --config_dir=/path/to/recipes/ config_name=my_recipe.yaml -> saves config_name_complete.yaml in /path/to/recipes/

    python merge_recipe --config_dir=/path/to/recipes/ config_name=my_recipe.yaml save_path=/other/recipes/dir/my_complete_recipe.yaml
     -> saves the complete receip in /other/recipes/dir/my_complete_recipe.yaml

### DIFF
--- a/src/super_gradients/scripts/merge_recipe.py
+++ b/src/super_gradients/scripts/merge_recipe.py
@@ -1,0 +1,47 @@
+"""
+Script that saves a merged (i.e no inheritance from other yaml configuration files),
+ .yaml file that can be ran on its own without the need to keep other configurations which the original
+  file inherits from.
+
+Usage:
+    python merge_recipe --config_name=cifar10_resnet -> saves cifar10_resnet_complete.yaml in path/to/super_gradients/recipes
+
+    python merge_recipe --config_dir=/path/to/recipes/ config_name=my_recipe.yaml -> saves config_name_complete.yaml in /path/to/recipes/
+
+    python merge_recipe --config_dir=/path/to/recipes/ config_name=my_recipe.yaml save_path=/other/recipes/dir/my_complete_recipe.yaml
+     -> saves the complete receip in /other/recipes/dir/my_complete_recipe.yaml
+
+:arg config_name: The .yaml config filename (can leave the .yaml postfix out, but not mandatory).
+
+:arg config_dir: The config directory path, as absolute file system path.
+ When None, will use SG's recipe directory (i.e path/to/super_gradients/recipes)
+
+:arg: The output path for the complete .yaml file.
+ When None, will use config_dir/config_name with a trailing "_complete.yaml".
+"""
+
+import argparse
+from hydra import initialize_config_dir, compose
+from hydra.core.global_hydra import GlobalHydra
+import pkg_resources
+from omegaconf import OmegaConf
+import os
+from super_gradients import init_trainer
+from super_gradients.common.environment.cfg_utils import normalize_path
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config_dir", type=str, default=pkg_resources.resource_filename("super_gradients.recipes", ""), help="The config directory path")
+    parser.add_argument("--config_name", type=str, help=".yaml filename")
+    parser.add_argument("--save_path", type=str, default=None, help="Destination path to the output .yaml file")
+    args = parser.parse_args()
+    if args.save_path is None:
+        args.save_path = os.path.join(args.config_dir, args.config_name).replace(".yaml", "") + "_complete.yaml"
+
+    # NEED TO REGISTER RESOLVERS FIRST
+    init_trainer()
+    GlobalHydra.instance().clear()
+    with initialize_config_dir(config_dir=normalize_path(args.config_dir), version_base="1.2"):
+        cfg = compose(config_name=args.config_name)
+        OmegaConf.save(config=cfg, f=args.save_path)
+        print(f"Successfully saved recipe at {args.save_path}")


src/super_gradients/scripts/merge_recipe.py
```diff
@@ -0,0 +1,47 @@
+"""
+Script that saves a merged (i.e no inheritance from other yaml configuration files),
+ .yaml file that can be ran on its own without the need to keep other configurations which the original
+  file inherits from.
+
+Usage:
+    python merge_recipe --config_name=cifar10_resnet -> saves cifar10_resnet_complete.yaml in path/to/super_gradients/recipes
```
BloodAxe, line 7: Just a 5 cents regarding naming.

I suggest using `export_recipe` name instead of `merge_recipe`. From my perspective `merge` verb represents the internal implementation details of **what** happens under the hood and not the **user** intention. 

Also, `merge` usually means you have two or more items of the same "quality" to merge (E.g two top-level configs), which is not the case here. What happens is config composition and not merge, strictly speaking.

shaydeci: Was feeling unease with merge TBH.
Changed in d08558ed5b14b7c26502fcb0ffcbdda6afb48ead

BloodAxe:
Maybe add a unit test to export some recipes to ensure everything is working as it should?

src/super_gradients/scripts/merge_recipe.py
```diff
@@ -0,0 +1,47 @@
...
+from super_gradients.common.environment.cfg_utils import normalize_path
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config_dir", type=str, default=pkg_resources.resource_filename("super_gradients.recipes", ""), help="The config directory path")
+    parser.add_argument("--config_name", type=str, help=".yaml filename")
+    parser.add_argument("--save_path", type=str, default=None, help="Destination path to the output .yaml file")
+    args = parser.parse_args()
+    if args.save_path is None:
+        args.save_path = os.path.join(args.config_dir, args.config_name).replace(".yaml", "") + "_complete.yaml"
```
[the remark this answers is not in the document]
Louis-Dupont: > If we don't want to force user providing a --save-path then is is logical to save output file in the cwd (E.g file should be generated at /home/bloodaxe/Desktop) and no in SG recipes dir (Which is hard to find if you don't know where SG is installed).

Agree with `cwd`, and let's make sure to 
1. Let the user know where the file is stored with a nice log
2. If `--save-path` not set, let the user know the output dir can be changed with `--save-path`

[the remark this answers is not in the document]
shaydeci: updated 0463b0cf2fb36de33ba4ce69e7eded1bd22c7974

src/super_gradients/scripts/merge_recipe.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # NEED TO REGISTER RESOLVERS FIRST
+    init_trainer()
```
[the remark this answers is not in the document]
shaydeci: fixed

shaydeci:
> Maybe add a unit test to export some recipes to ensure everything is working as it should?

Moved the logic to a helper method, which is whats tested in the added tests.

Louis-Dupont (commented):
How should we "communicate" about this ?
Ideally, it would be great if we mention this script somewhere, but I'm not sure where.
Maybe in `Recipes_Training.md` or `Recipes_Custom.md` ?

src/super_gradients/scripts/export_recipe.py
```diff
@@ -0,0 +1,34 @@
...
+ When None, will use the current working directory.
+
+"""
+
+import argparse
+from super_gradients.common.environment.cfg_utils import export_recipe
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config_dir", type=str, default=None, help="The config directory path")
```
BloodAxe, line 30: In convert_from_recipe I did a slightly different scheme, where first argument (config) is positional and mandatory. Second argument (output) is also positional but optional:

```
    parser.add_argument("config_name", type=str, help=".yaml filename")
    parser.add_argument("save_path", type=str, default=None, help="Destination path to the output .py file")
    parser.add_argument("--config_dir", type=str, default=pkg_resources.resource_filename("super_gradients.recipes", ""), help="The config directory path")
```

One still can use --config_name syntax, but it also enables use of command as follows: `convert_from_recipe.py coco2017_yolo_nas_s train_coco2017_yolo_nas_s.py` (Without --config_name).

Since second argument is optional, `convert_from_recipe.py coco2017_yolo_nas_s` would also work fine.

I suggest to add these changes to this script to make them consistent for the user.

src/super_gradients/common/environment/cfg_utils.py
```diff
@@ -167,3 +168,29 @@ def override_cfg(cfg: DictConfig, overrides: Union[DictConfig, Dict[str, Any]])
...
+    """
+    if save_path is None:
+        save_path = os.path.join(os.getcwd(), config_name).replace(".yaml", "") + "_complete.yaml"
+    # NEED TO REGISTER RESOLVERS FIRST
+    register_hydra_resolvers()
+    GlobalHydra.instance().clear()
+    with initialize_config_dir(config_dir=normalize_path(config_dir), version_base="1.2"):
+        cfg = compose(config_name=config_name)
+        OmegaConf.save(config=cfg, f=save_path)
+        print(f"Successfully saved recipe at {save_path}. \n" f"Recipe content:\n {cfg}")
```
BloodAxe, line 196: Was print() intended of should it be logger.info?

src/super_gradients/common/environment/cfg_utils.py
```diff
@@ -167,3 +168,29 @@ def override_cfg(cfg: DictConfig, overrides: Union[DictConfig, Dict[str, Any]])
...
+    :param config_name: The .yaml config filename (can leave the .yaml postfix out, but not mandatory).
+
+    :param save_path: The config directory path, as absolute file system path.
+        When None, will use SG's recipe directory (i.e path/to/super_gradients/recipes)
+
+    :param config_dir: The config directory path, as absolute file system path.
+        When None, will use SG's recipe directory (i.e path/to/super_gradients/recipes)
+
+    """
+    if save_path is None:
```
BloodAxe, line 188: Shouldn't we do this default path resolution from the main() function? IMO code of the method would be cleaner a bit if this logic will be in entry point.

shaydeci: 5fa8827578cb6f5a20c38257674c520c7d028d4d updated

src/super_gradients/scripts/export_recipe.py
shaydeci: done f791e29886ed0dffb9b12db6ef0e7bd52ebe996e

src/super_gradients/common/environment/cfg_utils.py
shaydeci: fixed

Louis-Dupont (dismissed):
LGTM

tests/unit_tests/test_export_recipe.py
```diff
@@ -0,0 +1,33 @@
...
+
+            with initialize_config_dir(config_dir=td, version_base="1.2"):
+                cfg = compose(config_name="cifar10_resnet_complete.yaml")
+
+            cfg = hydra.utils.instantiate(cfg)
+
+            self.assertEqual(cfg.training_hyperparams.max_epochs, 250)
+
+            # Clean up the created file after the test
+            os.remove(save_path)
```
BloodAxe, line 29: Not a huge deal, but since you're in TemporaryDirectory - it will be cleaned anyway when leaving `with`

BloodAxe (dismissed):
LGTM

BloodAxe (approved):
LGTM

PR #1562, opened 2023-10-23
Added metrics logging to checkpoint and separate yaml file

hakuryuu96:
### Issue description
@BloodAxe said SG does not support saving metrics, environment, and train config to the checkpoint. 

After getting more familiar with a codebase, I realized the task should be split into several parts / PRs. 

### PR description
This PR proposes a way to add metrics to a checkpoint. 
1. Metrics from train and validation steps are saved in all _.pth_ files.
2. Additionally, the metrics are stored in several _.yml_ files (best, latest, and epoch if specified in config). The file structure is the following:
```yaml
metrics:
  train:
   <metricClass1>: <float> 
   ...
   <metricClassn>: <float> 
  valid:
   <metricClass1>: <float> 
   ...
   <metricClassn>: <float> 
tracked_metric_name: <metricClassx>
```
Primarily, this is done for easier access to experiment metrics, so the user doesn't need to run TB or parse the checkpoint to access them. 

### Some ideas and notes
1. **The metric setup could be done in a bit more flexible way.**
Let's suppose the user needs to train a custom model with several classification heads and track different metrics. For now, all the metrics are gathered in a MetricCollection and each metric uses the same output of a model. Moreover, if the user wants to track the accuracy of some of the outputs, he can't do this. Now metrics are named exactly as the class is called and are stored in the dictionary under this name. It will be rewritten into the dictionary and one metric will remain. 

It would be great to have the possibility to specify which model output should be used for every metric (and maybe loss). It also could be configured somehow like
```yaml
metrics: 
    __target__: CustomMetricDict
        metric_head1:
            __target__: MetrciClass1
                param1: ...
                paramn: ...
           model_prediction_key: <str>
           target_key: <str>
        ...
        metric_headn: ...    
```
2.  **Mixing metrics and losses in a single dict** (sg_trainer.py, _get_train_loop_description_dict_ method, used as a return from _evaluate_)
When someone tries to access metric/loss values separately after executing the _evaluate()_ method, some routine on splitting losses and metrics from single joined dict should be done. In this PR, saving metrics to the dict could be done much easier, if they were separated (for now the metrics are selected from the dictionary by their names and form a new dictionary, see sg_trainer.py/_save_checkpoint/lines 671, 679). Also, treating losses and metrics as separate structures adds some points to the code understanding, as they are logically different ;) 
It looks like changing this will affect a bunch of places in the project, so if code owners respond to this problem, then it should become a separate PR :)
3. **_isinstance(metric, torch.Tensor)_**
There are some places in __save_checkpoint_ function where a type of metric value should be checked wrt its type (values returned in _validation/train_results_dict_). It looks like metrics there sometimes appear to have different types (torch.Tensor(dtype=float) or a usual float). It would be much more convenient to expect a single type of value there, and checks of this kind will disappear and make the code look cleaner 🪄 
For now, this routine has been moved to a separate function.


src/super_gradients/common/sg_loggers/base_sg_logger.py
```diff
@@ -329,6 +330,22 @@ def _save_checkpoint(self, path: str, state_dict: dict) -> None:
...
+        Initially added for saving metrics to yaml to store it in something easily parsable (easier than .pth checkpoints),
+        but who knows what it will be suited for later.
+
+        :param tag:         Identifier of the summary.
+        :param summary_dict:  Checkpoint summary_dict.
+        :param global_step: Epoch number.
+        """
+
+        name = tag + (f"_{global_step}" if global_step is not None else "") + ".yml"
+        with open(os.path.join(self._local_dir, name), "w") as outfile:
```
BloodAxe, line 345: encoding="utf-8", who knows what metric name a user with custom metric name can have

src/super_gradients/common/sg_loggers/base_sg_logger.py
```diff
@@ -329,6 +330,22 @@ def _save_checkpoint(self, path: str, state_dict: dict) -> None:
...
+            yaml.dump(summary_dict, outfile, default_flow_style=False)
+            outfile.close()
```
BloodAxe, line 347: No need to close file since it is already opened using `with` statement which guarantees close to be called.

src/super_gradients/common/sg_loggers/base_sg_logger.py
```diff
@@ -329,6 +330,22 @@ def _save_checkpoint(self, path: str, state_dict: dict) -> None:
...
 
+    @multi_process_safe
+    def add_yaml_summary(self, tag: str, summary_dict: dict, global_step: Optional[int] = None) -> None:
+        """Saves any dict to <experiment_folder>/<tag>.yaml
+        Initially added for saving metrics to yaml to store it in something easily parsable (easier than .pth checkpoints),
+        but who knows what it will be suited for later.
+
+        :param tag:         Identifier of the summary.
+        :param summary_dict:  Checkpoint summary_dict.
+        :param global_step: Epoch number.
```
BloodAxe, line 341: Indentation a bit off

src/super_gradients/common/sg_loggers/base_sg_logger.py
```diff
@@ -329,6 +330,21 @@ def _save_checkpoint(self, path: str, state_dict: dict) -> None:
         if self.save_checkpoints_remote:
             self.model_checkpoints_data_interface.save_remote_checkpoints_file(self.experiment_name, self._local_dir, name)
 
+    @multi_process_safe
+    def add_yaml_summary(self, tag: str, summary_dict: dict, global_step: Optional[int] = None) -> None:
```
BloodAxe, line 334: Perhaps you should also add support for W&B & ClearML & DagsHub loggers to upload this file to their platform as well (See how add_checkpoint is implemented there)

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -637,10 +637,16 @@ def _backward_step(self, loss: torch.Tensor, epoch: int, batch_idx: int, context
             # RUN PHASE CALLBACKS
             self.phase_callback_handler.on_train_batch_gradient_step_end(context)
 
+    def __maybe_get_item_from_tensor(self, value: Union[float, torch.Tensor]) -> float:
+        if isinstance(value, torch.Tensor):
+            return value.item()
+        return value
```
BloodAxe, line 643: I feel this can be simplified to `float(value)`:
```
>>> float(torch.tensor([1]).cuda())
1.0
>>> float(torch.tensor(1).cuda())
1.0

```

tests/unit_tests/test_train_with_torch_scheduler.py
```diff
@@ -13,7 +13,7 @@ def update(self, *args, **kwargs) -> None:
         pass
 
     def compute(self):
-        return 1
```
BloodAxe, line 16: If you do float(value) in __maybe_get_item_from_tensor you won't need this patch

src/super_gradients/common/sg_loggers/base_sg_logger.py
hakuryuu96: It seems, that the part of the code repeats in all those loggers + the logic of adding checkpoints repeats somehow, so I guess it should be refactored a bit. Please, let me do it in a separate pull request.

tests/unit_tests/test_train_with_torch_scheduler.py
hakuryuu96: Actually, I haven't found any place where the metric returns an integer value, so I decided to change it. It doesn't fail any tests and will be suitable when the issue with converting metric value to float is solved.

hakuryuu96:
After a discussion with @BloodAxe we decided to leave only saving metrics to a checkpoint for this pull request. Another pull request will come with propositions for sg_loggers refactoring and saving some trainer internals to yaml.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -655,10 +656,23 @@ def _save_checkpoint(
 
         # COMPUTE THE CURRENT metric
         # IF idx IS A LIST - SUM ALL THE VALUES STORED IN THE LIST'S INDICES
-        metric = validation_results_dict[self.metric_to_watch]
+        curr_tracked_metric = validation_results_dict[self.metric_to_watch]
```
BloodAxe, line 659: Let's do float(validation_results_dict[self.metric_to_watch])
This way both curr_tracked_metric &  all_metrics are already contain only floats. The idea is to cast it as early as possible

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -655,10 +656,23 @@ def _save_checkpoint(
...
+
+        # create metrics dict to save
+        valid_metrics_titles = get_metrics_titles(self.valid_metrics)
+
+        all_metrics = {
+            "tracked_metric_name": self.metric_to_watch,
+            "metrics": {"valid": {metric_name: float(validation_results_dict[metric_name]) for metric_name in valid_metrics_titles}},
+        }
+
+        train_metrics_titles = get_metrics_titles(self.train_metrics)
```
BloodAxe, line 669: Perhaps this line can be moved under `if train_metrics_dict is not None:`

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -655,10 +656,23 @@ def _save_checkpoint(
...
+
+        if train_metrics_dict is not None:
+            all_metrics["metrics"]["train"] = {metric_name: float(train_metrics_dict[metric_name]) for metric_name in train_metrics_titles}
 
         # BUILD THE state_dict
-        state = {"net": unwrap_model(self.net).state_dict(), "acc": metric, "epoch": epoch}
+        state = {"net": unwrap_model(self.net).state_dict(), "acc": curr_tracked_metric, "epoch": epoch, "all_metrics": all_metrics}
```
BloodAxe, line 675: The final checkpoint would contain: ["all_metrics"]["metrics"]["train"]
A all_metrics/metrics seems a bit 'extra' no?

BloodAxe (approved):
LGTM

BloodAxe (approved):
LGTM

Louis-Dupont (approved):
LGTM

PR #1567, opened 2023-10-24
Feature/sg 1198 mixed precision automatically changed with warning

Louis-Dupont:
disable `mixed_precision` when working on CPU (instead of crash). But raise a warning in that case to let the user know.

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -2,6 +2,7 @@ import copy
 import inspect
 import os
 import typing
+import warnings
 from copy import deepcopy
 from typing import Union, Tuple, Mapping, Dict, Any, List, Optional
 
@@ -1331,7 +1332,7 @@ class Trainer:
 
         self.pre_prediction_callback = CallbacksFactory().get(self.training_params.pre_prediction_callback)
 
-        self._initialize_mixed_precision(self.training_params.mixed_precision)
+        self.training_params.mixed_precision = self._initialize_mixed_precision(self.training_params.mixed_precision)
 
         self.ckpt_best_name = self.training_params.ckpt_best_name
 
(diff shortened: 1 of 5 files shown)


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1601,11 +1602,16 @@ def _set_test_metrics(self, test_metrics_list):
         self.test_metrics = MetricCollection(test_metrics_list)
 
     def _initialize_mixed_precision(self, mixed_precision_enabled: bool):
+
+        if mixed_precision_enabled and device_config.device == "cpu":
```
BloodAxe, line 1606: Maybe use `if mixed_precision_enabled and not torch.cuda.is_available():` instead? 
Think it's better for two reasons:
1) device_config.device == "cpu" does not account MPS device
2) Even if we use device != "cuda", comparing with string looks unsafe (Could device be "cuda:0"?)

Louis-Dupont: Doesnt `device_config.device == "cpu"` already cover all that ?
I think we should trust the values of `device_config`, otherwise it's weird to use this dataclass anywhere, no ?

That being said, I checked on where we set `device_config.device` and it is true we don't check whether GPU is available... So `device_config.device="cuda"` could technically be set with `setup_device(devuce="cuda")` even when only cpu is available (and leading to a crash later in the code).

Maybe we can change 
``` python
@dataclasses.dataclass
class DeviceConfig:
    device: str = "cuda" if torch.cuda.is_available() else "cpu"
    multi_gpu: str = None
    assigned_rank: int = dataclasses.field(default=_get_assigned_rank(), init=False)
```

To
``` python

@dataclasses.dataclass
class DeviceConfig:
    _device: str = "cuda" if torch.cuda.is_available() else "cpu"
    multi_gpu: str = None
    assigned_rank: int = dataclasses.field(default=_get_assigned_rank(), init=False)

    @property
    def device(self):
        return self._device

    @device.setter
    def device(self, value: str):
        if "cuda" in value and not torch.cuda.is_available():
            raise ValueError("CUDA is not available, cannot set device to cuda")
        self._device = value

    @property
    def is_cuda(self):
        return "cuda" in self._device

```

And then trust the dataclass, and keep `if mixed_precision_enabled and device_config.device == "cpu":`

What do you think?

(side note, why `device_count ` and not `is_available` ? Isnt `is_available` more restrictive and therefore better ?)

BloodAxe: I love device setter approach )

BloodAxe (approved):
LGTM

PR #1569, opened 2023-10-25
Bugfix/sg 1197 criterion params deprecation removal

shaydeci:
The assigner passed through the deprecation decorator for training params caused a key in training params to be DictConfig.

Since we already discussed the problems of having loss orgainzes in factory format, which might lead to further problems I opened this PR to completely go back to the way we used to work with criterion_params and loss.


BloodAxe (approved):
LGTM

PR #1574, opened 2023-10-26
Fixed issue with torch 1.12 issue with arange not supporting fp16 for CPU device.

BloodAxe:
On torch 1.12 arange_cpu is not implemented for fp16
Fix is to create arange with fp32 and then cast to fp16.


Louis-Dupont (approved):
LGTM

PR #1581, opened 2023-10-26
Added packages to pth checkpoint

hakuryuu96:
### Issue description
@BloodAxe said SG does not support saving metrics, environment, and train config to the checkpoint.

After getting more familiar with a codebase, I realized the task should be split into several parts / PRs. This is a second one ;)

### PR description
This PR proposes a way to add an environment/libs version to a separate field in the checkpoint .pth file.


BloodAxe (approved):
LGTM

Louis-Dupont (approved):
LGTM

PR #1583, opened 2023-10-27
Bugfix: Cannot ExtremeBatchPoseEstimationVisualizationCallback when keypoint_colors is not ListConfig

BloodAxe:
Fixed bug in handling keypoint_colors/edge_colors/edge_links argumens in ExtremeBatchPoseEstimationVisualizationCallback that prevented instantiation of this class when params above were not instances of ListConfig type.

Side note: PR into master_330, let's try releasing from this branch


shaydeci (approved):
LGTM

PR #1586, opened 2023-10-30
Updated "What are Recipes and How To Use Them" notebook

BloodAxe:


shaydeci (changes requested):
Looks good. Just two things:
1. Shouldn't we update Makefile to include the notebook docs (run_and_convert_notebooks_to_docs)
2. Please use pip install XXX &> /dev/null or drop output so we don't trash notebook output.

shaydeci (approved):
LGTM

PR #1587, opened 2023-10-30
Feature/sg 1190 transfer learning cls notebook

shaydeci:


shaydeci:
@Louis-Dupont 
@BloodAxe 
Updated pip command with your suggestions

BloodAxe:
Good, but I'm still not fully happy )
```
!!pip install -qq super_gradients==3.3.1
! pip install -qq matplotlib==3.1.3
! python -m pip install --upgrade pillow==9.2.0
```

1) Double-exclamation mark? `!!pip` would work? 
2) We have matplotlib in our requirements. No need to re-install older version (We have matplotlib>=3.3.4)
3) Same with Pillow, it was back in the day until we fixed this Pillow is_directory stuff, so now it should work outside the box

BloodAxe (approved):
LGTM

PR #1592, opened 2023-10-30
Add custom dataset notebook

Louis-Dupont:


BloodAxe:
Pls update the install commands:
```
!pip install data-gradients==0.2.1 gwpy &> /dev/null
!pip install super-gradients==3.3.1 gwpy &> /dev/null
!pip freeze gwpy &> /dev/null
```

to 

```
!pip install -qq super-gradients==3.3.1
```

(No need to specify DG version IMO since it's already in requirements)

Louis-Dupont:
Changing `pip install ...` [873d4e9](https://github.com/Deci-AI/super-gradients/pull/1592/commits/873d4e93a95745f16246ac7b30ef638e63890cf3)

BloodAxe:
Just a minor comment on the notebook - we don't need `prettyformatter` as `pprint` comes with a standard python 3.x (from pprint import pprint) so this dependency can be dropped from pip install list.

BloodAxe (approved):
LGTM

PR #1598, opened 2023-11-01
Fixed bug in ModelWeightAveraging class that led to corrupted model when metric to watch was NaN/Inf

BloodAxe:
A class was not checking whether observed metric is a finite number and included a state dict of the model that had inf/nan metric value into averaging. 

The fix was three-liner:
```
val = float(validation_results_dict[self.metric_to_watch])

if not np.isfinite(val):
  return False, None
```


Louis-Dupont (approved):
LGTM, just make sure the `greater_is_better` without default is fine for classification (as raised by ofri)

PR #1599, opened 2023-11-01
Fixed bug in _pad_image that did not support pad_value=(R,B,G) input

BloodAxe:


src/super_gradients/training/transforms/utils.py
```diff
@@ -89,20 +91,34 @@ def _get_bottom_right_padding_coordinates(input_shape: Tuple[int, int], output_s
...
+                raise ValueError(f"A pad_value tuple ({pad_value} length should be {num_channels} for an image with {num_channels} channels")
+
+            pad_value = tuple(pad_value)
+
+        padded_channels = []
+        for channel_index, pad_value_channel in enumerate(pad_value):
+            padded_channels.append(np.pad(image[..., channel_index], (pad_h, pad_w), "constant", constant_values=pad_value_channel))
+        return np.stack(padded_channels, axis=-1)
     else:
         return np.pad(image, (pad_h, pad_w), "constant", constant_values=pad_value)
```
Louis-Dupont, line 123:
```python
if isinstance(pad_value, typing.Sized) and len(pad_value) != num_channels:
                raise ValueError(f"A pad_value tuple ({pad_value} length should be {num_channels} for an image with {num_channels} channels")
```
If we check this when `len(image.shape) == 3`, we should also check it for Grayscale images.

Something like that
```python
    else:
        if not isinstance(pad_value, numbers.Number):
            raise ValueError(f"Single channel image (shape={image.shape}) only supports pad_value to be an int")
        return np.pad(image, (pad_h, pad_w), "constant", constant_values=pad_value)
```

src/super_gradients/training/transforms/utils.py
```diff
@@ -1,4 +1,6 @@
-from typing import Tuple
+import numbers
```
Louis-Dupont, line 1: I've never used/seen this lib, are you sure it's standard and installed in all `python >= 3.7` versions ?
It kind of looks like, but i just want to make sure

BloodAxe: Yes) I believe it was introduced since 3.0 so should be safe to use and it's recommended way to check whether value is a number

src/super_gradients/training/transforms/utils.py
BloodAxe: Thanks, fixed in 0ba650f2b2977a617529c584e08aebcae1b9ad07

src/super_gradients/training/transforms/utils.py
```diff
@@ -89,21 +91,45 @@ def _get_bottom_right_padding_coordinates(input_shape: Tuple[int, int], output_s
...
+        if isinstance(pad_value, numbers.Number):
+            pad_value = tuple([pad_value] * num_channels)
+        else:
+            if isinstance(pad_value, typing.Sized) and len(pad_value) != num_channels:
+                raise ValueError(f"A pad_value tuple ({pad_value} length should be {num_channels} for an image with {num_channels} channels")
+
+            pad_value = tuple(pad_value)
+
+        padded_channels = []
+        for channel_index, pad_value_channel in enumerate(pad_value):
```
ofrimasad, line 119: You can actually replace all of this new code with a single line:
```
return np.pad(image, (pad_h, pad_w, (0, 0)), "constant", constant_values=((pad_value, pad_value), (pad_value, pad_value), (0, 0)))
```
You don't even need the `if`
and since this function is going to be called many time, it is better if it is vectorized

ofrimasad (commented):
I think this can be solved in a vectorized way with a cleaned code

src/super_gradients/training/transforms/utils.py
BloodAxe: In short, we can't. Believe me I tried. Maybe I missed something but no combination did not work.

We also can't use cv2.copyMakeBorder since it does not work correct with C>4 (rest of channels padded with 0 instead of pad values)

This works fine if your pad value is scalar. However I was not able to make it work when pad value is array of [C] elements (for example pad_value = (127,128,129).

BloodAxe: @ofrimasad thanks for the tip. Fixed in a382457af54cd0a765087032f7c7c8dcafe1f566

ofrimasad (approved):
LGTM

Issue #1324, opened 2023-07-26
Failure in Training due to Empty Validation Metrics Dictionary when run_validation_freq is greater than 1


AlfredQin:
### Description

When the `run_validation_freq` parameter in `training_params` of `Trainer` is set to a value greater than 1 (e.g., `run_validation_freq: 5`), the training process fails. This appears to be due to an issue with the validation metrics dictionary (`valid_metrics_dict`).

### Details

The `valid_metrics_dict` is initialized as an empty dictionary. However, if the condition `(epoch + 1) < run_validation_freq` is met, the dictionary remains empty and is passed to the `_write_to_disk_operations` function. This results in an error when attempting to access `validation_results_dict[self.metric_to_watch]` in the `_save_checkpoint` function because the `validation_results_dict` is empty.

Here is the problematic code:

```python
# RUN TEST ON VALIDATION SET EVERY self.run_validation_freq EPOCHS
valid_metrics_dict = {}
if (epoch + 1) % self.run_validation_freq == 0:
    ...
    valid_metrics_dict = self._validate_epoch(context=context, silent_mode=silent_mode)
    ...
...
self._write_to_disk_operations(
    train_metrics_dict=train_metrics_dict,
    validation_results_dict=valid_metrics_dict,
    ...
)
...
...
def _write_to_disk_operations(
        self,
        train_metrics_dict: dict,
        validation_results_dict: dict,
        test_metrics_dict: dict,
        lr_dict: dict,
        inf_time: float,
        epoch: int,
        context: PhaseContext,
    ):
     ...
        # SAVE THE CHECKPOINT
        if self.training_params.save_model:
            self._save_checkpoint(self.optimizer, epoch + 1, validation_results_dict, context)
...
def _save_checkpoint(
    self,
    optimizer: torch.optim.Optimizer = None,
    epoch: int = None,
    validation_results_dict: Optional[Dict[str, float]] = None,
    context: PhaseContext = None,
) -> None:
    ...
    metric = validation_results_dict[self.metric_to_watch]


### Versions

Set run_validation_freq > 1 in the training parameters.
Run the training process.
Observe the error when the training tries to save a checkpoint.

BloodAxe:
Thanks for the bug report. We will investigate this behavior

hakuryuu96:
@BloodAxe please, see this [PR](https://github.com/Deci-AI/super-gradients/pull/1601). I've attempted to fix this.

PR #1601, opened 2023-11-01
Way to fix bug with validation frequency

hakuryuu96:
[https://github.com/Deci-AI/super-gradients/issues/1324](url)

This PR addresses the issue above. The behavior of saving checkpoints changes when max_epochs is not divisible by val_frequency. In this case, ckpt_latest won't be checked whether it is the best. A corresponding warning was added.


tests/end_to_end_tests/trainer_test.py
```diff
@@ -93,6 +93,22 @@ def test_checkpoint_content(self):
...
+        trainer.train(
+            model=model, training_params=training_params, train_loader=classification_test_dataloader(), valid_loader=classification_test_dataloader()
+        )
+        ckpt_filename = ["ckpt_best.pth", "ckpt_latest.pth"]
+        ckpt_paths = [os.path.join(trainer.checkpoints_dir_path, suf) for suf in ckpt_filename]
+        metrics = {}
+        for ckpt_path in ckpt_paths:
+            ckpt = torch.load(ckpt_path)
+            metrics[ckpt_path] = ckpt["metrics"]
+        assert metrics[ckpt_paths[0]] == metrics[ckpt_paths[1]]
```
Louis-Dupont, line 110: We use `self.assertEqual` instead of `assert`

Louis-Dupont (commented):
Looks good, just very small note in the test

hakuryuu96:
After a discussion with @BloodAxe we decided:
1) to validate the latest epoch
2) to validate the epochs in _save_ckpt_list_ even they are not divisible by _val_freq_

Corresponding changes have been added in last commit.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -756,6 +762,13 @@ def _init_arch_params(self) -> None:
         if arch_params is not None:
             self.arch_params.override(**arch_params.to_dict())
 
+    def _if_need_to_calc_validation(self, epoch: int) -> bool:
```
BloodAxe, line 765: May I suggest a (hopefully) better name for this function: `_should_run_validation_for_epoch`?

BloodAxe, line 765: A docstring for this method should explicitly indicate whether epoch is zero-based or starts with 1.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -756,6 +762,13 @@ def _init_arch_params(self) -> None:
...
+        is_run_val_freq_divisible = not bool((epoch + 1) % self.run_validation_freq)
```
BloodAxe, line 766: Can be simplified to `(epoch % self.run_validation_freq) == 0`

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -656,17 +656,13 @@ def _save_checkpoint(
...
-        # COMPUTE THE CURRENT metric
-        # IF idx IS A LIST - SUM ALL THE VALUES STORED IN THE LIST'S INDICES
-        curr_tracked_metric = float(validation_results_dict[self.metric_to_watch])
+        # Check whether we have to attempt the validation results
+        # and subtract 1 from epoch because epoch+1 is passed to this function
+        is_validation_calculated = self._if_need_to_calc_validation(epoch - 1)
+        curr_tracked_metric = None
 
-        # create metrics dict to save
-        valid_metrics_titles = get_metrics_titles(self.valid_metrics)
```
BloodAxe, line 664: I think this is definitely not the right place where to put this logic TBH.
I think add_checkpoint must be always called when metrics are present. IMO it violates SRP that `add_checkpoint ` must know of such think as validation frequency. 

Instead, the condition we use to trigger `_test_epoch` should be changed instead:

`if (epoch + 1) % self.run_validation_freq== 0` -> `if self._should_run_validation_for_epoch(epoch)` should be used.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1487,13 +1509,20 @@ def forward(self, inputs, targets):
...
+                self._write_scalars_to_logger(
+                    metrics=train_metrics_dict, epoch_values=self._epoch_start_logging_values, epoch=1 + epoch, inference_time=train_inf_time, tag="Train"
+                )
+
                 # RUN TEST ON VALIDATION SET EVERY self.run_validation_freq EPOCHS
                 valid_metrics_dict = {}
-                if (epoch + 1) % self.run_validation_freq == 0:
+                is_validation_calculated = self._should_run_validation_for_epoch(epoch)
+
+                if is_validation_calculated:
```
BloodAxe, line 1521: I think it's safe to inline the call as follows:

```
if self._should_run_validation_for_epoch(epoch):
   ...
```

IMO it's a bit clearer and does not introduce temp variable `is_validation_calculated`

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1521,20 +1555,21 @@ def forward(self, inputs, targets):
...
-                    self._write_to_disk_operations(
-                        train_metrics_dict=train_metrics_dict,
-                        validation_results_dict=valid_metrics_dict,
-                        test_metrics_dict=test_metrics_dict,
-                        lr_dict=self._epoch_start_logging_values,
-                        inf_time=inf_time,
-                        epoch=epoch,
-                        context=context,
-                    )
+                    if is_validation_calculated and self.training_params.save_model:
```
BloodAxe, line 1565: Same here https://github.com/Deci-AI/super-gradients/pull/1601/files#r1382418638

src/super_gradients/training/sg_trainer/sg_trainer.py
hakuryuu96: I assumed we need to check this condition several times, so we can store the result in a variable and not call the method every time. This way the code looks more compact and, from my point of view, does not harm the readability.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1521,20 +1555,21 @@ def forward(self, inputs, targets):
                     context.update_context(metrics_dict=test_metrics_dict)
                     self.phase_callback_handler.on_test_loader_end(context)
 
+                    self._write_scalars_to_logger(metrics=test_metrics_dict, epoch_values=None, epoch=1 + epoch, inference_time=test_inf_time, tag="Test")
```
BloodAxe, line 1558: Feels like we should simply inline _write_scalars_to_logger method as all it does - calling a logger.add_scalars a couple of times here. @shaydeci maybe you have better context on how sg_logger evolved over time, WDYT the suggested change?

BloodAxe (approved):
LGTM

PR #1602, opened 2023-11-02
Fixed ListConfig in pose estimation dataset classes

BloodAxe:
This PR fixes leakage of ListConfig (edge_links) variable that holds a reference to the whole config that that includes instances of dataset, collate fn and even COCO object stored in checkpoint file.


src/super_gradients/common/sg_loggers/base_sg_logger.py
```diff
@@ -348,3 +352,27 @@ def _save_code(self):
...
+        :param state_dict:  Checkpoint state_dict.
+        :param path:        Indicates the current path of the checkpoint
+                            (Used to print meaningful path if problematic key is detected)
+        """
+        if isinstance(state_dict, (ListConfig, DictConfig)):
+            raise ValueError(
+                f"Checkpoint state_dict element {path} contain ListCongfig and DictConfig."
+                f"Only types and Tensors are supported."
+                f"Most likely, you forgot to convert those to container using OmegaConf.to_container()."
+            )
```
Louis-Dupont, line 370: Do we really want to raise an error ?
It would suck if you train a model just to get an error when saving your checkpoint.
This also means that if we somehow introduce a bug in SG (or in a client delivery) where we save Listconfig, the model wouldn't be able to train at all without fixing.
Maybe we want this behavior because people often don't read warnings, but I think it's worth thinking twice about it, I dont know if you thought about that

src/super_gradients/training/datasets/pose_estimation_datasets/abstract_pose_estimation_dataset.py
```diff
@@ -50,6 +51,18 @@ def __init__(
             load_sample_fn=self.load_random_sample,
         )
         self.num_joints = num_joints
+
+        # Explicitly convert edge_links, keypoint_colors and edge_colors to lists of tuples
+        # This is necessary to ensure ListConfig objects do not leak to these properties
+        # and from there - to checkpoint's state_dict.
+        # Otherwise, through ListConfig instances a whole configuration file will leak to state_dict
+        # and torch.load will attempt to unpickle lot of unnecessary classes.
+        edge_links = [(int(from_idx), int(to_idx)) for from_idx, to_idx in edge_links]
```
Louis-Dupont, line 60: `OmegaConf.to_container()` cannot do this convertion? 
If not, it could be good to have a function that converts any DictConfig/ListConfig to native python objects, like here where you want `List[Tuple[int, int]]`.

src/super_gradients/common/sg_loggers/base_sg_logger.py
BloodAxe: Fixed in 4fa008e6717e1d0863d481372ef7ab418dd9c2a3

src/super_gradients/training/datasets/pose_estimation_datasets/abstract_pose_estimation_dataset.py
BloodAxe: Actually this snippet solves two issues at the same time:
1) Converts to primitive types 
2) Ensures that edge_links passed in the correct format (E.g array as N,2 shape or something that can be represented as an array of pairs)
So I think here is justified.
Unfortunately OmegaConf.to_container cannot take list or dict as input (It crashes when fed a primitive type)

src/super_gradients/common/sg_loggers/base_sg_logger.py
```diff
@@ -348,3 +351,29 @@ def _save_code(self):
             self.add_file(name)
             code = "\t" + code
             self.add_text(name, code.replace("\n", "  \n  \t"))  # this replacement makes tb format the code as code
+
+    def _sanitize_checkpoint(self, state_dict: dict) -> dict:
```
ofrimasad, line 355: Can we use a softer term, please... 
sanitize sounds like DictConfigs are bacteria.
Maybe,`_convert_to_native`?

ofrimasad (commented):
just a minor naming comment...

src/super_gradients/common/sg_loggers/base_sg_logger.py
BloodAxe: To some extent it is. Naming was intentional to indicate that we do a state dict cleanup to contain primitive types only. Maybe in future we will add even more strict checking here.

BloodAxe: @ofrimasad I mean if it's a show-stopper then I can rename it.
But the word is pretty common in SD https://github.com/search?q=sanitize&type=repositories

ofrimasad (approved):
LGTM

PR #1605, opened 2023-11-02
Add skip_resize for `model.predict`

Louis-Dupont:


src/super_gradients/training/processing/processing.py
```diff
@@ -84,20 +87,44 @@ def infer_image_input_shape(self) -> Optional[Tuple[int, int]]:
...
-    def __init__(self, processings: List[Processing]):
+    def __init__(self, processings: List[Processing], skip_image_resizing: bool = False):
+        """
+        :param processings:     List of Processing objects to compose.
+        :param skip_image_resizing: If True, skip the image resizing processings.
+        """
         self.processings = processings
+        self.include_image_resizing = not skip_image_resizing
+
+    def disable_image_resizing(self):
```
BloodAxe, line 109: I don't like the idea of modifying processing params in-place. I think it is error-prone and could be done differently in a (subjectively) simple manner:

We have list of processing operations. We know the subset of operations that should not be run if skip resizing is enabled: LongestMaxSize + Pad.

So to generate a new processing transforms we should iterate over existing transforms, look for these two transfom classes and exclude then, adding a KeypointsAutoPadding instead.

This operation can happen in get_pipeline method so that logic of the Pipeline class would be unchanged.

Louis-Dupont: Introducing a `get_equivalent...` method  [6cf8cc9](https://github.com/Deci-AI/super-gradients/pull/1605/commits/6cf8cc9e3446e8b27179fb75e4aa550569a4d182)
And moving the skip logic to `_get_pipeline` [2922599](https://github.com/Deci-AI/super-gradients/pull/1605/commits/292259960ecca72a0d5afdc46149ce624e666d24)

src/super_gradients/training/models/classification_models/base_classifer.py
```diff
@@ -47,19 +51,21 @@ def _get_pipeline(self, fuse_model: bool = True) -> ClassificationPipeline:
...
         return pipeline
 
-    def predict(self, images: ImageSource, batch_size: int = 32, fuse_model: bool = True) -> ImagesClassificationPrediction:
+    def predict(self, images: ImageSource, batch_size: int = 32, fuse_model: bool = True, skip_image_resizing: bool = False) -> ImagesClassificationPrediction:
         """Predict an image or a list of images.
 
         :param images:      Images to predict.
         :param batch_size:  Maximum number of images to process at the same time.
         :param fuse_model:  If True, create a copy of the model, and fuse some of its layers to increase performance. This increases memory usage.
+        :param skip_image_resizing: If True, the image processor will not resize the images.
```
BloodAxe, line 60: If it's not supported in classification pipeline, should we maybe not expose it at all?

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -172,9 +175,17 @@ def _get_pipeline(self, iou: Optional[float] = None, conf: Optional[float] = Non
 
         iou = iou or self._default_nms_iou
         conf = conf or self._default_nms_conf
+
+        # Ensure that the image size is divisible by 32.
+        if isinstance(self._image_processor, ComposeProcessing) and skip_image_resizing:
+            image_processor = self._image_processor.get_equivalent_compose_without_resizing()
+            image_processor.processings.append(DetectionAutoPadding(shape_multiple=(32, 32), pad_value=0))
```
BloodAxe, line 182: Two notes:
1) I think DetectionAutoPadding should be the first in the list of transforms (BC there could be some image permute operations at the end)
2) DetectionAutoPadding should be added inside `get_equivalent_compose_without_resizing`

BloodAxe:
Looking good!
I think what we also should add to this PR is updated docs on predict() that shows usage of this feature with a nice large image with a lot of small objects.

src/super_gradients/training/processing/processing.py
```diff
@@ -344,6 +395,120 @@ def _get_padding_params(self, input_shape: Tuple[int, int]) -> PaddingCoordinate
...
+        """
+        :param shape_multiple:  Tuple of (H, W) indicating the height and width multiples to which the input image dimensions will be padded.
+                                For instance, with a value of (32, 40), an input image of size (45, 67) will be padded to (64, 80).
+        :param pad_value:       Value to pad the image with.
+        """
+        self.shape_multiple = shape_multiple
+        self.pad_value = pad_value
+
+    def preprocess_image(self, image: np.ndarray) -> Tuple[np.ndarray, DetectionPadToSizeMetadata]:
+        is_input_channel_first = image.ndim == 3 and image.shape[0] < image.shape[2]
```
BloodAxe, line 410: `image.shape[0] < image.shape[2]` this is quite hacky way to discriminate HWC / CHW layout.
I'm really starting to think we should have some intermediate structure holding this info:
```
@dataclass
class ImageContainer:
  data: Union[np.ndarray, torch.Tensor]
  layout: ImageLayout # HWC, CHW, BHWC, BCHW
  color_format: str # RGB / BGR / GRAY
```

src/super_gradients/training/models/classification_models/base_classifer.py
Louis-Dupont: The problem is that we because of the way we instantiate models (`model.get` which is dynamic, and task agnostic), we don't have a static contract for `model.predict`.
I don't know if I'm clear, but basically in practice we cannot know the parameters available for `model.predict` of a specific model in advance (unless we dig into the class inheritence...).

So I prefered to have an API that is as close as possible for every task, i.e. here also including `skip_image_resizing` for the classification model, and raise an error if passed.
It's a bit dirty I need to say, but at least the user would get a nice error message (`raise ValueError("`skip_image_resizing` is not supported for classification models.")`) instead of `parameter name not valid` (or something like that) which would be super confusing.
Open for discussion if you have a better idea

src/super_gradients/training/models/detection_models/customizable_detector.py
Louis-Dupont: done here [3874c86](https://github.com/Deci-AI/super-gradients/pull/1605/commits/3874c8672d1af4b11403307c69d367348ccad499)

src/super_gradients/training/processing/processing.py
Louis-Dupont: As discussed, I dont need it eventually [3874c86](https://github.com/Deci-AI/super-gradients/pull/1605/commits/3874c8672d1af4b11403307c69d367348ccad499)
ButI agree on the point for future improvement

BloodAxe (approved):
LGTM

PR #1607, opened 2023-11-03
Enforced check of that all notebooks has matching SG version installed

BloodAxe:
This PR fixes a small issue with regex in checking version of SG package (Now both `pip install super-gradients` and `pip install super_gradietns` commands are parsed correctly)
It also simplify Makefile by splitting NOTEBOOKS variable into NOTEBOOKS_TO_CHECK and NOTEBOOKS_TO_RUN
These variables are no multi-line, which makes it easier to merge PRs where notebooks are added.
And lastly, now `check_notebooks_version_match` is a part of breaking change test to enforce our notebooks are up-to-date.


ofrimasad (commented):
@BloodAxe @shaydeci, correct me if I am wrong - this will force us to manually change all the notebooks with every version we release.

BloodAxe:
You are correct. When we want to release, we have to update version in SG notebook and (ideally) run them against this new version.

So the workflow is the following
- when we think we are ready to release (all stuff is already merged):
- update __version__
- pip install -e . (If it's not already)
- make run_and_convert_notebooks

Commit changes to reflect updated docs in documentation

shaydeci (approved):
LGTM

shaydeci (approved):
LGTM

PR #1612, opened 2023-11-06
Allow setting per-layer learning rates"

shaydeci:
- Added support for passing initial_lr as a dictionary.
- Removed the usages of update_param_groups and initialize_param_groups. Affected recipes had equivalent initial_lr mapping added to them (tested).
- For the edge cases of having an instantiated optimiser, I assign names to the parameter_groups and extrac an initial_lr mapping from them so our schedulers can be used.


src/super_gradients/recipes/cityscapes_ddrnet.yaml
```diff
@@ -55,7 +55,17 @@ architecture: ddrnet_23
 
 training_hyperparams:
   max_epochs: 500
-  initial_lr: 0.0075   # batch size 24
+  initial_lr: # batch size 24
+    default: 0.075
```
BloodAxe, line 59: Maybe reverse the mapping to make it shorter? Thinks it's more clear this way:
```
initial_lr:
  default: 0.0075
  head: 0.075
```

BloodAxe, line 59: Second thought. Do we want to use `default` or `_default_` (Similar to `_target_`) ?  
Not leaning towards either option tbh

src/super_gradients/recipes/training_hyperparams/coco2017_yolo_nas_train_params.yaml
```diff
@@ -8,7 +8,10 @@ warmup_initial_lr:  1e-6
 lr_warmup_steps: 1000
 lr_warmup_epochs: 0
 
-initial_lr:  2e-4
+initial_lr:
```
BloodAxe, line 11: If we have a single LR for all model, maybe let's not introduce mapping?

src/super_gradients/training/params.py
```diff
@@ -98,7 +99,7 @@
         # "lr_updates": {"type": "array", "minItems": 1},
         "lr_decay_factor": {"type": "number", "minimum": 0, "maximum": 1},
         "lr_warmup_epochs": {"type": "number", "minimum": 0, "maximum": 10},
-        "initial_lr": {"type": "number", "exclusiveMinimum": 0, "maximum": 10},
+        # "initial_lr": {"type": "number", "exclusiveMinimum": 0, "maximum": 10},
```
BloodAxe, line 102: If you want to support number | Dict[str,number]

```
"anyOf": [
    {
      "type": ["number", "string", "boolean", "null"]
    },
    {
      "type": "object",
      "patternProperties": {
        "^[a-zA-Z0-9_\.]+$": { "type": "number" }
      },
      "additionalProperties": false
    }
  ]
```

src/super_gradients/training/utils/optimizer_utils.py
```diff
@@ -99,11 +103,16 @@ def build_optimizer(net: nn.Module, lr: float, training_params) -> optim.Optimiz
...
     # OPTIMIZER PARAM GROUPS ARE SET USING DEFAULT OR MODEL SPECIFIC INIT
-    if hasattr(net, "initialize_param_groups"):
-        # INITIALIZE_PARAM_GROUPS MUST RETURN A LIST OF DICTS WITH 'named_params' AND OPTIMIZER's ATTRIBUTES PER GROUP
-        net_named_params = net.initialize_param_groups(lr, training_params)
-    else:
-        net_named_params = [{"named_params": net.named_parameters()}]
+    if hasattr(net, "initialize_param_groups") or hasattr(net, "update_param_groups"):
+        warnings.warn(
+            "initialize_param_groups and update_param_groups usages are deprecated since 3.3.2, will be removed in "
+            "3.4 and have no effect. \n "
```
BloodAxe, line 109: 3.4 -> 3.5 (or 3.6?)

src/super_gradients/training/utils/optimizer_utils.py
```diff
@@ -117,6 +126,97 @@ def build_optimizer(net: nn.Module, lr: float, training_params) -> optim.Optimiz
...
+    default_lr = lr_dict.get("default", None)
+    if default_lr is None:
+        raise RuntimeError("When passing initial_lr as dictionary, must pass 'default'.")
+    group_names = set(lr_dict.keys()) - {"default"}
+
+    for group_name in group_names:
+        lr = lr_dict[group_name]
+        named_params = [(name, param) for name, param in model.named_parameters() if name.startswith(group_name)]
+
+        if lr == 0:
```
BloodAxe, line 151: I think this check can be simplified/improved to this:

```
optimizable_params = [(name, param) for name, param in named_params if param.requires_grad and lr > 0]
param_groups.append({"named_params": optimizable_params , "lr": lr, "name": group_name})
```

Why: We have to respect original value of `requires_grad` from the model and not change it. 
Instead we just filter out what should not be optimized

src/super_gradients/training/utils/optimizer_utils.py
```diff
@@ -117,6 +126,97 @@ def build_optimizer(net: nn.Module, lr: float, training_params) -> optim.Optimiz
...
+            for name, param in named_params:
+                param.requires_grad = False  # Freeze the layer
+        else:
+            param_groups.append({"named_params": named_params, "lr": lr, "name": group_name})
+
+    if default_lr != 0:
+        default_named_params = [(name, param) for name, param in model.named_parameters() if all(name.startswith(group) is False for group in group_names)]
```
BloodAxe, line 158: Same here:
```
optimizable_params = [(name, param) for name, param in named_params if param.requires_grad]
```

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -350,12 +363,23 @@ def __init__(
                 f"Warmup steps will be capped to number of steps in epoch to avoid interfering with any pre-epoch LR schedulers."
             )
 
-        lr_warmup_steps = min(lr_warmup_steps, train_loader_len)
-        learning_rates = np.linspace(start=warmup_initial_lr, stop=initial_lr, num=lr_warmup_steps, endpoint=True)
-
-        self.lr = initial_lr
+        if isinstance(initial_lr, float) or isinstance(initial_lr, int):
```
BloodAxe, line 366: `if isinstance(initial_lr, numbers.Number)`

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -236,8 +236,10 @@ class LRCallbackBase(PhaseCallback):
 
     def __init__(self, phase, initial_lr, update_param_groups, train_loader_len, net, training_params, **kwargs):
         super(LRCallbackBase, self).__init__(phase)
+        if not isinstance(initial_lr, dict):
+            initial_lr = {"default": float(initial_lr)}
```
BloodAxe, line 240: I think it make sense to introduce `DEFAULT_LEARNING_RATE_KEY` variable with value of `default` (Or `_default_` see https://github.com/Deci-AI/super-gradients/pull/1612/files#r1386196047)
The goal is to minimize usage of hard-coded string constants

tests/unit_tests/test_lr_assignment.py
```diff
@@ -0,0 +1,103 @@
+import unittest
+
+from super_gradients.common.object_names import Models
+from super_gradients.training.models import LeNet
+from super_gradients.training.utils import HpmStruct
+from super_gradients.training.utils.optimizer_utils import separate_lr_groups
+from super_gradients.training import models
+
+
+class TestSeparateLRGroups(unittest.TestCase):
```
BloodAxe, line 10: Kudos for tests 👍 
I think two test-cases are missing:
1) When some layers of models has requires_grad==False. Test should check that total number of optimizable params is correct
2) Test case to check case when `initial_lr = {default: 1, some_layer: 0}`. In this case also number of optimizable params in all params groups should be correct

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1289,6 +1296,20 @@ def forward(self, inputs, targets):
...
+            self.optimizer = build_optimizer(net=unwrap_model(self.net), lr=self.training_params.initial_lr, training_params=self.training_params)
+        elif isinstance(self.training_params.optimizer, torch.optim.Optimizer):
+            if self.training_params.initial_lr is not None:
+                raise RuntimeError("An instantiated optimizer cannot be passed along initial_lr != None")
+            self.optimizer = self.training_params.optimizer
+
+            # NEED TO EXTRACT INITAL_LR FROM THE OPTIMIZER PARAM GROUPS
+            self.training_params.initial_lr = get_initial_lr_from_optimizer(self.optimizer)
+        else:
+            raise UnsupportedOptimizerFormat()
```
BloodAxe, line 1311: I think it would be very helpful to print some model statistics (similarly to what we print for dataset before we start training):
* Total number and number of optimizable parameters of the model
* A learning rate (If it's a single-value just print this value, if it's per-layer print it in detailed format.


Examples:

```
Model:         YoloNAS_S (34.2M parameters)
Learning rate: 1e-5
```


```
Model:          YoloNAS_S (34.2M parameters, 34.0M optimized)
Learning rates: 
  - default: 1e-5 (31M parameters)
  - head:    1e-4 (3M parameters)
```

src/super_gradients/training/utils/optimizer_utils.py
```diff
@@ -117,6 +126,97 @@ def build_optimizer(net: nn.Module, lr: float, training_params) -> optim.Optimiz
         optimizer_training_params = net_named_params
 
     # CREATE AN OPTIMIZER OBJECT AND INITIALIZE IT
-    optimizer = optimizer_cls(optimizer_training_params, lr=lr, **training_params.optimizer_params)
+    optimizer = optimizer_cls(optimizer_training_params, **training_params.optimizer_params)
+
+    return optimizer
+
+
+def separate_lr_groups(model: nn.Module, lr_dict: Dict[str, float]) -> List[Dict]:
```
BloodAxe, line 134: I just realized this method is missing one important piece of logic which correct handling of weight_decay on bias and normalization layers.

When `training_params.zero_weight_decay_on_bias_and_bn` is enabled, we should create twice the number of parameter groups for each matching prefix. See how it's done in toolbelt https://github.com/BloodAxe/pytorch-toolbelt/blob/develop/pytorch_toolbelt/optimization/functional.py#L69

When we have `lr=dict(default=0.1, head=0.01)`,
we should have `head` and `head_no_wd` parameter groups while the `head_no_wd` should contain bias & normalization params from the head.

BloodAxe (changes requested):
I went pretty brutal on this PR, sorry :)

src/super_gradients/recipes/cityscapes_ddrnet.yaml
shaydeci: I prefer "default" as _default_ might imply this argument is hydra related.

shaydeci: No head module for darned.
Going the other way around would be much longer.

src/super_gradients/training/utils/optimizer_utils.py
shaydeci: No it does not miss that logic, the name is pretty explanatory.
We currently handle weight decay through optimizer_params.
The feature here is to assign different lrs. This is more work which also introduces more breaking changes (moving from optmizer_params.weight_decay etc) which I disagree is a part of this feature.

BloodAxe:
And cherry on top - let's rename the PR title to be "Github release notes"-friendly. Something like: "Allow setting per-layer learning rates"

src/super_gradients/training/utils/optimizer_utils.py
shaydeci: It actually does perserve it.
It seperates the groups, but since we unpack **extra_optim_params we just have two groups with the same name.
I actually think its good that way.

BloodAxe: Alrighty then ^_^

src/super_gradients/recipes/training_hyperparams/coco2017_yolo_nas_train_params.yaml
shaydeci: fixed 966180d62ef7e8c1a5eb1e2a98a2f9a8783c9af3

src/super_gradients/training/params.py
shaydeci: fixed https://github.com/Deci-AI/super-gradients/commit/966180d62ef7e8c1a5eb1e2a98a2f9a8783c9af3

src/super_gradients/training/utils/optimizer_utils.py
shaydeci: fixed https://github.com/Deci-AI/super-gradients/commit/966180d62ef7e8c1a5eb1e2a98a2f9a8783c9af3

tests/unit_tests/test_lr_assignment.py
shaydeci: added https://github.com/Deci-AI/super-gradients/commit/966180d62ef7e8c1a5eb1e2a98a2f9a8783c9af3

src/super_gradients/training/utils/callbacks/callbacks.py
shaydeci: I kind of feel this is a bit of an overkill to have an environment variable to for "default" string...
I will add it if you  insist on it.

BloodAxe: this is not meant to be environment variable , rather a plan variable: `DEFAULT_LEARNING_RATE_KEY = "default"` used solely for purpose of not placing "default" string constant in many places and use `DEFAULT_LEARNING_RATE_KEY` instead in the code (In YAML files is still would be "default" anyway)
No hard feelings in having it thought

BloodAxe:
One last remark from me - please add integration test for any model of your choice where we train a model for one short epoch with backbone lr=0, nech=0.1xLR, head=LR to ensure all works end-to-end
And then - LGTM :)

shaydeci:
> One last remark from me - please add integration test for any model of your choice where we train a model for one short epoch with backbone lr=0, nech=0.1xLR, head=LR to ensure all works end-to-end
> And then - LGTM :)

Adde unit test training with frozen part of the net.

BloodAxe (approved):

PR #1615, opened 2023-11-06
Update README

BloodAxe:


documentation/source/welcome.md
```diff
@@ -33,28 +33,14 @@ Check out our [Quickstart tutorial](QuickstartBasicToolkit.md) to get learn the
 You can also start from our tutorial on [Detection](ObjectDetection.md), [Segmentation](Segmentation.md) or [Pose Estimation](PoseEstimation.md).
  
 ## What's New (v3.1.0)
```
ofrimasad, line 35: The heading still says 3.1.0

README.md
```diff
@@ -76,7 +80,11 @@ model = models.get(Models.YOLO_NAS_M, pretrained_weights="coco")
 <div align="center">
 <img src="https://github.com/Deci-AI/super-gradients/raw/master/documentation/assets/SG_img/Object Detection@2xDark.png" width="800px">
 </div>
-
+#### Pose Estimation
```
ofrimasad, line 83: for some reason, #### does not work.

ofrimasad (commented):
2 minor comments

README.md
```diff
@@ -46,11 +46,15 @@ ________________________________________________________________________________
...
+YOLO-NAS and YOLO-NAS-POSE architectures are out! 
+The new YOLO-NAS delivers state-of-the-art performance with the unparalleled accuracy-speed performance, outperforming other models such as YOLOv5, YOLOv6, YOLOv7 and YOLOv8.
+A YOLO-NAS-POSE model for pose estimation is also available, delivering state-of-the-art accuracy/performance tradeoff.
+
+Check these out here: [YOLO-NAS](YOLONAS.md) & [YOLO-NAS-POSE](YOLONAS-POSE.md).
 
 <div align="center">
-<img src="https://github.com/Deci-AI/super-gradients/raw/master/documentation/source/images/yolo_nas_frontier.png" width="800px">
+<img src="https://github.com/Deci-AI/super-gradients/raw/master/documentation/source/images/yolo_nas_frontier.png" height="600px">
+<img src="https://github.com/Deci-AI/super-gradients/raw/master/documentation/source/images/yolo_nas_pose_frontier_t4.png" height="600px">
```
Louis-Dupont, line 57: The width is different for both images and it looks a bit off. Maybe set `width=` instead of `height=`?

README.md
```diff
@@ -123,6 +132,13 @@ pip install super-gradients
...
 
+__________________________________________________________________________________________________________
+Version 3.4.0 (November 6, 2023)
+
+* [YoloNAS-Pose](YOLONAS-POSE.md) model released - a new frontier in pose estimation
+* Added option to export a recipe to a single YAML file or to a standalone train.py file 
+* Other bugfixes & minor improvements. Full release notes available [here](https://github.com/Deci-AI/super-gradients/releases/tag/3.4.0)
+
 __________________________________________________________________________________________________________
 Version 3.1.3 (July 19, 2023)
```
Louis-Dupont, line 143: It's a little weird to show `3.4.0` and then `3.1.3` and `3.1.1`, there is no clear logic and it may look a bit unprofessional.
Maybe we should just remove `3.1.3` and `3.1.1`, and try to stick to adding the release notes in every the next versions ?

Louis-Dupont (commented):
Small notes

README.md
Louis-Dupont: (see here)
image

README.md
BloodAxe: I'm not a big fan of putting release notes in the README tbh.
Usually a dedicated file is more clean way to keep those, however I don't see why need this at all. Releases page is pretty good at providing clean picture what is new.
If we look at major libraries like pytorch - they don't update their README with each new version.

ofrimasad (approved):
LGTM

PR #1621, opened 2023-11-07
[Feature] lazy implementation of video predictions 

hakuryuu96:
Problem: OOM when predicting detection for a big video
Solution: replace saving each frame in some lists to generator objects and implement lazy prediction pipeline


src/super_gradients/training/utils/media/video.py
```diff
@@ -30,6 +30,15 @@ def load_video(file_path: str, max_frames: Optional[int] = None) -> Tuple[List[n
     return frames, fps
 
 
+def lazy_load_video(file_path: str, max_frames: Optional[int] = None) -> Tuple[Generator, int]:
```
BloodAxe, line 33: Function type annotation does not match real return value (Tuple of 2 vs 3)

src/super_gradients/training/utils/media/video.py
```diff
@@ -75,7 +98,7 @@ def save_video(output_path: str, frames: List[np.ndarray], fps: int) -> None:
     if check_is_gif(output_path):
         save_gif(output_path, frames, fps)
     else:
-        save_mp4(output_path, frames, fps)
+        lazy_save_mp4(output_path, frames, fps)
```
BloodAxe, line 101: I think you can actually modify exissting function and instead of taking List[[np.ndarray] make it taking Iterable[np.ndarray] which would cover both scenarios (lazy & precomputed) predictions

src/super_gradients/training/utils/media/video.py
Louis-Dupont: Also I would replace `Generator` with one of the two
- `Generator[np.ndarray, None, None]` most standard typing for generators, but too verbose IMO
- `Iterator[np.ndarray]` I prefer this. It is mentioned in the python doc so I think it's fair to use https://docs.python.org/3.6/library/typing.html#typing.Generator

src/super_gradients/training/utils/media/video.py
```diff
@@ -137,6 +178,24 @@ def _validate_frames(frames: List[np.ndarray]) -> Tuple[float, float]:
     return max_height, max_width
 
 
+def _validate_frame(frame, control_height, control_width) -> None:
+    """Validate the frames to make sure that every frame has the same size and includes the channel dimension. (i.e. (H, W, C))
+
+    :param frames:  Frames representing the video, each in (H, W, C), RGB. Note that all the frames are expected to have the same shape.
+    :return:        (Height, Weight) of the video.
```
Louis-Dupont, line 185: The function doesnt return anything

src/super_gradients/training/utils/predict/prediction_results.py
```diff
@@ -504,20 +507,21 @@ def save(
 class VideoDetectionPrediction(VideoPredictions):
     """Object wrapping the list of image detection predictions as a Video.
 
-    :attr _images_prediction_lst:   List of the predictions results
+    :attr _images_prediction_gen:   List of the predictions results
     :att fps:                       Frames per second of the video
     """
 
-    _images_prediction_lst: List[ImageDetectionPrediction]
+    _images_prediction_gen: Generator
```
Louis-Dupont, line 514: See prevous comment with Generator typing

src/super_gradients/training/utils/predict/prediction_results.py
```diff
@@ -504,20 +507,21 @@ def save(
...
+    n_frames: int
 
     def draw(
         self,
         box_thickness: int = 2,
         show_confidence: bool = True,
         color_mapping: Optional[List[Tuple[int, int, int]]] = None,
         class_names: Optional[List[str]] = None,
-    ) -> List[np.ndarray]:
+    ) -> Generator:
```
Louis-Dupont, line 524: That's a breaking change because people may have a piece of code like this
``` python
video_frames_with_preds = model.predict('video.mp4').draw()
```

Maybe we should have a method `lazy_draw` (not sure about the name) that would incorporate all the drawing logic and yield frames, use it in `draw`, `show` and `save`, but with `draw` still returning a `List[np.ndarray]`, WDYT?

fmarchenko-altris: The reason for the changes here is that storing the results in a list would become painful if someone would like to detect objects on a longer video. Assuming that the "list" scenario is limited, the idea is to provide a more general approach to operating the predictions, which will become public and provide users more flexibility.

I agree with you, it's definitely a breaking change. It would be great to know @BloodAxe and @shaydeci opinions regarding this.

BloodAxe: My 5 cents is that returning List[np.ndarray] is still a huge risk of OOM. Yes, it is a BC, but perhaps in this particular scenario I tend to introduce it. I think in most cases people still have somthing like
```
for frame in model.predict().draw():
  ... bla bla bla ..
```

hakuryuu96: The reason for the changes here is that storing the results in a list would become painful if someone would like to detect objects on a longer video. Assuming that the "list" scenario is limited, the idea is to provide a more general approach to operating the predictions, which will become public and provide users more flexibility.

I agree with you, it's definitely a breaking change. It would be great to know @BloodAxe and @shaydeci opinions regarding this.

src/super_gradients/training/utils/media/video.py
```diff
@@ -78,64 +116,61 @@ def save_video(output_path: str, frames: List[np.ndarray], fps: int) -> None:
...
+            )
+        _validate_frame(frame, video_height, video_width)
         video_writer.write(cv2.cvtColor(frame, cv2.COLOR_RGB2BGR))
 
     video_writer.release()
 
 
-def _validate_frames(frames: List[np.ndarray]) -> Tuple[float, float]:
-    """Validate the frames to make sure that every frame has the same size and includes the channel dimension. (i.e. (H, W, C))
+def _validate_frame(frame, control_height, control_width) -> None:
```
BloodAxe, line 158: Missing type annotations

BloodAxe:
Minor comments left. LGTM once addressed 💪

src/super_gradients/examples/predict/pose_estimation_predict_video.py
```diff
@@ -0,0 +1,25 @@
...
+model = models.get("yolo_nas_pose_l", pretrained_weights="coco_pose")
+
+# We want to use cuda if available to speed up inference.
+model = model.to("cuda" if torch.cuda.is_available() else "cpu")
+
+video_path = "slickback.mp4"
+
+# Download the video to the local file system.
+with open(video_path, mode="wb") as f:
+    response = requests.get("https://deci-pretrained-models.s3.amazonaws.com/sample_images/slickback.mp4")
```
BloodAxe, line 15: We don't have video at that URL.
So probably it make sense to drop this example snippet

BloodAxe (approved):
LGTM

PR #1624, opened 2023-11-08
remove ast feature_version

Louis-Dupont:
Removing `feature_version` because it's not supported in `python==3.7`


BloodAxe (approved):
LGTM

PR #1635, opened 2023-11-13
Introduce DetectionSample dataclass

BloodAxe:
What this PR does:

1) It introduce new `AbstractDetectionTransform` class with abstract `apply_to_sample` method. This class becomes a new base class for detection transforms.
2) Existing `DetectionTransform` with the same signature is still there and is supported.
3) Base `DetectionDataset` uses DetectionSample for passing into transforms. This happens "under the hood" and from the user perspective nothing has changed. The only possilbe scenario when existing code may break is when one has overriden `apply_transforms` method. In all other scenarios it would work exactly as before.
4) On-the fly conversion from/to old detection sample `dict` to `DetectionSample` is done by `LegacyDetectionTransformMixin`, which is added to all existing transforms.


src/super_gradients/training/samples/detection_sample.py
```diff
@@ -0,0 +1,78 @@
...
+        clamped_boxes[..., [0, 2]] = np.clip(clamped_boxes[..., [0, 2]], 0, image_width - 1)
+        clamped_boxes[..., [1, 3]] = np.clip(clamped_boxes[..., [1, 3]], 0, image_height - 1)
+        clamped_boxes = xyxy_to_xywh(clamped_boxes, image_shape=(image_height, image_width))
+
+        self.bboxes_xywh = clamped_boxes
+        return self
+
+    def filter_by_mask(self, mask: np.ndarray) -> "DetectionSample":
+        """
+        Remove pose instances with respect to given mask.
```
shaydeci, line 53: pose instances -> object instances ?

src/super_gradients/training/samples/detection_sample.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        self.bboxes_xywh = self.bboxes_xywh[mask]
+        self.labels = self.labels[mask]
+        if self.is_crowd is not None:
+            self.is_crowd = self.is_crowd[mask]
+        return self
+
+    def filter_by_bbox_area(self, min_bbox_area: Union[int, float]) -> "DetectionSample":
+        """
+        Remove pose instances that has area of the corresponding bounding box less than a certain threshold.
```
shaydeci, line 71: pose instances -> object instances ?

src/super_gradients/training/transforms/detection/detection_longest_max_size.py
```diff
@@ -0,0 +1,63 @@
...
+import numpy as np
+
+from super_gradients.common.object_names import Transforms, Processings
+from super_gradients.common.registry import register_transform
+from super_gradients.training.samples import DetectionSample
+from .abstract_detection_transform import AbstractDetectionTransform
+
+
+@register_transform(Transforms.DetectionLongestMaxSize)
+class DetectionLongestMaxSize(AbstractDetectionTransform):
```
shaydeci, line 14: Why are some transforms under here while others left in transforms.py?
And why are they inheriting differently from the ones there?

Will it work with our current way of passing samples in detectiondataset ?

shaydeci (changes requested):

src/super_gradients/training/samples/detection_sample.py
BloodAxe: Fixed in e64d3e98e1fed5bcce5074f88845d615737f23d0 thanks!

src/super_gradients/training/transforms/detection/detection_longest_max_size.py
BloodAxe: > Why are some transforms under here while others left in transforms.py?
Existing transforms left in original place to make sure PR diff shows meaningful results. We will move those to individual files eventually, but that will happen later.
However new transforms that I introduced in this PR (why I did that is answered next) are placed straight to a new/correct place.

> And why are they inheriting differently from the ones there?
We introduce new `AbstractDetectionTransform` class with new signature (`apply_to_sample` insteada of `__call__`). To keep backward compatibility, existing transforms also inherit from `LegacyDetectionTransformMixin` which adds support of `__call__` to be compatible with current dataset classes.
However new transformations are not inheriting from LegacyDetectionTransformMixin since we will drop old API anyway so there is no point in inheriting from `LegacyDetectionTransformMixin` for new classes. 
So no, new classes will not work existing dataset.

src/super_gradients/training/samples/detection_sample.py
BloodAxe: Fixed, thanks

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -468,11 +508,24 @@ def apply_transforms(self, sample: Dict[str, Union[np.ndarray, Any]]) -> Dict[st
...
-            sample["additional_samples"] = self._get_additional_inputs_for_transform(transform=transform)
-            sample = transform(sample=sample)
-            sample.pop("additional_samples")  # additional_samples is not useful after the transform
-        return sample
+            detection_sample.additional_samples = [
+                self.__convert_input_dict_to_detection_sample(s) for s in self._get_additional_inputs_for_transform(transform=transform)
+            ]
+            detection_sample = transform.apply_to_sample(sample=detection_sample)
+            detection_sample.additional_samples = None
+            if isinstance(transform, DetectionTargetsFormatTransform):
```
shaydeci, line 522: This is a bit sketchy here...
So if some user for some reason puts target transform in the middle of the pipeline and we just apply it at the end behind the scenes?
Was this intended?

src/super_gradients/training/transforms/detection/abstract_detection_transform.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    def __init__(self, additional_samples_count: int = 0):
+        """
+        :param additional_samples_count: (int) number of samples that must be extra samples from dataset. Default value is 0.
+        """
+        self.additional_samples_count = additional_samples_count
+
+    @abstractmethod
+    def apply_to_sample(self, sample: DetectionSample) -> DetectionSample:
```
shaydeci, line 22: why are we calling apply_to_sample instead of call ?
For backward compatibility (which, is not at such a high risk IMO since theese are only been called from within DetectionDataset..), maybe its better to use __call__() with some processing in the beginning according to instance of sample.

src/super_gradients/training/transforms/transforms.py
```diff
@@ -608,10 +629,37 @@ def __call__(self, sample: dict) -> dict:
...
+            sample.image = img
+            sample.is_crowd = np.array([0] * len(targets) + [1] * len(crowd_targets), dtype=bool)
+            sample.bboxes_xywh = np.concatenate([targets[:, 0:4], crowd_targets[:, 0:4]], axis=0, dtype=sample.bboxes_xywh.dtype)
+            sample.labels = np.concatenate([targets[:, 4], crowd_targets[:, 4]], axis=0, dtype=sample.labels.dtype)
         return sample
 
+    def get_equivalent_preprocessing(self):
+        raise NotImplementedError("get_equivalent_preprocessing is not implemented for non-deterministic transforms.")
+
+    # def __call__(self, sample: dict) -> dict:
```
shaydeci, line 642: Don't forget to remove/de-comment (depends on what we decide regarding __call__(), see earlier comments).

shaydeci (changes requested):

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -457,6 +458,45 @@ def _get_cached_image(self, index: int, cached_image_shape: Tuple[int, int]) ->
...
+            is_crowd=is_crowd,
+            additional_samples=None,
+        )
+
+    def __convert_detection_sample_to_dict(self, detection_sample: DetectionSample, include_crowd_target: bool) -> Dict[str, Union[np.ndarray, Any]]:
+        image = detection_sample.image
+        crowd_mask = detection_sample.is_crowd > 0
+        crowd_labels = detection_sample.labels[crowd_mask]
+        crowd_bboxes_xyxy = detection_sample.bboxes_xyxy[crowd_mask]
+        crowd_target = np.concatenate([crowd_bboxes_xyxy, crowd_labels[..., None]], axis=-1)
```
shaydeci, line 486: This is gonna crash when we there are no crowd bboxes since -
np.concatenate([np.array([]) ,x[...,None]]) have different dims.

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -457,6 +458,45 @@ def _get_cached_image(self, index: int, cached_image_shape: Tuple[int, int]) ->
...
+
+        labels = detection_sample.labels[~crowd_mask]
+        bboxes_xyxy = detection_sample.bboxes_xyxy[~crowd_mask]
+        target = np.concatenate([bboxes_xyxy, labels[..., None]], axis=-1)
```
shaydeci, line 490: Same as previous comment.

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
BloodAxe: It is sketchy, that is for sure. However this DetectionTargetsFormatTransform anyway should go last (It even listed in documentation). So if user mess up - it will be messed up anyway. 
I honestly see no other way in doing this.

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
BloodAxe: The example above should not happen, since we guaranteed that empty boxes has always shape of (0,4) so concat would work in this case.

src/super_gradients/training/models/detection_models/yolo_nas/dfl_heads.py
```diff
@@ -196,24 +196,24 @@ def _init_weights(self):
...
-            reg_distri, cls_logit = getattr(self, f"head{i + 1}")(feat)
-            # cls and reg
-            # Note we don't apply sigmoid on class predictions to ensure good numerical stability at loss computation
-            cls_score_list.append(torch.permute(cls_logit.flatten(2), [0, 2, 1]))
-            reg_distri_list.append(torch.permute(reg_distri.flatten(2), [0, 2, 1]))
-        cls_score_list = torch.cat(cls_score_list, dim=1)
-        reg_distri_list = torch.cat(reg_distri_list, dim=1)
-
-        return cls_score_list, reg_distri_list, anchors, anchor_points, num_anchors_list, stride_tensor
+    # @torch.jit.ignore
```
shaydeci, line 199: Please decomment

shaydeci (approved):
LGTM

PR #1637, opened 2023-11-14
Update predict + detection TL notebooks and update makefile/readme

Louis-Dupont:


BloodAxe:
# detection_transfer_learning.ipynb

1) CHECKPOINT_DIR = '/home/notebook_ckpts/' is absolute path. I suggest to change it to use use relative paths to not assume any specific folder structure  and OS (/home would fail on windows)

2) Notebook emits warning ` Deprecation: original_target_format should be of type ConcatenatedTensorFormat instead of DetectionTargetsFormat.Support for DetectionTargetsFormat will be removed in 3.1`

BloodAxe:
# notebooks/how_to_run_model_predict.ipynb

1) Do we want to add pose estimation predict example?

Louis-Dupont:
> # detection_transfer_learning.ipynb
> 
> 1) CHECKPOINT_DIR = '/home/notebook_ckpts/' is absolute path. I suggest to change it to use use relative paths to not assume any specific folder structure  and OS (/home would fail on windows)
> 
> 2) Notebook emits warning ` Deprecation: original_target_format should be of type ConcatenatedTensorFormat instead of DetectionTargetsFormat.Support for DetectionTargetsFormat will be removed in 3.1`


Done

> # notebooks/how_to_run_model_predict.ipynb
> 1. Do we want to add pose estimation predict example?

Done

BloodAxe (approved):
LGTM

PR #1638, opened 2023-11-14
Add notebook for fine-tuning YoloNAS with QAT

BloodAxe:


shaydeci:
Few comments:
- In section 1. you specify some info on multi_gpu (probably leftovers from the time that we passed it to Trainer ctor). Lets remove it or call setup_device somewhere just for demonstration (I think the rather makes more sense).
- The helper method in 6. is basically Trainer.ptq(..).
- I don't get the need for the call for Trainer.qat() in the last section. If we already called train on the model after replacing its layers with Q/DQ ones this is quite equivalent.
- If we want to just demonstrate that qat can be done with 1 click and keep section 8. lets add some explanation about it and not leave it as "wrapping up"

BloodAxe:
Updated the notebook

src/super_gradients/training/models/detection_models/yolo_nas/dfl_heads.py
```diff
@@ -263,10 +263,10 @@ def out_channels(self):
         return None
 
     def forward(self, feats: Tuple[Tensor]):
-        if self.training:
-            return self.forward_train(feats)
-        else:
-            return self.forward_eval(feats)
+        # if self.training:
```
shaydeci, line 266: Decomment please.

src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -1208,11 +1210,40 @@ def on_validation_loader_end(self, context: PhaseContext) -> None:
             self._reset()
 
     def _gather_extreme_batch_images_and_log(self, context, loader_name: str):
-        images_to_save = self.process_extreme_batch()
-        images_to_save = maybe_all_gather_np_images(images_to_save)
-        if self.max_images > 0:
-            images_to_save = images_to_save[: self.max_images]
+        from pytorch_toolbelt.utils.distributed import all_gather
```
shaydeci, line 1213: pytorch_toolbelt?

shaydeci (changes requested):
Notebook looks great.
There are some debugging leftovers I mentioned in the comments.
Also there's an empty code cell iat the bottom of the notebook.
Apart from those its LGTM on my side.

src/super_gradients/training/utils/callbacks/callbacks.py
BloodAxe: That was leaked by accident from another branch ^_^

src/super_gradients/training/models/detection_models/yolo_nas/dfl_heads.py
BloodAxe: Fixed in 3073d4b95173e24388ed7481fd4a4bd72816b4a8, thanks for spotting.
Not sure how it leaked to this branch, prob. by misclick

shaydeci (approved):
LGTM

PR #1639, opened 2023-11-14
Improve LR and WD assignment message on training start

shaydeci:
Fixed lr assignment prints in the beginning of the training. The problem was, that if we use zero wd on bn and bias we get two groups with the same name.
I also improved he way it looks (formatting etc).
Example:

```
[2023-11-14 13:26:44] INFO - sg_trainer_utils.py - TRAINING PARAMETERS:
    - Mode:                         Single GPU
    - Number of GPUs:               0          (0 available on the machine)
    - Full dataset size:            4          (len(train_set))
    - Batch size per GPU:           4          (batch_size)
    - Batch Accumulate:             1          (batch_accumulate)
    - Total batch size:             4          (num_gpus * batch_size)
    - Effective Batch size:         4          (num_gpus * batch_size * batch_accumulate)
    - Iterations per epoch:         1          (len(train_loader))
    - Gradient updates per epoch:   1          (len(train_loader) / batch_accumulate)
    - Model: LeNet  (62.01K parameters, 62.01K optimized)
    - Learning Rates and Weight Decays:
      - fc3: (850.00 parameters). LR: 0.1 (850.00 parameters).  WD: 0.0, (10.00 parameters), WD: 0.0001, (840.00 parameters). 
      - default: (61.16K parameters). LR: 0.2 (61.16K parameters).  WD: 0.0, (226.00 parameters), WD: 0.0001, (60.93K parameters). 
```


src/super_gradients/training/utils/sg_trainer_utils.py
```diff
@@ -444,13 +444,21 @@ def get_callable_param_names(obj: callable) -> Tuple[str]:
     return tuple(inspect.signature(obj).parameters)
 
 
+def _format_value(value: float) -> str:
+    if value >= 1e6:
+        return f"{value / 1e6:.2f}M"
+    elif value >= 1e3:
+        return f"{value / 1e3:.2f}K"
+    else:
+        return f"{value:.2f}"
```
BloodAxe, line 453: Add cast to int, since number of parameters can't be fractional number

src/super_gradients/training/utils/sg_trainer_utils.py
```diff
@@ -459,27 +467,37 @@ def get_lr_info(model: nn.Module, param_groups: List[Dict[str, Union[str, float,
...
         else:
-            precision_group = max(0, 4 - int(group_params).bit_length())
-            info_str += f"      - {group_name}: {group_lr} ({group_params:.{precision_group}f}M parameters)\n"
+            lr_wd_groups[group_name]["wd_params"][group_wd] += group_params
+
+    for group_name, info in lr_wd_groups.items():
+        all_group_params = info["params"]
+        lr_str = ", ".join([f"LR: {lr_val} ({_format_value(lr_params)} parameters)" for lr_val, lr_params in info["lr_params"].items()]) + ". "
+        wd_str = ", ".join([f"WD: {wd_val}, ({_format_value(wd_params)} parameters)" for wd_val, wd_params in info["wd_params"].items()]) + ". "
+        info_str += f"      - {group_name}: ({_format_value(all_group_params)} parameters). {lr_str} {wd_str}\n"
```
BloodAxe, line 500: Maybe add left padding for group name to maximum length of parameter groups (to make them all aligned vertically during print)?

BloodAxe (commented):
Just a few cosmetic suggestions

BloodAxe (approved):
LGTM

PR #1641, opened 2023-11-14
Added support of multiple test loaders in train_from_config

BloodAxe:
How it supposed to look in YAML file

```
defaults:
  - cifar10_resnet

test_dataloaders:
  cifar10: cifar10_val
  cifar100: cifar100_val

dataset_params:
  test_dataset_params:
    cifar10:
      root: ./data/cifar10
      train: False
      transforms:
        - Resize:
            size: 32
        - ToTensor
        - Normalize:
            mean:
              - 0.4914
              - 0.4822
              - 0.4465
            std:
              - 0.2023
              - 0.1994
              - 0.2010
      target_transform: null
      download: True

    cifar100:
      root: ./data/cifar100
      train: False
      transforms:
        - Resize:
            size: 32
        - ToTensor
        - Normalize:
            mean:
              - 0.4914
              - 0.4822
              - 0.4465
            std:
              - 0.2023
              - 0.1994
              - 0.2010
      target_transform: null
      download: True

hydra:
  searchpath:
    - pkg://super_gradients.recipes

```

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -294,12 +294,14 @@ class Trainer:
                 if not isinstance(test_dataloaders, Mapping):
                     raise ValueError("`test_dataloaders` should be a mapping from test_loader_name to test_loader_params.")
 
-                if test_dataloader_params.keys() != test_dataset_params.keys():
+                if test_dataloader_params is not None and test_dataloader_params.keys() != test_dataset_params.keys():
                     raise ValueError("test_dataloader_params and test_dataset_params should have the same keys.")
 
             for dataset_name, dataset_params in test_dataset_params.items():
                 loader_name = test_dataloaders[dataset_name] if test_dataloaders is not None else None
-                loader = dataloaders.get(loader_name, dataset_params=test_dataset_params[dataset_name], dataloader_params=test_dataloader_params)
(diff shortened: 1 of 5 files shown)


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -284,6 +284,26 @@ def train_from_config(cls, cfg: Union[DictConfig, dict]) -> Tuple[nn.Module, Tup
             dataloader_params=cfg.dataset_params.val_dataloader_params,
         )
 
+        test_loaders = {}
+        if "test_dataset_params" in cfg.dataset_params:
+            test_dataloaders = get_param(cfg, "test_dataloaders")
```
shaydeci, line 289: I feel this is getting a bit all over the place...
test_dataloader_parameters are optional, they must fit the keys, test_dataset_params that are mappings...

I feel the form where we pass the dataset class name through the dataloader params might simplify all of this.

I.e, like in the pose recipes:
```
train_dataloader_params:
  dataset: COCOPoseEstimationDataset
  batch_size: 8
  num_workers: 8
  drop_last: True
  pin_memory: False
  shuffle: True
  collate_fn: YoloNASPoseCollateFN
```

I tend to think in general this is a preferred way, and think we should migrate at some point.

shaydeci (changes requested):

shaydeci (approved):
LGTM

PR #1642, opened 2023-11-16
Fix Invalid syntax at convert_recipe_to_code.py

seunghalee1226:
I added **double quotes** around the name parameter in dataloaders.get() in the code below.

```
    train_dataloader = dataloaders.get(
        name="{train_dataloader}",
        dataset_params={train_dataset_params},
        dataloader_params={train_dataloader_params},
    )
    val_dataloader = dataloaders.get(
        name="{val_dataloader}",
        dataset_params={val_dataset_params},
        dataloader_params={val_dataloader_params},
    )
```

### DIFF
--- a/src/super_gradients/convert_recipe_to_code.py
+++ b/src/super_gradients/convert_recipe_to_code.py
@@ -179,13 +179,13 @@ def main():
     )
 
     train_dataloader = dataloaders.get(
-        name="{train_dataloader}",
+        name={wrap_in_quotes_if_string(train_dataloader)},
         dataset_params={train_dataset_params},
         dataloader_params={train_dataloader_params},
     )
 
     val_dataloader = dataloaders.get(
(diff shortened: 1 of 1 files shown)


BloodAxe:
Thanks for spotting this. I think adding quotes is not correct fix since `train_dataloader` can be `None` and wrapping quotes would make it `"None"` which is a string.
Instead please use `wrap_in_quotes_if_string` method: name={wrap_in_quotes_if_string(train_dataloader)}`. That should do the job.

seunghalee1226:
Oh, I just realized that the train/valid_dataloader variables could be None. As you mentioned earlier, I used the wrap_in_quotes_if_string method. Without wrap_in_quotes_if_string, the conversion to a string type doesn't happen in train/valid dataloader, and this was causing a bug.

BloodAxe (approved):
LGTM

PR #1643, opened 2023-11-16
Add automatic model transition to CUDA if FP16 export is requested

BloodAxe:


shaydeci (approved):
LGTM with one small suggestion- should we print some message to let the user know we are using CUDA device for this case?
We don't have to (I guess this would usually happen by accident), just wondering.

BloodAxe:
> LGTM with one small suggestion- should we print some message to let the user know we are using CUDA device for this case?
> 
> We don't have to (I guess this would usually happen by accident), just wondering.

I added it at first but then thought - user would see it and "so what?". This is not "action required" message. Thought for sake "it just works" there no point to go verbose. Can add if you think it would help.

shaydeci:
> > LGTM with one small suggestion- should we print some message to let the user know we are using CUDA device for this case?
> > We don't have to (I guess this would usually happen by accident), just wondering.
> 
> I added it at first but then thought - user would see it and "so what?". This is not "action required" message. Thought for sake "it just works" there no point to go verbose. Can add if you think it would help.

up to you (:

PR #1652, opened 2023-11-21
Prevent modification of the input argument (targets) in DetectionVisualization.visualize_batch

BloodAxe:
Thi PR fixes bug in ExtremeBatchDetectionVisualizationCallback (No targets are rendered)


shaydeci (approved):
LGTM

PR #1665, opened 2023-11-26
OhemLoss:  bug fix

danielafrimi:
OhemLoss forward, if all pixels are ignore labels, return empty loss tensor and pass it to logits device


BloodAxe:
@danielafrimi Thanks! But please fix the PR using [signed commits](https://github.com/Deci-AI/super-gradients/blob/master/CONTRIBUTING.md#signed-commits) otherwise we would not be able to merge your pr

PR #1669, opened 2023-11-28
 OhemLoss: bug fix

danielafrimi:
OhemLoss forward, if all pixels are ignore labels, return empty loss tensor and pass it to logits device


BloodAxe (approved):
LGTM

PR #1678, opened 2023-11-30
Segmentation sample + albumentation support for segmentation

shaydeci:
- Introduced SegmentationSample.
- Added support for segmentaiton throught the albumentations pipeline adapter. Our segmentaiton transforms work with PIL images, so theres a little back and forth of casting since I also took out the hardcoded Normalize and ToTensor which we had in our segmentaiton datasets.
- Removal of hard coded Notmalize and ToTensor from SegmentationDataset (added in all of our recipes equivalent SegToTensor and SegNormalize).
- Added integration test.


src/super_gradients/training/samples/segmentation_sample.py
```diff
@@ -0,0 +1,31 @@
...
+    :param additional_samples: (Optional) List of additional samples for the same image.
+    """
+
+    __slots__ = ["image", "mask"]
+
+    image: Union[np.ndarray, torch.Tensor]
+    mask: Union[np.ndarray, torch.Tensor]
+
+    def __init__(self, image: Union[np.ndarray, torch.Tensor, Image], mask: Union[np.ndarray, torch.Tensor, Image]):
+        self.image = image
```
BloodAxe, line 30: If we want to allow input to be PIL image, we can convert it here to np.ndarray in __init__ since both Albu works only with np.ndarray and conversion to tensor also uses numpy and not PIL image

src/super_gradients/training/samples/segmentation_sample.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@dataclasses.dataclass
+class SegmentationSample:
+    """
+    A data class describing a single object detection sample that comes from a dataset.
+    It contains both input image and target information to train an object detection model.
+
+    :param image:              Associated image with a sample. Can be in [H,W,C] or [C,H,W] format
+    :param bboxes_xywh:        Numpy array of [N,4] shape with bounding box of each instance (XYWH)
```
BloodAxe, line 18: Docs from DetectionSample

src/super_gradients/training/transforms/transforms.py
```diff
@@ -743,6 +755,33 @@ def get_equivalent_preprocessing(self):
         raise NotImplementedError("get_equivalent_preprocessing is not implemented for non-deterministic transforms.")
 
 
+@register_transform(Transforms.SegToTensor)
+class SegToTensor(AbstractSegmentationTransform, LegacySegmentationTransformMixin):
+    def apply_to_sample(self, sample: SegmentationSample) -> SegmentationSample:
+        sample.image = to_tensor(sample.image)
+        sample.mask = torch.from_numpy(np.array(sample.mask)).long()
```
BloodAxe, line 762: Let's check that this transform can be used with binary segmentation (BCE loss) and semantic segmentation (CE loss).
A few pitfalls that I'm aware of:
- For BCE the shape must match the shape of preds (BCHW) and also targets should be of float format (and not long, nn.BCEWithLogistLoss will throw an error if target is long if I'm not mistaken)
- OpenCV may silently squeeze channel dimension (HW1 -> HW) in resize/pad/etc. In Albu it is handled by `@preserve_channel_dim` decorator. We have want to have similar decorator OR have the argument in `SegToTensor(dummy_channel_dim=True)` to explicitly tell whether dummy channel dim is needed.
- Maybe same of output dtype in `SegToTensor(mask_dtype=...)`

src/super_gradients/training/samples/segmentation_sample.py
shaydeci: The reason for the support for PIL image is because this is what all of our segmentation transforms expect.
What we should aim for IMO is to support PIL images in segmentation sample and gradually reproduce our recipes with albumentations transforms, then at that point deprecate and eventually remove them.

src/super_gradients/training/transforms/pipeline_adaptors.py
```diff
@@ -62,11 +67,17 @@ def post_transforms_processing(self, sample):
...
+            sample = DetectionSample(
                 image=sample["image"],
                 bboxes_xyxy=np.array(sample["bboxes"]),
                 labels=np.array(sample["labels"]),
                 is_crowd=np.array(sample["is_crowd"]),
                 additional_samples=None,
             )
-        return sample["image"]
+        elif self.sample_type == SampleType.SEGMENTATION:
+            sample = SegmentationSample(image=Image.fromarray(sample["image"]), mask=Image.fromarray(sample["mask"]))
```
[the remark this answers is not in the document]
shaydeci: The reason for the casting to PIL image is because this is what all of our segmentation transforms expect.
So if the next transform in the pipeline is a native one we have to cast it.
See also my next comment about how I see the future support for segmentation transforms.

src/super_gradients/training/transforms/transforms.py
shaydeci: added 713401666e053c8196a8e979b6d6c485167d25d3

src/super_gradients/training/transforms/transforms.py
```diff
@@ -743,6 +756,54 @@ def get_equivalent_preprocessing(self):
...
+class SegToTensor(AbstractSegmentationTransform, LegacySegmentationTransformMixin):
+    """
+    Converts SegmentationSample images and masks to PyTorch tensors.
+
+    :param mask_output_dtype (Optional[str]): The desired output data type for the mask tensor.
+    :param add_mask_dummy_dim (bool): Whether to add a dummy channels dimension to the mask tensor.
+    """
+
+    def __init__(self, mask_output_dtype: Optional[str] = None, add_mask_dummy_dim: bool = False):
+        self.mask_output_dtype = mask_output_dtype
```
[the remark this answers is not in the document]
shaydeci: This was a mistake, I updated the type to torch.dtype 8764dcefe439a613c540da6bd1f2bda7e0a59433

BloodAxe (approved):
LGTM

PR #1686, opened 2023-12-06
fix dataset plot and improve the plotting

Louis-Dupont:
- Pascal VOC fixed (it used to use a deprecated params that actually wasnt supported anymore for plotting)
- Making `visualize_image` a public method with default params.
- Moving from "custom" plotting to using the `visualize_image`

### DIFF
--- a/src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
+++ b/src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
@@ -21,15 +21,14 @@ from super_gradients.common.factories.transforms_factory import TransformsFactor
 from super_gradients.common.object_names import Datasets, Processings
 from super_gradients.common.registry.registry import register_dataset
 from super_gradients.module_interfaces import HasPreprocessingParams
-from super_gradients.training.datasets.data_formats.default_formats import XYXY_LABEL
+from super_gradients.training.datasets.data_formats.default_formats import LABEL_XYXY
 from super_gradients.training.datasets.data_formats.formats import ConcatenatedTensorFormat, LabelTensorSliceItem
 from super_gradients.training.transforms.detection.legacy_detection_transform_mixin import LegacyDetectionTransformMixin
 from super_gradients.training.transforms.transforms import AbstractDetectionTransform, DetectionTargetsFormatTransform, DetectionTargetsFormat
 from super_gradients.training.utils.detection_utils import get_class_index_in_target
 from super_gradients.training.utils.utils import ensure_is_tuple_of_two
 from super_gradients.training.datasets.data_formats import ConcatenatedTensorFormatConverter
-from super_gradients.training.utils.visualization.utils import generate_color_mapping
-from super_gradients.training.utils.visualization.detection import draw_bbox
+from super_gradients.training.utils.detection_utils import DetectionVisualization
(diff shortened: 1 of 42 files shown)


src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -529,34 +539,62 @@ def plot(self, max_samples_per_plot: int = 16, n_plots: int = 1, plot_transforme
...
-                boxes = sample["target"][:, 0:4]
+                image = image.astype(np.uint8)
+                image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)  # Detection dataset works with BGR images, so we have to convert to RGB
+
+                # Convert to XYXY_LABEL format
+                targets_format_converter = ConcatenatedTensorFormatConverter(input_format=input_format, output_format=XYXY_LABEL, image_shape=image.shape)
+                targets_xyxy_label = targets_format_converter(targets)
+
+                # PLOT TARGETS
+                color_mapping = generate_color_mapping(len(self.classes))
```
BloodAxe, line 576: We have DetectionVisualization that contains many similar routines we use here and I wonder maybe we can concentrate all drawing logic in that class and call DetectionVisualization._visualize_image (make sense to make method public)? WDYT?

Louis-Dupont: I pushed it here [2b2dbc7](https://github.com/Deci-AI/super-gradients/pull/1686/commits/2b2dbc784016fb255e8c9bf602b8c3fe43e99318)

BloodAxe:
Overall LGTM, just ensure that you've updated public API change in notebooks (model_export does use DetectionVisualization)

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -423,16 +425,20 @@ def draw_box_title(
         :param y2: Y coordinate of the bottom right corner of the bounding box
         :param class_id: A corresponding class id
         :param pred_conf: Class confidence score (optional)
-        :param is_target: Indicate if the bounding box is a ground-truth box or not
-
+        :param bbox_prefix: Prefix to add to the title of the bounding boxes
+        :param is_target: Not Supported Anymore. Please use  `pred_conf` and `bbox_prefix` instead.
         """
+        if is_target:
```
BloodAxe, line 431: Although this method is public I think there are really low chances someone is using it aside. So it should be safe to just remove the argument completely

Louis-Dupont: Removed in [efd3579](https://github.com/Deci-AI/super-gradients/pull/1686/commits/efd357918383b7a750dc690d31cbf535d2089b2a)

BloodAxe (approved):
LGTM

ani-mal:
When will this change be released. I am running into an error because the `box[1]` value is `int64` and not `int` for `draw_box_tiles(...)`, this PR seems to address this issue:

Missing `int` casting here on version 3.5.0:
 
https://github.com/Deci-AI/super-gradients/blob/4d68589d5e846a1183a0bcbbd77deb2722fdc925/src/super_gradients/training/utils/detection_utils.py#L459-L462

PR #1687, opened 2023-12-06
Adjust bbox thickness and text to bbox size

Louis-Dupont:


src/super_gradients/training/utils/visualization/detection.py
```diff
@@ -27,16 +27,36 @@ def draw_bbox(
     :param y2:              y-coordinate of the bottom-right corner of the bounding box.
     """
 
+    bbox_width = x2 - x1
+    bbox_height = y2 - y1
+    diag_length = np.sqrt(bbox_width**2 + bbox_height**2)
+
+    if box_thickness is None:
+        # Calculate bbox thickness as a percentage of the geometric mean of bbox width and height
+        box_thickness = int(max(1, diag_length * 0.007))
```
shaydeci, line 36: Where is this magic number coming from ?
I wonder whether we should also take in consideration the resolution of the entire image. I think box thickness of 1 pixel would probably become invisble (will it actuallly ever be 1?).

shaydeci (changes requested):

src/super_gradients/training/utils/visualization/detection.py
Louis-Dupont: There is no big theory behind it, it's a quick and dirty value I just used. Here are a few examples of how big a bbox needs to be to be displayed at a different `box_thickness`: 
- `202x202` bbox -> `diag_length =285.7` px -> `box_thickness = 2`
- `303x303` bbox -> `box_thickness = 3`
- `404x404` bbox -> `box_thickness = 4`

But IMO it's hard to understand what it means with the numbers, I think it's nicer to see directly on the images. 

Concerning the value of 1, and it's not invisible; all the small objects shown in the examples are with 1 px.

shaydeci (dismissed):
LGTM on my end, but build seems to fail.

shaydeci (approved):
LGTM

PR #1694, opened 2023-12-10
Make `(QA)RepVGG` NAS-able using similar API as `ConvBnAct`

NatanBagrov:
Background: when applying neural architecture search, one might want to experiment with different blocks. To be able to easily experiment different blocks, all "searchable" blocks should contain a shared core API.

For Conv blocks, `kernel_size` and `padding` are part of the core API. This however, is not relevant for `RepVGG` blocks (and variants), because these blocks only support `3x3` and `dilation` kernel size and padding, respectively.

This PR adds `kernel_size` and `padding` to be part of the RepVGG variants' API, so these blocks become easily searchable, and on the other hand also checks that the values passed are those which are indeed supported.

### DIFF
--- a/src/super_gradients/modules/qarepvgg_block.py
+++ b/src/super_gradients/modules/qarepvgg_block.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Type, Union, Mapping, Any, Optional
 
 import torch
@@ -60,8 +61,7 @@ class QARepVGGBlock(nn.Module):
         use_alpha: bool = False,
         use_1x1_bias: bool = True,
         use_post_bn: bool = True,
-        kernel_size: int = 3,
-        padding: int = 1,
+        **kwargs,
     ):
         """
         :param in_channels: Number of input channels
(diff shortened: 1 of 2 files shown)


BloodAxe:
I understand the motivation, but the proposed change worsens a public class API.
By introducing a two parameters that cannot be changed with are adding unnecessary complexity to the codebase which is not directly needed by SG itself. So from my perspective this should be fixed on the NAS (by having a custom RepVGG class to "consume" extra argument"). 
@ofrimasad @shaydeci what do you guys think?

NatanBagrov:
Thanks for the comment. Another option is to inherit, from these classes, **in SG**, with `NAS(QA)RepVGG`. On one hand, it solves the same problem, on the other hand it introduces two classes that are used for NAS purposes only. Of course, connecting this to the broader offering (beyond SG) - it makes sense, to me.

shaydeci (commented):
I agree with @BloodAxe on this.

NatanBagrov:
Well, ok

PR #1696, opened 2023-12-11
Bugfixes in DistanceBasedDetectionMetrics and few improvements in DetectionMetrics

BloodAxe:
Fixes a few bugs in DistanceBasedDetectionMetric:
* A crash when crowd targets were present (Wrong shapes of tensors)
* Not respecting a `device`

A few improvements to DetectionMetrics:
* If `class_names` are present ensures they are converted to list and not retained as ListConfig
* A state dict that should be synced in DDPP and output metrics now has explicit key variables to reference. No more `f"matching_info{self._get_range_str()}"` in many places.


BloodAxe:
@DimaBir pardon for poking you, but it seems there is a test `test_random_case_generation_and_verification` is failing now and I don't quite get why. 
I did some fixes to your metric but nothing too big to affect that. Would appreciate if you can take a look!

tests/unit_tests/detection_metrics_distance_based_test.py
```diff
@@ -284,7 +271,7 @@ def test_random_case_generation_and_verification(self):
         expected_mAP, expected_f1 = self.calculate_expected_metrics(self, target_precision, target_recall)
 
         # Validate the results
-        self.validate_results(results, precision=target_precision, recall=target_recall, mAP=expected_mAP, F1=expected_f1)
+        self.validate_results(results, precision=target_precision, recall=target_recall, mAP=None, F1=expected_f1)
```
Louis-Dupont, line 274: Why set `mAP=None` ?

Louis-Dupont (approved):
LGTM, just minor question

tests/unit_tests/detection_metrics_distance_based_test.py
BloodAxe: In the previous version the mAP was not checked at all for all cases: https://github.com/Deci-AI/super-gradients/pull/1696/files#diff-9739186c39900e4bbc707a4059031225ac260667e5814fc2fdfb5a896e3a7502L90

In the remaining place the random generation of targets is somewhat non-deterministic, and mAP and since the confidence of predicted boxes is always 1 then their order actually impacts the mAP score. So for this specific case mAP is not checked and in all other cases where we have deterministic boxes it is checked

PR #1697, opened 2023-12-11
Fix ExtremeBatchDetectionVisualizationCallback when multiscale collate fn in use

BloodAxe:
Fixes a bug that occurs when using ExtremeBatchDetectionVisualizationCallback and using DDP and multiscale collate fn.
Also fixes an issue of wrong epoch being logged which prevented images to being displayed in WandB


Louis-Dupont (dismissed):
LGTM

shaydeci (changes requested):

shaydeci (dismissed):
LGTM

shaydeci (approved):
LGTM

PR #1702, opened 2023-12-12
Ensure that ImagePermute is present in processings

BloodAxe:
One can train a model with a following set of transforms:
```
  transforms:
    - DetectionStandardize:
        max_value: 255
    - DetectionTargetsFormatTransform:
        output_format: LABEL_CXCYWH
```

Note there is no ImagePermute transform and CrowdDetectionPPYoloECollateFN/PPYoloECollateFN will do HWC->CHW conversion automatically. 
However, in this case trained model cannot be used with `model.predict()` because we rely on presence of this transform in processings.

This PR ensure that ImagePermute is added if it's missing. 
Maybe we should not have ImagePermute at all in the first place and to HWC->CHW always, but removing this step would be a breaking change.


src/super_gradients/training/pipelines/pipelines.py
```diff
@@ -285,6 +285,16 @@ def __init__(
         if isinstance(image_processor, list):
             image_processor = ComposeProcessing(image_processor)
 
+        has_image_permute = False
+
+        for image_processing in image_processor.processings:
+            if isinstance(image_processing, ImagePermute):
+                has_image_permute = True
+                break
```
Louis-Dupont, line 293: What do you think about this ?
```python
has_image_permute  = any(isinstance(image_processing, ImagePermute) for image_processing in image_processor.processings)
```

Louis-Dupont (approved):
LGTM, just small note

src/super_gradients/training/pipelines/pipelines.py
BloodAxe: Agree, looks way better. Fixed in a4632b8c73d28b37f1fd52ca3fa321370d83b912

Louis-Dupont (approved):
LGTM

PR #1705, opened 2023-12-14
Feature/sg 1277 depth estimation dataset support

shaydeci:
- Base class for depth estimation dataset.
- NYUv2 Depth Estimaiton dataset class + subset of examples for testing.
- DepthEstimationSample + PipelineAdapteor support for it.
- Unit tests and Albumentations integration test.


src/super_gradients/training/samples/depth_estimation_sample.py
```diff
@@ -0,0 +1,35 @@
...
+    :param image:              np.ndarray, Image of [H, W, (C if colorful)] shape.
+    :param depth_map:          Depth map of [H, W] shape.
+    """
+
+    __slots__ = ["image", "depth_map"]
+
+    image: np.ndarray
+    depth_map: np.ndarray
+
+    def __init__(self, image: np.ndarray, depth_map: np.ndarray):
```
BloodAxe, line 24: With depth estimation there is one important nuance that should be explicitly taken care of which is invalid areas in the target depth map.
Historically, a monocular depth estimation datasets were obtained from the stereo setup. By computing disparity map between two cameras authors came up with a depth map. However stereo matching inherently is noisy process and due to the occlusion there may be some areas that cannot be estimated. 

And to account for that often these "invalid" regions has some magic value in those pixel values. (This is dataset specific and could be anything - -1, nan, inf, 0 - whatever basically).

So we should respect this value. It is important to keep track of this value as it is also should be considered in loss and metric. So I suggested use a "nan" value and document this in both dataset and depth sample that non-finite values in depth map correspond to ignored values.

Similary a specific dataset class (NYU or any other) should take care of converting their internal depth representation format to our and mask those invalid regions as nans.

BloodAxe: The motivation of using "nan" is simple - any math operation `x op nan` would produce `nan` so we don't have to worry about spatial operations like resize (I'm theoretizing here actually, we need to check)

ofrimasad: Deliberately introducing NaNs into the code is not a good practice. 
They are sneaky and have a tendency to slowly spread. 
You multiply by NaN, and you get a NaN. You then sum and get NaN, then your gradients have NaN, and only after ten epochs everything collapse.

src/super_gradients/common/object_names.py
```diff
@@ -424,6 +424,7 @@ class Datasets:
     PASCAL_VOC_AND_AUG_UNIFIED_DATASET = "PascalVOCAndAUGUnifiedDataset"
     COCO_KEY_POINTS_DATASET = "COCOKeypointsDataset"
     COCO_POSE_ESTIMATION_DATASET = "COCOPoseEstimationDataset"
+    NYUV2DATASET = "NYUv2Dataset"
```
ofrimasad, line 427: please add the word "depth" to the name

src/super_gradients/training/samples/depth_estimation_sample.py
BloodAxe: I we document properly that nan correspond to invalid region and respect this value in loss and metric we should be good I think 
The alternative is to either have a validity_mask or explicit invalid_value in the sample object. Which frankly would make things more complicated as now one have to pass two objects to loss and metric classes (depth targets and valid_mask/invalid_value)

src/super_gradients/training/datasets/depth_estimation_datasets/abstract_depth_estimation_dataset.py
```diff
@@ -0,0 +1,86 @@
...
+        random_index = random.randrange(0, num_samples)
+        return self.load_sample(random_index)
+
+    def __getitem__(self, index: int) -> DepthEstimationSample:
+        sample = self.load_sample(index)
+        for transform in self.transforms:
+            sample = transform.apply_to_sample(sample)
+        return sample
+
+    def plot(self, max_samples_per_plot: int = 8, n_plots: int = 1, plot_transformed_data: bool = True):
```
ofrimasad, line 48: There are a few coloring schemes for depth estimation (for some reason).
Lets allow the user to pass a coloring scheme here (you can implement only one for now):
- BLUE_TO_YALLOW (viridis)
- ORANGE_TO_PURPLE (plasma inverted)
- RAINBOW (rainbow)

src/super_gradients/training/transforms/depth_estimation/abstract_depth_estimation_transform.py
```diff
@@ -0,0 +1,22 @@
...
+        """
+        Apply transformation to given depth estimation sample.
+        Important note - function call may return new object, may modify it in-place.
+        This is implementation dependent and if you need to keep original sample intact it
+        is recommended to make a copy of it BEFORE passing it to transform.
+
+        :param sample: Input sample to transform.
+        :return:       Modified sample (It can be the same instance as input or a new object).
+        """
+        raise NotImplementedError()
```
ofrimasad, line 22: don't we also want to have `__call__(self, sample: DepthEstimationSample) -> DepthEstimationSample:` ?

ofrimasad (commented):
very nice. 
a few minor comments

src/super_gradients/training/transforms/depth_estimation/abstract_depth_estimation_transform.py
shaydeci: Thre's actually no reason to keep apply_Sample for depth estimation (this is used in our other legact transforms for other tasks since we introduced their respective mixin, in order to handle the newly introduces sample classes).
Moved to __ call __ in dddf9d35e5954a6aa379d3ec3e3adc8fb7cf6f65

src/super_gradients/training/samples/depth_estimation_sample.py
shaydeci: @BloodAxe and I agreed this would be taken care of in the inheriting classes load_sample methods.
I am keepingthe code as is for now as this preprocessed version of NYUv2 already excludes nans, infs and negative values.

src/super_gradients/training/datasets/depth_estimation_datasets/abstract_depth_estimation_dataset.py
shaydeci: sure. added in 2541c95c4caa5e132da97133c06feeb1632bb7b0

src/super_gradients/training/datasets/depth_estimation_datasets/abstract_depth_estimation_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+
+                image, depth_map = sample.image, sample.depth_map
+
+                # Plot the image
+                axes[0, img_i].imshow(image)
+                axes[0, img_i].axis("off")
+                axes[0, img_i].set_title(f"Sample {index}")
+
+                # Plot the depth map side by side with the selected color scheme
+                axes[1, img_i].imshow(depth_map, cmap=color_scheme)
```
BloodAxe, line 79: What do you think about having a DepthVisualization class or something to have a method to generate a depth map visualization using cv2.colorMap? 
Why a method and not ax.imshow:
- We can have a single method to visualize depthmap that we can use in extreme batch callback, model.predict().show() and here in dataset.plot
- It would also mask invalid regions of the depth map (if it's a target) and render them with gray color
- For predictions/targets what I found very helpful is to drop 1-2% of the most extreme values on both ends of the spectrum of depth targets before normalizing depth to 0..255 range and then apply colormap. This way data outliers has less effect on the overall visualization

shaydeci: Good idea, added.

image

shaydeci: inverse + drop_extreme_percentage=5
image

src/super_gradients/training/utils/visualization/depth_estimation.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+
+        # Drop extreme values
+        min_val = np.percentile(depth_map, drop_extreme_percentage)
+        max_val = np.percentile(depth_map, 100 - drop_extreme_percentage)
+        depth_map = np.clip(depth_map, min_val, max_val)
+
+        # Normalize to 0-255
+        depth_map_normalized = ((depth_map - min_val) / (max_val - min_val) * 255).astype(np.uint8)
+        if inverse:
```
BloodAxe, line 34: A few important aspects:
1) I beleive inversion should be done as 1 /  depth, this is how disparity map is defined
2) Usually when using direct depth for visualization a COLORMAP_VIRIDIS is fine choice
3) For disparity (inverse=True) a COLORMAP_MAGMA is a default choice (IDK why)
4) OpenCV uses BGR notation, so I think we should to BGR->RGB conversion at the very end

src/super_gradients/training/utils/visualization/depth_estimation.py
```diff
@@ -0,0 +1,39 @@
...
+        :param depth_map:               Input depth map as a NumPy array.
+        :param color_scheme:            OpenCV color scheme for the depth map visualization.
+        :param drop_extreme_percentage: Percentage of extreme values to drop.
+        :param inverse:                 Apply inversion (255 - depth_map) if True.
+
+        :return:                        Processed colormap of the depth map for visualization.
+        """
+
+        # Drop extreme values
+        min_val = np.percentile(depth_map, drop_extreme_percentage)
```
BloodAxe, line 28:
```
if drop_extreme_percentage > 0:
 min_val = np.percentile(depth_map, drop_extreme_percentage)
 max_val = np.percentile(depth_map, 100 - drop_extreme_percentage)
 depth_map = np.clip(depth_map, min_val, max_val)
```

src/super_gradients/training/utils/visualization/depth_estimation.py
```diff
@@ -0,0 +1,39 @@
...
+        max_val = np.percentile(depth_map, 100 - drop_extreme_percentage)
+        depth_map = np.clip(depth_map, min_val, max_val)
+
+        # Normalize to 0-255
+        depth_map_normalized = ((depth_map - min_val) / (max_val - min_val) * 255).astype(np.uint8)
+        if inverse:
+            depth_map_normalized = 255 - depth_map_normalized
+
+        # Apply colormap
+        colormap = cv2.applyColorMap(depth_map_normalized, color_scheme)
```
BloodAxe, line 38: At this point we should also fill invalid regions:
`colormap[ignored_areas] = (127, 127, 127)`

src/super_gradients/training/datasets/depth_estimation_datasets/abstract_depth_estimation_dataset.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        num_samples = len(self)
+        random_index = random.randrange(0, num_samples)
+        return self.load_sample(random_index)
+
+    def __getitem__(self, index: int) -> DepthEstimationSample:
+        sample = self.load_sample(index)
+        for transform in self.transforms:
+            sample = transform(sample)
+        return sample
```
BloodAxe, line 48: So we will be adding a collate function to go from `List[DepthEstimationSample]` to `images` and `targets`?

src/super_gradients/training/utils/visualization/depth_estimation.py
shaydeci: 3d38e6a39442e075789f4b1fdb95fdffe2f135d7 see updated

src/super_gradients/training/utils/visualization/depth_estimation.py
shaydeci: 3d38e6a39442e075789f4b1fdb95fdffe2f135d7

src/super_gradients/training/utils/visualization/depth_estimation.py
shaydeci: 3d38e6a39442e075789f4b1fdb95fdffe2f135d7

src/super_gradients/training/utils/visualization/depth_estimation.py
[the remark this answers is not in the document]
shaydeci: 3d38e6a39442e075789f4b1fdb95fdffe2f135d7

src/super_gradients/training/datasets/depth_estimation_datasets/abstract_depth_estimation_dataset.py
shaydeci: a4bcf66f64ff61c9968bee71706b12de2ac07033 accident, fixed

BloodAxe (approved):
LGTM

PR #1709, opened 2023-12-17
Feature/sg 1277 Depth Estimation Metrics Support

shaydeci:


src/super_gradients/training/metrics/depth_estimation_metrics.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if preds.ndim == 4:
+            preds = preds.squeeze(1)
+
+        if self.apply_sigmoid:
+            preds = torch.sigmoid(preds)
+
+        if self.ignore_val is not None:
+            non_ignored = preds != self.ignore_val
+            preds = preds[non_ignored]
```
BloodAxe, line 46: I think it should be computed in a slightly different fashion.
The `preds = preds[non_ignored]` line would translate `[B,1,H,W]` tensor into `[N]` single-dimensional array which will be passed to metric and treated as a single element.

However the depth metrics usually compute the statistics per image and then average.
So what we should be doing instead is:
```
for pred_i, tgt_i in zip(preds, targets):
  if self.ignore_val is not None:
      non_ignored = tgt_i != self.ignore_val
      pred_i = pred_i[non_ignored]
      tgt_i = tgt_i[non_ignored]
  self.update(pred_i, tgt_i)
```

src/super_gradients/common/object_names.py
```diff
@@ -35,6 +35,16 @@ class Metrics:
     PIXEL_ACCURACY = "PixelAccuracy"
     POSE_ESTIMATION_METRICS = "PoseEstimationMetrics"
     DETECTION_METRICS_DISTANCE_BASED = "DetectionMetricsDistanceBased"
+    DELTAMETRIC = "DeltaMetric"
```
BloodAxe, line 38: Should we add `Depth` prefix to these metrics? MSE is quite common name and maybe it may cause a confusion if some day we decide to use MSE for image reconstruction

src/super_gradients/training/metrics/depth_estimation_metrics.py
```diff
@@ -0,0 +1,250 @@
...
+    :param ignore_val: Value to be ignored when computing the metric.
+    :param apply_sigmoid: Whether to apply the sigmoid function to predictions before updating the metric.
+    """
+
+    def __init__(self, ignore_val: Optional[float] = None, apply_sigmoid: bool = False):
+        super().__init__(squared=True, ignore_val=ignore_val, apply_sigmoid=apply_sigmoid)
+
+
+@register_metric(Metrics.RMSE)
+class RMSE(MeanSquaredErrorWithIgnored):
```
BloodAxe, line 91: This class is the same as MSE. Base class inheritance is wrong?

src/super_gradients/training/metrics/depth_estimation_metrics.py
shaydeci: Not true, It is mathemtically equivalent to take the mean over the pixels of each image (see formulas and go up the parents class you would see that there are two states being updated - total pixels, which is computed after flattening and extracting numeral() or shape[0] and the sum of squared errors).
So this is unnecessary.

src/super_gradients/training/metrics/depth_estimation_metrics.py
shaydeci: Accidentally left default value for squared argument which is the difference. Fixed.

src/super_gradients/training/metrics/depth_estimation_metrics.py
```diff
@@ -0,0 +1,250 @@
...
+    """MeanAbsoluteError, MeanAbsolutePercentageError
+    Mean Squared Error metric for depth estimation with support for ignored values.
+
+    :param squared: If True returns MSE value, if False returns RMSE value.
+    :param ignore_val: Value to be ignored when computing the metric.
+    :param apply_sigmoid: Whether to apply the sigmoid function to predictions before updating the metric.
+    """
+
+    def __init__(self, squared: bool = True, ignore_val: Optional[float] = None, apply_sigmoid: bool = False):
+        super().__init__(squared=squared)
```
Louis-Dupont, line 63: The first `super().__init__` refers to the first class (`MeanSquaredError` here) right ?
I think in case of double inheritence it would be better to be more explicit about every init: `MeanSquaredError.__init__`

src/super_gradients/training/metrics/depth_estimation_metrics.py
```diff
@@ -0,0 +1,250 @@
...
+
+    :param ignore_val: Value to be ignored when computing metrics.
+    :param apply_sigmoid: Whether to apply the sigmoid function to predictions before updating the metric.
+    """
+
+    def __init__(self, ignore_val: Optional[float] = None, apply_sigmoid: bool = False):
+        self.ignore_val = ignore_val
+        self.apply_sigmoid = apply_sigmoid
+
+    def process_preds_and_target(self, preds: Union[Tensor, Sequence[Tensor]], target: Tensor) -> Tuple[Tensor, Tensor]:
```
Louis-Dupont, line 22: Do we need this `DepthEstimationMetricMixin ` class only as a "utils toolkit" ?

If so, maybe we could simplify the code/inheritance by replacing this mixin by a function:

```python
def format_depth_preds_and_targets(preds: Union[Tensor, Sequence[Tensor]], target: Tensor, ignore_val: Optional[float] = None, apply_sigmoid: bool = False)
```

And then use it like this


```python
class MeanSquaredErrorWithIgnored(MeanSquaredError):
    def __init__(self, squared: bool = True, ignore_val: Optional[float] = None, apply_sigmoid: bool = False):
        super().__init__(squared=squared)
        self.ignore_val = ignore_val
        self.apply_sigmoid = apply_sigmoid

    def update(self, preds: Tensor, target: Tensor) -> None:
        preds, target = format_depth_preds_and_targets(preds, target, ignore_val=self.ignore_val, apply_sigmoid=self.apply_sigmoid)
        super().update(preds, target)
```


Alternatively, we can always have keep the class, but use it with composition instead of inheritence

```python
class MeanSquaredErrorWithIgnored(MeanSquaredError):
    def __init__(self, squared: bool = True, ignore_val: Optional[float] = None, apply_sigmoid: bool = False):
        super().__init__(squared=squared)
        # Note: We should rename `DepthEstimationMetricMixin` in this case
        self.formater = DepthEstimationMetricMixin(ignore_val=ignore_val, apply_sigmoid=apply_sigmoid)

    def update(self, preds: Tensor, target: Tensor) -> None:
        preds, target =self.formater.process_preds_and_target(preds, target)
        super().update(preds, target)
```

IMO it makes the inheritance logic simpler and less confusing. It also emphasizes on the facts these are just internal utils/formating steps, and not really a "contract" that other classes need to know about.
What do you think?

Louis-Dupont (commented):
I left some notes about `DepthEstimationMetricMixin` let me know your thoughts

src/super_gradients/training/metrics/depth_estimation_metrics.py
Louis-Dupont: Note: this may not be relevant if we go for something proposed in the next comment (maybe check it first)

src/super_gradients/training/metrics/depth_estimation_metrics.py
shaydeci: For me it is more clear this way, but I might be biased. I also think the code is cleaner and more consistent when inheriting from the mixin class.
@BloodAxe Whats your opinion ?

src/super_gradients/training/metrics/depth_estimation_metrics.py
shaydeci: sure, updated in b86a122ac667ca07f5fe149d0fe795b4ca4e727d

src/super_gradients/training/metrics/depth_estimation_metrics.py
BloodAxe: 1) A "canonical" mixin should not have a constructor. While the `DepthEstimationMetricMixin` has. So we are dealing with multiple inheritance class here. Not necessary a bad thing, but maybe the name of the base class should be changed and not use `mixin` word to avoid confusion.

2) Another thing that worries me is the naming of the depth metrics. I think they should have a common name prefix like "Depth"-bla-bla-bla. This is to discriminate depth-related metrics from other tasks.

3) A dedicated class just to support ignore value looks like a great overkill from my perspective `MeanSquaredErrorWithIgnored`. We should have one class (MeanSquaredError) that would allow user to set the `ignore_value` attribute. I just don't see any benefit in having second class here. I cann't see a use case where you want to compute both  `MeanSquaredError` and `MeanSquaredErrorWithIgnored` metrics simultaneously (Which can justify presence of two classes - to get metrics a unique names).

shaydeci: Ok went with your suggestion which we discussed (similar idea to what you suggested @Louis-Dupont , but the other way around - the base class receives the metric instead).
Added depth prefixes too.

BloodAxe (approved):
LGTM

BloodAxe (approved):
LGTM

PR #1710, opened 2023-12-18
Feature/sg 1271 add keypoint aug

Louis-Dupont:
Adding support for Albumentation augmentation for keypoint.

Notes:
- Using `HorizontalFlip` raises an exception, we ask the user to use `KeypointsRandomHorizontalFlip` instead.
- `area` and `additional_samples` are dropped when using albumentation (this is what was done with other tasks)

### Example
It's not easy to compare since the transforms are non-deterministic, so here I just show 2 examples with old transforms vs with new transforms.

#### SG transforms
image
image

#### Albumentation transforms
image
image


src/super_gradients/common/factories/transforms_factory.py
```diff
@@ -46,6 +47,7 @@ def get(self, conf: Union[str, dict]):
                 if "bbox_params" in conf[_type].keys():
                     bbox_params = conf[_type]["bbox_params"]
                     self._check_bbox_params(bbox_params)
+                    conf = deepcopy(conf)  # Avoid changing the original config.
                     conf[_type]["bbox_params"].update(self.FIXED_BBOX_PARAMS)
```
Louis-Dupont, line 51: There is a bug due to the fact we modify inplace `conf`: calling `AlbumentationsTransformsFactory` twice on the same config will raise an error inside of `_check_bbox_params`.
Creating a copy prevents this from happening (we don't modify original).

src/super_gradients/training/transforms/pipeline_adaptors.py
```diff
@@ -39,6 +42,18 @@ def __call__(self, sample, *args, **kwargs):
...
+        elif isinstance(sample, PoseEstimationSample):
+            self.sample_type = SampleType.POSE_ESTIMATION
+
+            from albumentations.augmentations import HorizontalFlip
+
+            if any(isinstance(t, HorizontalFlip) for t in self.composed_transforms):
+                raise TypeError(
+                    "`HorizontalFlip` from Albumentation is not supported. "
+                    "Please use the `KeypointsRandomHorizontalFlip` from SuperGradients instead.\n"
+                    "Note: You should set it like other SuperGradients transforms, and not like other Albumentations."
```
BloodAxe, line 54: Can we maybe print an example how it is supposed to look? I think users would be having a hard time figuring it out by themselves. Just a small code or YAML snippet to show how it meant to look would be super helpful.

src/super_gradients/training/transforms/pipeline_adaptors.py
```diff
@@ -55,6 +70,18 @@ def prep_for_transforms(self, sample):
...
+
+            bboxes_xyxy = xywh_to_xyxy(bboxes=np.array(sample.bboxes_xywh), image_shape=sample.image.shape)
+
+            sample = {
+                "image": sample.image,
+                "bboxes": bboxes_xyxy,
+                "labels": np.zeros(sample.bboxes_xywh.shape[0]),  # Dummy value, this is required for Albumentation. Here, all classes are the same.
+                "mask": np.array(sample.mask),
+                "is_crowd": sample.is_crowd,
+                "keypoints": sample.joints,
```
BloodAxe, line 83: I think we should also pass keypoint_params to A.Compose to indicate format of keypoints (XY).
Also, we have to ensure a few important aspects:
- That transforms can handle (x,y,visibility) format (visibility should not be changed by Albumentations)
- We should update visibility status of each keypoint if it goes outside the image bounds. Ideally we should check that after each transform.
- That we don't remove keypoints (When doing a crop for instance).
- That order of keypoints does not change during transforms. (The index of keypoint mark it's joint so order of keypoint should be fixes)

src/super_gradients/training/transforms/pipeline_adaptors.py
Louis-Dupont: Like this ?

```
TypeError: `HorizontalFlip` from Albumentation is not supported. Please use the `KeypointsRandomHorizontalFlip` from SuperGradients instead.
Note: You should set it like other SuperGradients transforms, and not like other Albumentations.

Example:

FROM 
   > transforms:
   >    - Albumentations:
   >        Compose:
   >            transforms:
   >                - HorizontalFlip:
   >                    p: 1

TO 
   > transforms:
   >    - KeypointsRandomHorizontalFlip:
   >        prob: 1
   >        flip_index: [ 0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
   >        # Note these indexes are COCO-specific. If you're using a different dataset, you will need to change these accordingly.
   >        # The `flip_index` array defines pairs of keypoints to exchange during a horizontal flip. This ensures accurate mapping of corresponding keypoints on mirrored body parts after flipping.

```

BloodAxe: Yea, that would be great exception tip!

src/super_gradients/training/transforms/pipeline_adaptors.py
Louis-Dupont: - When a joint is out of image, we set to (0, 0, 0)
- When a skeleton is full of (0,0,0) do we remove the full skeleton?
- Or, do we drop it only when the bbox is removed ? Or both ?

Louis-Dupont: This commit covers it all, but I'm not sure about the skeleton filtering. 
[2c3d0b7](https://github.com/Deci-AI/super-gradients/pull/1710/commits/2c3d0b79b875b3a102c3df18e43c4891c5c5f35e)

tests/integration_tests/albumentations_test.py
```diff
@@ -161,6 +166,202 @@ def test_coco_segmentation_albumentations_intergration(self):
...
+            box_thickness=None,
+            keypoint_radius=None,
+        )
+
+        visualize_image(image=image_with_keypoints)
+
+        # Make sure we raise an error when using unsupported transforms
+        with self.assertRaises(TypeError):
+            transforms = [
+                KeypointsRescale(height=320, width=640),
```
shaydeci, line 294: Additional test case: It would be great if we could add a native transform after the albumentations part.

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
```diff
@@ -91,7 +91,7 @@ def _apply_transforms(cls, sample: PoseEstimationSample, transforms: List[Abstra
         """
         applied_transforms_so_far = []
         for t in transforms:
-            if t.additional_samples_count == 0:
+            if not hasattr(t, "additional_samples_count") or t.additional_samples_count == 0:
```
shaydeci, line 94: I don't think its a good idea just to go on in case we did not load additional samples.
Suppose some user inherits from pose estimation dataset and does not load additional samples, they would have a hard time figuring out why Mosaic is not working.

shaydeci (changes requested):

tests/integration_tests/albumentations_test.py
```diff
@@ -161,6 +166,202 @@ def test_coco_segmentation_albumentations_intergration(self):
...
+                include_empty_samples=True,
+                edge_links=edge_links,
+                edge_colors=edge_colors,
+                keypoint_colors=keypoint_colors,
+                transforms=transforms,
+            )
+
+            _ = next(iter(unsupported_ds))
+
+        # transforms = [
```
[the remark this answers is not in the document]
Louis-Dupont: Leftover from testing, I removed it https://github.com/Deci-AI/super-gradients/pull/1710/commits/663cf8ed7c6b110fb76fafbb112be52f3d948f89

src/super_gradients/training/transforms/keypoints/keypoints_compose.py
Louis-Dupont: `t` is the transform, not the sample. So here I just check if the transform has the attribute `additional_samples_count`. If not, I assume its not a SG transform and therefore doesnt support the logic around `additional_samples_count`.

Also, here I am not changing the logic, it's just that albumentation transforms don't have attribute `additional_samples_count`, so this piece of code crashes. 

Does that make sens to you?

tests/integration_tests/albumentations_test.py
Louis-Dupont: Like this c9a0b749275e4bcf0f491697a1075f8cd82b6cb8 ?

src/super_gradients/training/transforms/pipeline_adaptors.py
Louis-Dupont: I added `sample_sanitize` here 65345a579331e52c8fff35873bfe5fb3fd126e74

shaydeci (approved):
LGTM from my side

src/super_gradients/training/transforms/pipeline_adaptors.py
```diff
@@ -83,6 +135,34 @@ def post_transforms_processing(self, sample):
...
+
+            if len(sample["bboxes"]) == 0:
+                sample["bboxes"] = np.zeros((0, 4))
+            if len(sample["is_crowd"]) == 0:
+                sample["is_crowd"] = np.zeros((0))
+
+            bboxes_xywh = xyxy_to_xywh(bboxes=np.array(sample["bboxes"]), image_shape=sample["image"].shape)
+
+            # Update value of keypoints that are not inside the image anymore.
+            h, w = sample["image"].shape[1], sample["image"].shape[0]
```
BloodAxe, line 148: I think the h and w here may be put in a wrong order. If we assume image shape is HWC, then shape[1] is width and shape[0] is height.

src/super_gradients/training/transforms/pipeline_adaptors.py
```diff
@@ -83,6 +135,34 @@ def post_transforms_processing(self, sample):
...
+            keypoints = np.array(
+                [keypoint if (0, 0) <= (keypoint[0], keypoint[1]) < (w, h) else (keypoint[0], keypoint[1], 0) for keypoint in sample["keypoints"]]
+            )
+            keypoints = keypoints.reshape(-1, sample["n_joints"], 3)
+
+            keypoints = keypoints[sample["labels"]]  # remvoe the ones associated with a bbox that was removed.
```
BloodAxe, line 154: I think we also should filter `is_crowd[sample["labels"]]` if it's present.

src/super_gradients/training/transforms/pipeline_adaptors.py
Louis-Dupont: Oh yeah, I don't know why it's reversed I didnt see, thanks!

src/super_gradients/training/transforms/pipeline_adaptors.py
Louis-Dupont: Actually no because `is_crowd` is `label` of `bbox_params` so it's filtered along with `labels`. I double checked and moving to `is_crowd[sample["labels"]]` would even make the code crash (filtering twice so there is index error)

src/super_gradients/common/factories/transforms_factory.py
```diff
@@ -74,6 +75,8 @@ def _check_bbox_params(self, bbox_params):
             raise ValueError(f"Unexpected fixed bbox_params keys: {fixed_keys}. Fixed bbox_params {self.FIXED_BBOX_PARAMS} cannot be overriden.")
 
     def _check_keypoint_params(self, keypoint_params):
+        if len(keypoint_params.keys()):
+            raise ValueError("`keypoint_params` should be left empty. Please leave it as `{keypoint_params: None}` for .json or `keypoint_params: ` for .yaml")
```
Louis-Dupont, line 79: If the user sets any `keypoint_params` the code could crash with weird errors. There is no relevant params for users to set.
At the same time, we still need to know that the user is using keypoints, so the trick is to ask to the user to have `keypoint_params` but leave it empty. Not great, but couldnt find a better way after playing around for quiet some time

src/super_gradients/training/transforms/pipeline_adaptors.py
Louis-Dupont: @BloodAxe About your idea of using `sample.filter_by_mask(labels)`.
It works if we won't have the `is_crowd` param. This is `is_crowd` is already filtered by albumentation, so it fails if we use `filter_by_mask` since it will try to filter it once more and fail (same error as in my previous message).
Basically, `bboxes` and `is_crowd` are mapped as `bbox`, so both are filtered together when a bbox is invalid. `keypoints` is the only remaining param that we need to filter.

shaydeci (approved):

PR #1718, opened 2023-12-25
removed entries from averaged state dict

shaydeci:


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -721,8 +721,11 @@ def _save_checkpoint(
 
         if self.training_params.average_best_models:
             net_for_averaging = unwrap_model(self.ema_model.ema if self.ema else self.net)
-
             state["net"] = self.model_weight_averaging.get_average_model(net_for_averaging, validation_results_dict=validation_results_dict)
+
+            # REMOVE UNNECESSARY ITEMS FROM AVERAGED STATE DICT
+            for key_to_remove in ["optimizer_state_dict", "scaler_state_dict", "ema_net"]:
```
BloodAxe, line 727: An `ema_net` is not necessary present, so pop on missing key will raise an exception.

Maybe let's use this:?
```
excluded_items = ["optimizer_state_dict", "scaler_state_dict", "ema_net"]
state = dict((key, value) for (key, value) in state if key not in excluded_items)
```

shaydeci: no it won't since there is a default value (that None argument).

BloodAxe (approved):
LGTM

BloodAxe (approved):
LGTM

PR #1719, opened 2023-12-25
Fixed bug in DetectionMixup implementation that prevents YoloXTrainingStageSwitchCallback to work as intended

BloodAxe:
When YoloXTrainingStageSwitchCallback triggers a disable of DetectionMixup there is a bug that prevents No-Op flow of this transform


shaydeci (approved):
LGTM

PR #1720, opened 2023-12-25
Feature/sg 1196 fix breaking change test

Louis-Dupont:
Instead of always checking the breaking change relatively to `master`, this PR first fetches the name of the branch that the current PR is being merging into (this could be something like `master/2.0.0` for instance).
In order to test this, I created a dummy PR (https://github.com/Deci-AI/super-gradients/pull/1731) that introduces a breaking change and is being merged to this branch. You can see in the logs that the "source" branch is not master, but is `feature/SG-1196-fix_breaking_change_test` as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -246,7 +246,8 @@ jobs:
             python3 -m pip install -r requirements.dev.txt
             python3 -m pip install gitpython==3.1.0
 
-            # GET THE BRANCH THAT THE CURRENT PR IS BEING MERGED INTO
+            # GET THE BRANCH THAT THE CURRENT PR IS BEING MERGED INTO.
+            # THIS IS ONLY RELEVANT FOR CI RUNS, WHICH IS WHY IT'S NOT PART OF THE PYTHON CODE.
             if [ -n "${CIRCLE_PULL_REQUEST}" ]; then
               PR_NUMBER=$(basename ${CIRCLE_PULL_REQUEST})
               echo "Pull Request Number: ${PR_NUMBER}"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -245,8 +245,28 @@ jobs:
...
+              # This is required since it's possible this branch would not be available locally
+              git fetch origin ${BRANCH_MERGED_INTO}:${BRANCH_MERGED_INTO}
+            else
+              echo "This job does not appear to be running as part of a pull request."
+            fi
+
+            # TESTS
             coverage run --source=super_gradients -m unittest tests/breaking_change_tests/unit_test.py
-            coverage run --source=super_gradients -m unittest tests/breaking_change_tests/test_detect_breaking_change.py
+            coverage run --source=super_gradients -m unittest tests/breaking_change_tests/test_detect_breaking_change.py --branch_merged_into=${BRANCH_MERGED_INTO}
```
Louis-Dupont, line 269: I did this as part of the CI because it's ONLY relevant in the CI.

BloodAxe: Should we set BRANCH_MERGED_INTO to default `master` if the CIRCLE_PULL_REQUEST is not set?
The call `            coverage run --source=super_gradients -m unittest tests/breaking_change_tests/test_detect_breaking_change.py --branch_merged_into=${BRANCH_MERGED_INTO}` seems to be unconditionally attempt to read from this variable so I wonder what would happen if it's not available.

Louis-Dupont: If it's not available, the default is `master` in the python side.
But it's true that it would better be explicit. I changed it here [e93be17](https://github.com/Deci-AI/super-gradients/pull/1720/commits/e93be17641223a74bd3a73cf5a3f6c1910d6c026)

shaydeci (approved):
LGTM from my side.

PR #1727, opened 2023-12-26
PPYoloELoss not does not require reg_max parameter

BloodAxe:
This PR drops the necessity to specify `reg_max` argument in the Loss which should match with the reg_max in the model.
We have all the infromation to infer it's value solely from model's outputs.

### DIFF
--- a/src/super_gradients/training/losses/ppyolo_loss.py
+++ b/src/super_gradients/training/losses/ppyolo_loss.py
@@ -657,7 +657,7 @@ class PPYoloELoss(nn.Module):
         :param classification_loss_weight: Classification loss weight
         :param iou_loss_weight:            IoU loss weight
         :param dfl_loss_weight:            DFL loss weight
-        :param reg_max:                    Number of regression bins (Must match the number of bins in the PPYoloE head)
+        :param reg_max:                    (Deprecated) Number of regression bins. Default is None (will be inferred from model's outputs)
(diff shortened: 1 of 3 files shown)


src/super_gradients/training/losses/ppyolo_loss.py
```diff
@@ -656,14 +657,19 @@ def __init__(
...
+        :param reg_max:                    (Deprecated) Number of regression bins. Default is None (will be inferred from model's outputs)
         :param use_batched_assignment:     Whether to use batched targets assignment or sequential (per-image).
                                            Default is True (batched).
                                            Batched assignment can be faster when number of the target per image is more or
                                            less the same across the batch, but it has higher peak GPU memory usage.
                                            Sequential assignment has lower peak GPU memory usage and preferable for cases
                                            when number of targets per image varies a lot.
         """
+        if reg_max is not None:
+            warnings.warn(
```
shaydeci, line 669: Please add proper deprecation versions info (deprecated since XXX will be removed in YYY).

shaydeci (changes requested):
Please add deprecation versions, then LGTM on my end.

src/super_gradients/training/losses/ppyolo_loss.py
BloodAxe: Fixed in 2be902acadb623fed44ca7e4b24a31059cfc8ba4

shaydeci (approved):
LGTM

PR #1737, opened 2024-01-03
PPYoloECollate to support random_aspect_ratio parameter

BloodAxe:
Added random_aspect_ratio argument to enable random resize in collate function that is not preserving aspect ratio.

### DIFF
--- a/src/super_gradients/training/utils/collate_fn/ppyoloe_collate_fn.py
+++ b/src/super_gradients/training/utils/collate_fn/ppyoloe_collate_fn.py
@@ -16,13 +16,19 @@ class PPYoloECollateFN(DetectionCollateFN):
     Collate function for PPYoloE training
     """
 
-    def __init__(self, random_resize_sizes: Union[List[int], None] = None, random_resize_modes: Union[List[int], None] = None):
+    def __init__(
+        self, random_resize_sizes: Union[List[int], None] = None, random_resize_modes: Union[List[int], None] = None, random_aspect_ratio: bool = False
+    ):
         """
-        :param random_resize_sizes: (rows, cols)
+        :param random_resize_sizes: List of sizes to randomly resize the image to. If None, will not resize.
+        :param random_resize_modes: List of interpolation modes to randomly resize the image to. If None, will not resize.
+        :param random_aspect_ratio: If True, will randomly choose both width and height from random_resize_sizes.
+                                    If False, will randomly choose only value which will be the width and height of the images.
         """
         super().__init__()
         self.random_resize_sizes = random_resize_sizes
         self.random_resize_modes = random_resize_modes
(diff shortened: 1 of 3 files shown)


src/super_gradients/training/utils/collate_fn/ppyoloe_collate_fn.py
```diff
@@ -16,13 +16,19 @@ class PPYoloECollateFN(DetectionCollateFN):
     Collate function for PPYoloE training
     """
 
-    def __init__(self, random_resize_sizes: Union[List[int], None] = None, random_resize_modes: Union[List[int], None] = None):
+    def __init__(
```
shaydeci, line 19: Maybe its a good opprotunity to make PPYoloECollateFN DetectionCollateFN since it essentially hides our multiscale feature (which is unfortunate).

src/super_gradients/training/utils/collate_fn/ppyoloe_collate_fn.py
```diff
@@ -16,13 +16,19 @@ class PPYoloECollateFN(DetectionCollateFN):
...
+        self, random_resize_sizes: Union[List[int], None] = None, random_resize_modes: Union[List[int], None] = None, random_aspect_ratio: bool = False
+    ):
         """
-        :param random_resize_sizes: (rows, cols)
+        :param random_resize_sizes: List of sizes to randomly resize the image to. If None, will not resize.
+        :param random_resize_modes: List of interpolation modes to randomly resize the image to. If None, will not resize.
+        :param random_aspect_ratio: If True, will randomly choose both width and height from random_resize_sizes.
```
shaydeci, line 25: This would be a bit more clear to me:
"If True, the height and width sizes will be drawn independently from random_resize_sizes".

shaydeci (changes requested):
I find the fact that we only support rectangular images for the multi scale a bit limiting. And now that we wish to sample random aspect ratios we are abusing the random_resize_sizes attribute.
This eventually is also limiting to always use ratios which we might not intend to in the first place when there's a large variave in random_resize_sizes.

Allow me to suggest the following:
- Introduce random_aspect_ratios instead of random_aspect_ratio.
- (Optionally now or at some point in life) allow tuples for random_resize_sizes.

Also, left a comment about pulling up everything here to DetectionCollateFN. There's no reason this good feature should be hidden from other models (:
But thats more of a note and definitely not a must for this PR.

src/super_gradients/training/utils/collate_fn/ppyoloe_collate_fn.py
BloodAxe: Updated docs in 664599da5e0148804dcd055ef68857b2ecfdeff3

shaydeci (approved):
LGTM

PR #1738, opened 2024-01-03
Allow passing additional params to .predict() for detection models

BloodAxe:
This PR expose additional parameters to `.predict()` method for better control of the predict behavior.

It also changes the output format of PPYoloE/YoloNAS models in `train` stage. Now model returns same format of the outputs in train/eval which simplify the model and makes it possible to use ExtremeBatchDetectionVisualizationCallback on train. It's just easier to push this change simultaneously as it also touches post-prediction callback

Why:
* To have fine-grained control on how many detections before and after NMS should be used
* To control class-agnostic NMS and whether multiple boxes per label are needed


src/super_gradients/training/models/detection_models/pp_yolo_e/post_prediction_callback.py
```diff
@@ -26,20 +37,21 @@ def __init__(self, score_threshold: float, nms_threshold: float, nms_top_k: int,
...
+        self.class_agnostic_nms = class_agnostic_nms
 
-    def forward(self, outputs, device: str):
+    @torch.no_grad()
+    def forward(self, outputs: Any, device: str = None):
         """
 
         :param x: Tuple of (bboxes, scores) of shape [B, Anchors, 4], [B, Anchors, C]
-        :param device:
+        :param device: (Deprecated) Not used anymore device parameter
```
shaydeci, line 47: Can we add a proper deprecation message here ?

shaydeci (commented):
Did we check if the decoding part that now is being done during training slows down the training process?
I guesss that was the thought behind not performing it.

BloodAxe:
> Did we check if the decoding part that now is being done during training slows down the training process?
> I guesss that was the thought behind not performing it.

1) We didn't
2) Not sure the speed was the intention here tbh. 

The added decoding is just a a few permute/reshape operations, meshgrid and some multiplication to decode bboxes so that should not be anything noticeable, but we can benchmark if if think it's necessary. While training YoloNAS with this regime I didn't notice any increase in training time

shaydeci (approved):
LGTM

DanTaranis:
hey how do I use this ? 

in the model.predict() ?

DanTaranis:
is this in production yet?

PR #1741, opened 2024-01-05
allow removing `anti_alias` and adding `downsample_mode` to update code

NatanBagrov:
At the moment, there is a deprecation warning when using `anti_alias`. However, the parameter has no default, thus cannot be switched with the desired update (`downsample_mode='anti_alias'`).

This fix handles it.

### DIFF
--- a/src/super_gradients/training/models/segmentation_models/unet/unet_encoder.py
+++ b/src/super_gradients/training/models/segmentation_models/unet/unet_encoder.py
@@ -122,14 +122,15 @@ class ConvBaseStage(BackboneStage, ABC):
         out_channels: int,
         stride: int,
         num_blocks: int,
-        anti_alias: bool,
+        anti_alias: Optional[bool] = None,
         downsample_mode: Optional[Union[str, DownSampleMode]] = None,
         **kwargs,
     ):
         blocks = []
         # Init down-sample module
-        if anti_alias:
+        if anti_alias is not None:  # captures `False` and `True`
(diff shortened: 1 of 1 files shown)


shaydeci (dismissed):
LGTM

src/super_gradients/training/models/segmentation_models/unet/unet_encoder.py
```diff
@@ -122,14 +122,15 @@ def build_stage(
...
-        anti_alias: bool,
+        anti_alias: Optional[bool] = None,
         downsample_mode: Optional[Union[str, DownSampleMode]] = None,
         **kwargs,
     ):
         blocks = []
         # Init down-sample module
-        if anti_alias:
+        if anti_alias is not None:  # captures `False` and `True`
             logger.warning("`anti_alias` argument is deprecated and will be removed in future versions.")
```
Louis-Dupont, line 132: It's not from you, but the current warning is not super explicit. 
It would be much better to have something like:
```
logger.warning("`anti_alias` argument is deprecated and will be removed in future versions. Please use `downsample_mode='anti_alias'` instead.")
```

NatanBagrov: [f214b0e](https://github.com/Deci-AI/super-gradients/pull/1741/commits/f214b0e15fda86c0836d50fcab927fa641f5a76d)

BloodAxe (approved):
LGTM

PR #1743, opened 2024-01-07
added albumentations tutorial notebook

shaydeci:


BloodAxe:
`! pip install git+https://github.com/Deci-AI/super-gradients.git@master` is not great. Also missing -qq
`! pip install albumentations` is also not great. 
Let's add some version specifier to albumentations and exact version for SG 🙏

BloodAxe:
Other than than - it's good to me ✅

shaydeci:
> `! pip install git+https://github.com/Deci-AI/super-gradients.git@master` is not great. Also missing -qq
> `! pip install albumentations` is also not great.
> Let's add some version specifier to albumentations and exact version for SG 🙏

Thanks, updated!
The segmentation sample + albu support is not in a published release yet so I updated the SG command to use rc tag.
Once wel release I will update the command.

Louis-Dupont (commented):
LGTM, just 2 notes about the pip install part

- `pip install git+https://github.com/Deci-AI/super-gradients.git@master` This is good for dev (since no version was released yet with the latest changes) but I don't think we want to push a notebook with this. This would make the notebook subject to breaking changes, so it's much better if we can have a fixed version. So I would prefer to
    - Release a new minor version
    - Update the notebook with this minor version, and only then merge the notebook to the repo. 

- Let's use `pip install -qq ...` for all pip install to reduce the amount of logs. (Here it's only used for `datasets`)


Minor note:
- The yaml formatting looks weird on my screen, I think it's due to the `\` in `yaml\`.

image

Louis-Dupont (approved):
LGTM

PR #1744, opened 2024-01-08
added gha trigger to release and put the unless condition under the same 'when' condition as the release

Yonatan-Kaplounov:


shaydeci (approved):
LGTM

PR #1747, opened 2024-01-08
Add support for resolving strings in PhaseCallback

shaydeci:
Added the optino to pass string that would be resolved to the right phase in phase callback.
This is mainly to support passing phases when using recipes in a convinient way.
A factory didnt seem appropriate for this.


src/super_gradients/training/utils/callbacks/base_callbacks.py
```diff
@@ -896,7 +896,18 @@ class PhaseCallback(Callback):
...
+        if isinstance(phase, str):
+            try:
+                # Convert the string to the corresponding Phase enum value
+                phase = Phase[phase]
+            except KeyError:
+                # Handle the case where the string does not match any Phase enum member
+                raise ValueError(f"Invalid phase string: '{phase}'. Must be one of: {[p.name for p in Phase]}")
+        elif not isinstance(phase, Phase):
+            # Handle the case where the phase is neither a string nor a Phase enum member
+            raise TypeError("phase must be a string or a Phase enum member")
```
Louis-Dupont, line 909: Not blocking, just small suggestion

I think would prefer to leave the responsability of this to the `Phase`, this would be a nice utils that could be used in other contexts and IMO it makes more sense to let the `Phase` class handle it anyway. 
``` python
class Phase(Enum):
    ...

    @staticmethod
    def from_string(phase_str):
        try:
            return Phase[phase_str]
        except KeyError:
            raise ValueError(f"Invalid phase string: '{phase_str}'. Must be one of: {[p.name for p in Phase]}")
```

and then
``` python

    def __init__(self, phase: Union[Phase, str]):
        if isinstance(phase, str):
            phase = Phase.from_string(phase)
        elif not isinstance(phase, Phase):
            raise TypeError("phase must be a string or a Phase enum member")
```

Or even something like

``` python

class Phase(Enum):
    ...

    @staticmethod
    def resolve(phase: Union[str, "Phase"]) -> "Phase":
        if isinstance(phase, str):
            try:
                # Convert the string to the corresponding Phase enum value
                return Phase[phase]
            except KeyError:
                # Handle the case where the string does not match any Phase enum member
                raise ValueError(f"Invalid phase string: '{phase}'. Must be one of: {[p.name for p in Phase]}")
        elif isinstance(phase, Phase):
            return phase
        else:
            raise TypeError("phase must be a string or a Phase enum member")
```
and then 
``` python
    def __init__(self, phase: Union[Phase, str]):
        phase = Phase.resolve(phase)
```

Louis-Dupont (approved):
LGTM, just left a small suggestion

Louis-Dupont (approved):
LGTM

PR #1748, opened 2024-01-09
updated returned dtype from predict

shaydeci:


BloodAxe (approved):
LGTM

BloodAxe:
Ok I actually change my mind regarding this PR :)

I think it should be done differently. Instead of modifying `_check_target_args` that are used only in `save` and `show` methods, we should instead modify the `_decode_model_output` method that fills the prediction results in the first place.

Why: By ensuring that labels as integers straight after post-prediction callback we guarantee that predicted class labels are integers in the prediction result.

```
    def _decode_model_output(self, model_output: Union[List, Tuple, torch.Tensor], model_input: np.ndarray) -> List[DetectionPrediction]:
        """Decode the model output, by applying post prediction callback. This includes NMS.

        :param model_output:    Direct output of the model, without any post-processing.
        :param model_input:     Model input (i.e. images after preprocessing).
        :return:                Predicted Bboxes.
        """
        post_nms_predictions = self.post_prediction_callback(model_output, device=self.device)

        predictions = []
        for prediction, image in zip(post_nms_predictions, model_input):
            prediction = prediction if prediction is not None else torch.zeros((0, 6), dtype=torch.float32)
            prediction = prediction.detach().cpu().numpy()
            predictions.append(
                DetectionPrediction(
                    bboxes=prediction[:, :4],
                    confidence=prediction[:, 4],
                    labels=prediction[:, 5].astype(int), # <- HERE
                    bbox_format="xyxy",
                    image_shape=image.shape,
                )
            )

        return predictions
```

BloodAxe:
UPD: And maybe add a unit test to ensure labels in predict result are integers

BloodAxe (approved):
LGTM

PR #1749, opened 2024-01-09
Improve YoloNAS model flexibility

BloodAxe:
This PR adds a few features that improves the flexibility of YoloNAS:

* Allow `context` module of the YoloNAS backbone to be optional
* Allow YoloNASStem to support user-defined stride
* Allow setting upsample mode in YoloNASUpStage


src/super_gradients/modules/detection_modules.py
```diff
@@ -76,13 +81,12 @@ def out_channels(self):
         return self._out_channels
 
     def forward(self, x):
-
+        y = x
```
shaydeci, line 84: Don't think we really need this line right ?

src/super_gradients/modules/sampling.py
```diff
@@ -33,6 +33,53 @@ def make_upsample_module(scale_factor: int, upsample_mode: Union[str, UpsampleMo
     return module
 
 
+def make_upsample_module_with_explicit_channels(
```
shaydeci, line 36: Seems to me like we can extend make_upsample_module functinoality to support exlicit channels, can't we ?

shaydeci (changes requested):

src/super_gradients/modules/sampling.py
BloodAxe: I doubt. First, the pixel shuffle implementation differs for this method. The new method uses 1x1 conv + pixel shuffle if channel combination does not match while other method doesn't.
Secondly the new method requires in/out channels and changing the signature in all places where it is used would be quite a task (Since in/output channels number may not be that available). 
Lastly the old method used not only in this repository, so this could be also a BC.

src/super_gradients/modules/detection_modules.py
BloodAxe: Indeed. Fixed in 652d3d267beecd50e214d60ad03c3d330baafd6c

shaydeci (approved):
LGTM

PR #1751, opened 2024-01-10
Fix import issue in convert_recipe_to_code

BloodAxe:
This PR fixes two issues in `convert_recipe_to_code`

1) There was a bug with `_target_: numpy.arange` and similar constructions caused by missing import statement. Now the code adds necessary imports to generated code based on present `_target_`'s

2) There also was a bug with string parameters not properly quoted which is also fixed

### DIFF
--- a/src/super_gradients/convert_recipe_to_code.py
+++ b/src/super_gradients/convert_recipe_to_code.py
@@ -225,9 +225,9 @@ def convert_recipe_to_code(config_name: Union[str, pathlib.Path], config_dir: Un
 
     cfg.checkpoint_params = hydra.utils.instantiate(cfg.checkpoint_params)
 
-    # strict_load = cfg.checkpoint_params.strict_load
-    # if isinstance(strict_load, Mapping) and "_target_" in strict_load:
-    #     strict_load = hydra.utils.instantiate(strict_load)
+    strict_load = cfg.checkpoint_params.strict_load
+    if isinstance(strict_load, Mapping) and "_target_" in strict_load:
+        strict_load = hydra.utils.instantiate(strict_load)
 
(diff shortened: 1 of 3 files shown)


src/super_gradients/convert_recipe_to_code.py
```diff
@@ -57,45 +57,125 @@ def try_import_black():
...
+    """
+    if path == "":
+        raise ImportError("Empty path")
+    from importlib import import_module
+
+    parts = path.split(".")
+    for part in parts:
+        if not len(part):
+            raise ValueError(f"Error loading '{path}': invalid dotstring. Relative imports are not supported.")
+    assert len(parts) > 0
```
Louis-Dupont, line 165: Minor, but why do we need this ? If it fails, the error would be super unintuitive for the user. So I assume it should never be true and is here more as self-documenting ?

Louis-Dupont (approved):
LGTM, just minor question that you can skip

src/super_gradients/convert_recipe_to_code.py
BloodAxe: Basically it was like that in hydra. 
It is to prevent user attempt to use `_target_: .something.foo` or import from parent (`..something.else`)

Louis-Dupont: Mmmhh but if it happens the user would get a cryptic error, no ?
Wouldnt it be better to at least write this in the assert error message ?

Louis-Dupont (dismissed):
LGTM (just check my comment before merging, in case you think it's relevant)

Louis-Dupont (approved):
LGTM

Louis-Dupont (approved):
LGTM

PR #1752, opened 2024-01-10
Improve DetectionMetrics implementation to properly handle score_thresholds

BloodAxe:
This PR changes the default value of `calc_best_score_thresholds` of `DetectionMetrics` and inheriting classes to `True` which adds additional metric value logged - an optimal confidence threshold value that maximized F1 score.

And as a bonus it fixes a bug in implementation of the metric to work correctly when custom class names are set.


shaydeci (approved):
Nice one! LGTM

PR #1755, opened 2024-01-12
Fix DP

Louis-Dupont:
Note: I want to run the recipe sanity check on this, but they are still removed from the CI


BloodAxe (approved):
LGTM

PR #1756, opened 2024-01-12
changes logic from only one to exactly one parameter for step lr

NatanBagrov:
This fix makes sure that either `lr_updates` contains something (non-empty list), or `step_lr_update_freq` exists, but not both.

In case where `lr_updates` is not set then it defaults to an empty list, which has no impact on the scheduler - but the code works -- not the best UX

### DIFF
--- a/src/super_gradients/training/utils/callbacks/callbacks.py
+++ b/src/super_gradients/training/utils/callbacks/callbacks.py
@@ -406,7 +406,7 @@ class StepLRScheduler(LRCallbackBase):
             )
 
         if step_lr_update_freq is None and len(lr_updates) == 0:
-            raise ValueError("At least one of [lr_updates, step_lr_update_freq] parameters should be passed" f" to {StepLRScheduler.__name__} constructor")
+            raise ValueError(f"At least one of [lr_updates, step_lr_update_freq] parameters should be passed to {StepLRScheduler.__name__} constructor")
 
         if step_lr_update_freq:
             max_epochs = self.training_params.max_epochs - self.training_params.lr_cooldown_epochs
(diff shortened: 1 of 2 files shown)


src/super_gradients/training/utils/callbacks/callbacks.py
```diff
@@ -399,8 +399,8 @@ class StepLRScheduler(LRCallbackBase):
 
     def __init__(self, lr_updates, lr_decay_factor, step_lr_update_freq=None, **kwargs):
         super().__init__(Phase.TRAIN_EPOCH_END, **kwargs)
-        if step_lr_update_freq and len(lr_updates):
-            raise ValueError("Only one of [lr_updates, step_lr_update_freq] should be passed to StepLRScheduler constructor")
+        if not ((step_lr_update_freq is None) ^ (len(lr_updates) == 0)):
+            raise ValueError(f"Exactly one of [lr_updates, step_lr_update_freq] should be passed to {StepLRScheduler.__name__} constructor")
```
BloodAxe, line 403: I think exception message can be a bit improved to show two different messages:

```
if step_lr_update_freq and len(lr_updates):
    raise ValueError("Parameters lr_updates and step_lr_update_freq are mutually exclusive and cannot be passed to StepLRScheduler constructor simultaneously")

if step_lr_update_freq is None and len(lr_updates) == 0:
    raise ValueError(f"At least one of [lr_updates, step_lr_update_freq] parameters should be passed to StepLRScheduler constructor")
```

BloodAxe (dismissed):
LGTM

src/super_gradients/training/utils/callbacks/callbacks.py
NatanBagrov: Sure, why not, [f4f4afb](https://github.com/Deci-AI/super-gradients/pull/1756/commits/f4f4afbd07ec45c7575e841ee502966bd9b079e2)

BloodAxe (approved):
LGTM

BloodAxe (approved):
LGTM

PR #1757, opened 2024-01-15
Fix potential unwanted simplification of ONNX graph when exported model with random weights

BloodAxe:
# Problem

A RepVGG block has default alpha value of 1. This parameter controls the scalar multiply factor for residual branch. During NAS benchmarking of the model usually done using randomly initialized weights.
In this case a default value of 1 is simplified by ONNX so operation `alpha * x` reduced to simple addition. Elimination of multiplication makes the benchmark result a bit too optimistic. 

After model has been trained this alpha most likely will have the value that is different from 1 and hence the graph will contain the multiplication.

# Solution

Adds a small noise to default value such that multiplication is not eliminated from the graph for model with random weights.

### DIFF
--- a/src/super_gradients/modules/qarepvgg_block.py
+++ b/src/super_gradients/modules/qarepvgg_block.py
@@ -147,7 +147,14 @@ class QARepVGGBlock(nn.Module):
             self.identity = None
 
         if use_alpha:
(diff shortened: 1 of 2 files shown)


shaydeci (approved):
LGTM

PR #1758, opened 2024-01-15
fixed by casting training params and one hot dtype fixed

shaydeci:
The schema for initial_lr fails when training with yaml files since the optimizer_params are of DictConfig instance.
I casted th training_params to dict before we pass the to TrainingParams as entries.
Also I found a small bug when trying to run the ddrnet recipe - onehot crashes when target is uint.


BloodAxe (approved):
LGTM

PR #1772, opened 2024-01-18
Feature/sg 1292 compute roc pr curve

Louis-Dupont:
- Adding `MetricOutput` typing which represents the Metric output we support in SG.
- Introducing `PlottableMetricOutput which can be returned from Metrics
- Introducing functions that can filter the metric output based on whether we want only scalars (to log/print) or graphs (to save as image) 

@BloodAxe I kept the `scalar` attribute because being able to support a metric that has no scalar associated is much more complex/risky than it seems.


src/super_gradients/common/sg_loggers/utils.py
```diff
@@ -0,0 +1,171 @@
+from dataclasses import dataclass
+from typing import List, Optional, Tuple
+
+import numpy as np
+
+
+@dataclass
+class PlottableMetricOutput:
```
BloodAxe, line 8: I think we should not make this class responsible for plotting itself.

We cannot see all the use-cases right now. Ex: may want to draw multiple lines on same plot (PR curve per each class). Or histograms or something else. For each of these use scenarios we would need to modify this class. 

Instead we should render the plot inside the metric as a regular image and return it `PlottableMetricOutput(value=image)`. So logic of generation of the image becomes responsibility of the Metric and not the class itself or logger.

Louis-Dupont: @BloodAxe after our discussion I changed the code accordingly. See new description and new code

src/super_gradients/training/metrics/detection_metrics.py
```diff
@@ -38,16 +41,16 @@ class DetectionMetrics(Metric):
                                             Could be either instance of IouThreshold, a tuple (lower bound, upper_bound) or single scalar.
     :param recall_thres:                    Recall threshold to compute the mAP.
     :param score_thres:                     Score threshold to compute Recall, Precision and F1.
-    :param top_k_predictions:               Number of predictions per class used to compute metrics, ordered by confidence score
+    :param top_k_predictions:               Number of predictions per class used to compute metric_outputs, ordered by confidence score
```
BloodAxe, line 44: This is not necessary renaming. I think it's better to keep the `metrics` term as is in docstrings

src/super_gradients/common/sg_loggers/metric_outputs.py
```diff
@@ -0,0 +1,38 @@
+from dataclasses import dataclass
+from typing import Union, Dict
+
+import numpy as np
+import torch
+
+
+@dataclass
+class PlottableMetricOutput:
```
BloodAxe, line 9: Thanks for making the changes we talked about 🙏 
I feel we are still not quite there yet tho) So my take on this class is this:

```
from typing import TypeAlias

@dataclass
class PlottableMetricOutput:
    """Metric output that can is represented as a graph.
    :attr image:    The image of the graph representing the metric value. 
                    This image will be logged as an image.
    """

    image: np.ndarray

ScalarOutput: TypeAlias = Union[float, int, torch.Tensor]

MetricOutput = Union[PlottableMetricOutput, ScalarOutput]
```

Imagine a case when you want to generate confusion matrix. What `scalar` value would correspond here?

So we should not be storing anything but `image` in `PlottableMetricOutput`. If we want to log area under curve for some reason - this should be logged as an additional value to the metric dict.

If we change the dataclass to suggested above, the `get_scalar_metric_output` can go.

src/super_gradients/recipes/training_hyperparams/coco2017_yolox_train_params.yaml
```diff
@@ -1,7 +1,7 @@
 defaults:
   - default_train_params
 
-max_epochs: 300
+max_epochs: 1
```
BloodAxe, line 4: Don't forget to revert back when done testing)

src/super_gradients/training/metrics/detection_metrics.py
```diff
@@ -199,16 +203,20 @@ def update(self, preds, target: torch.Tensor, device: str, inputs: torch.tensor,
...
         mean_ap, mean_precision, mean_recall, mean_f1, best_score_threshold = -1.0, -1.0, -1.0, -1.0, -1.0
         accumulated_matching_info = getattr(self, self.state_key)
         best_score_threshold_per_cls = np.zeros(self.num_cls)
         mean_ap_per_class = np.zeros(self.num_cls)
 
+        precision, recall = [], []
+
         if len(accumulated_matching_info):
-            matching_info_tensors = [torch.cat(x, 0) for x in list(zip(*accumulated_matching_info))]
+            import torch
```
BloodAxe, line 217: Import should be probably done on top-level

src/super_gradients/common/sg_loggers/metric_outputs.py
```diff
@@ -0,0 +1,38 @@
...
+
+MetricOutput = Union[PlottableMetricOutput, float, int, torch.Tensor]
+
+
+def get_scalar_metric_output(metric_output: MetricOutput) -> Union[float, int]:
+    return metric_output.scalar if isinstance(metric_output, PlottableMetricOutput) else metric_output
+
+
+def get_scalar_metric_outputs(metric_outputs: Dict[str, MetricOutput]) -> Dict[str, Union[float, int]]:
+    return {name: get_scalar_metric_output(metric_output=value) for name, value in metric_outputs.items()}
```
BloodAxe, line 34: Similarly to `get_plottable_metric_outputs` the implementation could be:
```
return {k: v for k, v in metric_outputs.items() if isinstance(v, ScalarOutput)}
```

Louis-Dupont: Changed as part of the big change 42ea92e2300738d1e4915d623fa46b52305b4821

src/super_gradients/training/metrics/detection_metrics.py
Louis-Dupont: Sure, removed

src/super_gradients/recipes/training_hyperparams/coco2017_yolox_train_params.yaml
Louis-Dupont: Sure, removed

src/super_gradients/common/sg_loggers/metric_outputs.py
Louis-Dupont: Changed here https://github.com/Deci-AI/super-gradients/commit/42ea92e2300738d1e4915d623fa46b52305b4821

src/super_gradients/training/metrics/detection_metrics.py
Louis-Dupont: This was a mistake, I rolledback

BloodAxe (changes requested):
This PR needs some changes.
Currently there are some strange scalars with same names as plot metrircs are logged to TB which needs to be investigated

PR #1779, opened 2024-01-23
Fixes for 3.6.0 release

shaydeci:
- Fixes for recipe tests: dtype fix for final transform in the segmentation recipes + added dtype long in dtype factory.

- Rewritten the notebook test + parallelization for each notebook (except for excluded which we need to take care of).


BloodAxe (approved):
LGTM

PR #1785, opened 2024-01-24
Deprecate cache and cache_dir parameters support in DetectionDataset

BloodAxe:
Some groundwork PR before actually reaching pycocotools removal.

This PR removes support of caching images (Saving resized copies to cache_dir). 
The goal is to get rid of unused code pieces and trim fat from DetectionDataset as it's already too big.


Louis-Dupont (commented):
LGTM and I agree with your view. I'm not approving just because I want @shaydeci to also give his opinion on that

shaydeci (approved):
LGTM

PR #1786, opened 2024-01-24
Deprecate tight_box_rotation parameters in COCODetectionDataset

BloodAxe:
None of the recipes use this parameter.
Currently this parameter cannot be used anyway since we don't propagate segmentation masks to DetectionSample. Albumentations does not support this feature either.


src/super_gradients/training/datasets/detection_datasets/coco_format_detection.py
```diff
@@ -28,24 +27,26 @@ def __init__(
...
         :param data_dir:                Where the data is stored.
         :param json_annotation_file:    Name of the coco json file. Path relative to data_dir.
         :param images_dir:              Name of the directory that includes all the images. Path relative to data_dir.
-        :param tight_box_rotation:      bool, whether to use of segmentation maps convex hull as target_seg
-                                            (check get_sample docs).
         :param with_crowd:              Add the crowd groundtruths to __getitem__
         :param class_ids_to_ignore:     List of class ids to ignore in the dataset. By default, doesnt ignore any class.
+        :param tight_box_rotation:      This parameter is deprecated and will be removed in a SuperGradients 3.8.
         """
+        if tight_box_rotation is not None:
```
ofrimasad, line 44: can you please implement a decorator for this: 
```
@deprecated_argument(name='tight_box_rotation', deprecated_v='3.6.0', remove_v='3.9.0', message='....')
```

ofrimasad (commented):
can you please implement a decorator for deprecated_argument
I though we already have it...

src/super_gradients/training/datasets/detection_datasets/coco_format_detection.py
BloodAxe: We have decorator for method, but not for an argument. Kk, will add

BloodAxe: Added in 24cf66ca088ab0f7a5a84bdc8b8d834496e6683a

ofrimasad (approved):
LGTM

PR #1789, opened 2024-01-25
Bug/sg 1247 reoarganize tests

shaydeci:
Moved some tests that were not triggered (i.e not part of the suite) to integration tests. Other ones I either removed completely (there were some really old ones that were completely irrelevant). Theres one test I for shlefnet that had part of it that was failing (again, not part of the suite and has not been running for ages...added a fixme comment for now).

Paralleized the unit tests for the build. We dont use the suite any more but I kept it as a convinient means to run them in an IDE.

Build now takes half the time (50 minutes -> ~25 minutes).


.circleci/config.yml
```diff
@@ -142,6 +142,7 @@ commands:
 
 jobs:
   build:
+    parallelism: 16  # Adjust based on your needs
```
Louis-Dupont, line 145: Why 16 ? Is it the number of tests or just a good-enough value ?

Louis-Dupont (approved):
LGTM, just small question out of curiosity

.circleci/config.yml
BloodAxe: I this is max number of nodes Ran allowed ^_^

PR #1791, opened 2024-01-25
Remove pycocotools

BloodAxe:
This PR removes the pycocotools dependency from SG
It replace it with in-house parser of COCO annotation that is identical to COCO implementation

Contains changes from https://github.com/Deci-AI/super-gradients/pull/1786 so better to be merged second


shaydeci (changes requested):

src/super_gradients/training/utils/detection_utils.py
```diff
@@ -170,9 +170,9 @@ def calc_bbox_iou_matrix(pred: torch.Tensor):
     return ious
 
 
-def change_bbox_bounds_for_image_size(boxes: np.ndarray, img_shape: Tuple[int, int]) -> np.ndarray:
+def change_bbox_bounds_for_image_size_inplace(boxes: np.ndarray, img_shape: Tuple[int, int]) -> np.ndarray:
```
shaydeci, line 173: Although low chance, this might interfere with anyone expecting the old behaviour when calling the old method and expecting it to be inplace...
Perhaps a flag with the old behaviour as default would be preferable.

.circleci/config.yml
```diff
@@ -539,6 +540,7 @@ jobs:
             source << parameters.sg_new_env_name >>/bin/activate
             python3.8 -m pip install --upgrade setuptools pip wheel
             python3.8 -m pip install -r requirements.txt
+            python3.8 -m pip install -r requirements.dev.txt
```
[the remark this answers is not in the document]
BloodAxe: We have a few test checking the scores of our mAP implementation vs COCOEval. For this reason we need pycocotools installed when running tests. So that's why it is added to dev dependencies

src/super_gradients/training/utils/detection_utils.py
BloodAxe: Thanks! True, keeping a BC is a good suggestions. Fixed in fa50ec998ae65c2997a5f3dff3a6cbf729590827

shaydeci (approved):
LGTM

PR #1824, opened 2024-02-09
Added explicit antialias=False to ensure we can export torchvision Resize to ONNX

BloodAxe:
In https://github.com/pytorch/vision/releases/tag/v0.17.0 release a default value of antialias argument was changed. New value is incompatible with ONNX models.
So this PR adds explicit old value to keep tests green


shaydeci (approved):
LGTM

PR #1840, opened 2024-02-16
allow flexibility to provide absolute path to annotations

NatanBagrov:
Why: in cases where the data is stored in a read-only directory, we would like to be flexible with different annotation files (e.g., pseudo-labels) - not having to put them under the `data_dir` (which we might not have access to).


BloodAxe (approved):
LGTM

PR #1844, opened 2024-02-18
Added a class balanced distributed sampler

NatanBagrov:
**Context:** in cases where some classes are less frequent than others, we'd want to sample images containing these classes more often.

**Current implementation:**
1. `IndexMappingDatasetWrapper` - gets a dataset and a list of indices (`mapping`), and wraps the dataset: `wrapper[i] == self.dataset[self.mapping[item]]`
2. An implementation of https://arxiv.org/pdf/1908.03195.pdf, that returns a float per index (image) indicating the scarcity of classes that are in that image (larger = more scarce = repeat it more often).
3. `DetectionClassBalancedDistributedSampler` that uses (1) with (2) for Detection datasets.

**Discussion:**
Current implementation supports distributed mode. Non-distributed mode can also be supported, but will result more code.
Some alternative implementations I can spot:
1. Do not wrap original dataset, but rather override `__iter__` from `DistributedSampler` - it is possible, however, causes coupling and code duplication. Current impelentation (`super().__init__(dataset=IndexMappingDatasetWrapper(dataset, repeat_indices), *args, **kwargs)`) is a bit less verbose.
2. Eventually, we might use `WeightedSampler`, however, it is not supported in distributed mode. There are some implementations of `DistributedWeightedSampler` in-the-wild. We can try these.
3. Perhaps we should consider a different approach where the heavy lifting is on the dataset size. For example:
    ```yaml:
    dataset: class_balanced_wrapper
       inner_dataset: coco
       ...
    ```
    The issue, however, is that we cannot inherit `inner_class`, no inherit `DetectionDataset` because we don't want to pass all the `__init__` parameters.
 4. Another option is to modify `DetectionDataset` and add a parameter for class balancing (i.e., `oversampling_threshold`) and do the heavy lifting inside `DetectionDataset`. This is less verbose, however, adds more responsibility to `DetectionDataset` (compared to Composition)

Open for discussion :)

### DIFF
--- a/src/super_gradients/training/datasets/balancing_classes_utils.py
+++ b/src/super_gradients/training/datasets/balancing_classes_utils.py
@@ -1,0 +1,88 @@
+import math
+from typing import Callable, List, Optional
+
+import numpy as np
+from torch.utils.data import Dataset
+
+
+def _default_oversample_heuristic(cat_id, cat_freq, oversample_threshold):
+    """
+    How to oversample a category. This heuristic is based on https://arxiv.org/pdf/1908.03195.pdf.
+    A value of 1.0 corresponds to no oversampling.
+    """
+    return max(1.0, math.sqrt(oversample_threshold / cat_freq))
+
+
+def get_repeat_factors(
+    *,
+    index_to_classes: Callable[[int], List[int]],
+    num_classes: int,
+    dataset_length: int,
+    ignore_empty_annotations: bool = False,
+    oversample_threshold: Optional[float] = None,
+    oversample_heuristic: Optional[Callable[[int, int, float], float]] = _default_oversample_heuristic,
+) -> List[float]:
+    """
+    Oversampling scarce classes from detection dataset, following sampling strategy described in https://arxiv.org/pdf/1908.03195.pdf.
+    Implementation based on https://github.com/open-mmlab/mmdetection (Apache 2.0 license).
+
+    :param index_to_classes: A function that maps an index to a list of classes (labels).
+    :param num_classes: Number of classes.
+    :param dataset_length: Length of the dataset.
+    :param ignore_empty_annotations: Whether to ignore empty annotations.
+    :param oversample_threshold: A frequency threshold (fraction, 0-1). Classes that are *less frequent* than this threshold will be oversampled.
+                                    Default: None. If None, the median of the class frequencies will be used.
+                                    See step (1) below to understand how frequencies are computed.
+
+
+    The repeat factor is computed as followed.
+    1. For each category c, compute the fraction # of images that contain it (its frequency): :math:`f(c)`
+    2. For each category c, compute the category-level repeat factor: :math:`r(c) = max(1, sqrt(t/f(c)))`
+    3. For each image I, compute the image-level repeat factor: :math:`r(I) = max_{c in I} r(c)`
+
+    Returns a list of repeat factors (length = dataset_length). How to read: result[i] is a float, indicates the repeat factor of image i.
+    """
+
+    # 0. pre-fetch the categories to reduce time
+    sample_id_to_categories = {idx: index_to_classes(idx) for idx in range(dataset_length)}
+
+    # 1. For each category c, compute the fraction # of images that contain it: f(c)
+    category_freq = dict()
+    for idx in range(dataset_length):
+        cat_ids = sample_id_to_categories[idx]
+        if len(cat_ids) == 0 and not ignore_empty_annotations:
+            cat_ids = {num_classes}
+        for cat_id in cat_ids:
+            category_freq[cat_id] = category_freq.get(cat_id, 0) + 1
+    for k, v in category_freq.items():
+        category_freq[k] = v / dataset_length
+
+    # 2. For each category c, compute the category-level repeat factor: r(c) = max(1, sqrt(t/f(c)))
+    if oversample_threshold is None:
+        oversample_threshold = np.median(list(category_freq.values()))
+    category_repeat = {cat_id: oversample_heuristic(cat_id, cat_freq, oversample_threshold) for cat_id, cat_freq in category_freq.items()}
+
+    # 3. For each image I, compute the image-level repeat factor: r(I) = max_{c in I} r(c)
+    repeat_factors = []
+    for idx in range(dataset_length):
+        cat_ids = sample_id_to_categories[idx]
+        if len(cat_ids) == 0:  # no annotations for this image
+            repeat_factors.append(1.0 * (not ignore_empty_annotations))  # 0.0 if we ignore, 1.0 otherwise. We are not oversampling empty annotations.
+            continue
+        repeat_factor = max({category_repeat[cat_id] for cat_id in cat_ids})
+        repeat_factors.append(repeat_factor)
+
+    return repeat_factors
+
+
+class IndexMappingDatasetWrapper(Dataset):
+    def __init__(self, dataset: Dataset, mapping: List[int]):
+        super().__init__()
+        self.dataset = dataset
+        self.mapping = mapping
+
+    def __getitem__(self, item):
+        return self.dataset[self.mapping[item]]
+
+    def __len__(self):
+        return len(self.mapping)
--- a/src/super_gradients/training/datasets/detection_datasets/detection_dataset_class_balancing_wrapper.py
+++ b/src/super_gradients/training/datasets/detection_datasets/detection_dataset_class_balancing_wrapper.py
@@ -1,0 +1,34 @@
+import math
+from functools import partial
+from typing import Optional, List
+from torch.utils.data import DistributedSampler
+
+from super_gradients.common.object_names import Samplers
+from super_gradients.common.registry import register_sampler
+from super_gradients.training.datasets import DetectionDataset
+from super_gradients.training.datasets.balancing_classes_utils import get_repeat_factors, IndexMappingDatasetWrapper
+
+
+def extract_labels_from_target(dataset: DetectionDataset, idx: int) -> List[int]:
+    return dataset._load_sample_annotation(idx)["target"][:, -1]
+
+
+@register_sampler(Samplers.DISTRIBUTED_DETECTION_CLASS_BALANCING)
+class DetectionClassBalancedDistributedSampler(DistributedSampler):
+    """
+    DetectionClassBalancedDistributedSampler is a distributed sampler that over-samples scarce classes in detection datasets.
+    """
+
+    def __init__(self, dataset: DetectionDataset, oversample_threshold: Optional[float] = None, *args, **kwargs) -> None:
+        repeat_factors = get_repeat_factors(
+            index_to_classes=partial(extract_labels_from_target, dataset),
+            num_classes=len(dataset._all_classes),
+            dataset_length=len(dataset),
+            ignore_empty_annotations=dataset.ignore_empty_annotations,
+            oversample_threshold=oversample_threshold,
+        )
+        repeat_indices = []
+        for dataset_idx, repeat_factor in enumerate(repeat_factors):
+            repeat_indices.extend([dataset_idx] * math.ceil(repeat_factor))
+
+        super().__init__(dataset=IndexMappingDatasetWrapper(dataset, repeat_indices), *args, **kwargs)
--- a/tests/unit_tests/class_balancing_test.py
+++ b/tests/unit_tests/class_balancing_test.py
@@ -1,0 +1,110 @@
+import numpy as np
+import torch
+import unittest
+
+from torch.utils.data import Dataset
+
+from super_gradients.training.datasets.balancing_classes_utils import get_repeat_factors
+
+
+class SingleLabelUnbalancedDataset(Dataset):
+    def __init__(self, num_classes: int) -> None:
+        super().__init__()
+        self.num_classes = num_classes
+        self.ignore_empty_annotations = False
+
+    def __len__(self) -> int:
+        return self.num_classes
+
+    def __getitem__(self, idx: int) -> torch.Tensor:
+        return torch.tensor([idx] * idx)  # no class 0
+
+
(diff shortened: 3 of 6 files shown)


src/super_gradients/training/datasets/detection_datasets/detection_dataset_class_balancing_wrapper.py
```diff
@@ -0,0 +1,34 @@
...
+from super_gradients.training.datasets import DetectionDataset
+from super_gradients.training.datasets.balancing_classes_utils import get_repeat_factors, IndexMappingDatasetWrapper
+
+
+def extract_labels_from_target(dataset: DetectionDataset, idx: int) -> List[int]:
+    return dataset._load_sample_annotation(idx)["target"][:, -1]
+
+
+@register_sampler(Samplers.DISTRIBUTED_DETECTION_CLASS_BALANCING)
+class DetectionClassBalancedDistributedSampler(DistributedSampler):
```
shaydeci, line 17: We have a dedicated module for our samplers, please move it there.

src/super_gradients/training/datasets/balancing_classes_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        if len(cat_ids) == 0:  # no annotations for this image
+            repeat_factors.append(1.0 * (not ignore_empty_annotations))  # 0.0 if we ignore, 1.0 otherwise. We are not oversampling empty annotations.
+            continue
+        repeat_factor = max({category_repeat[cat_id] for cat_id in cat_ids})
+        repeat_factors.append(repeat_factor)
+
+    return repeat_factors
+
+
+class IndexMappingDatasetWrapper(Dataset):
```
shaydeci, line 78: I feel a bit unease about having a different dataset passed to the sampler then what we actually have in our data loader.
If anything happens under the hood (which I think does) things could go wrong.
I got a feeling @BloodAxe would be on the same page as me here, but what do you say ?

tests/unit_tests/class_balancing_test.py
```diff
@@ -0,0 +1,110 @@
...
+        self.ignore_empty_annotations = False
+
+    def __len__(self) -> int:
+        return self.num_classes
+
+    def __getitem__(self, idx: int) -> torch.Tensor:
+        return torch.tensor([idx, 0])  # class 0 appears everywhere, other classes appear only once.
+
+
+class ClassBalancingTest(unittest.TestCase):
```
shaydeci, line 36: Please add to suite.

tests/unit_tests/dataset_index_mapping_test.py
```diff
@@ -0,0 +1,42 @@
...
+        self.ignore_empty_annotations = False
+
+    def __len__(self) -> int:
+        return self.num_classes
+
+    def __getitem__(self, idx: int) -> torch.Tensor:
+        return torch.tensor(idx)
+
+
+class DatasetIndexMappingTest(unittest.TestCase):
```
shaydeci, line 22: Please add to suite.

tests/unit_tests/detection_class_balancing_distributed_sampler_test.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy as np
+import torch
+import unittest
+
+from torch.utils.data import Dataset
+
+from super_gradients.training.datasets.detection_datasets.detection_dataset_class_balancing_wrapper import DetectionClassBalancedDistributedSampler
+
+
+class DummyDetectionDataset(Dataset):  # NOTE: we implement the needed stuff from DetectionDataset, but we do not inherit it because the ctor is massive
```
shaydeci, line 13: Please add to suite.

tests/unit_tests/class_balancing_test.py
```diff
@@ -0,0 +1,110 @@
+import numpy as np
+import torch
+import unittest
+
+from torch.utils.data import Dataset
+
+from super_gradients.training.datasets.balancing_classes_utils import get_repeat_factors
+
+
```
shaydeci, line 9: I think a comprehensive test that actually runs in DDP is missing.
Please add one (can be done through the ci config - see sanity_tests workflow).

src/super_gradients/training/datasets/balancing_classes_utils.py
```diff
@@ -0,0 +1,88 @@
...
+                                    Default: None. If None, the median of the class frequencies will be used.
+                                    See step (1) below to understand how frequencies are computed.
+
+
+    The repeat factor is computed as followed.
+    1. For each category c, compute the fraction # of images that contain it (its frequency): :math:`f(c)`
+    2. For each category c, compute the category-level repeat factor: :math:`r(c) = max(1, sqrt(t/f(c)))`
+    3. For each image I, compute the image-level repeat factor: :math:`r(I) = max_{c in I} r(c)`
+
+    Returns a list of repeat factors (length = dataset_length). How to read: result[i] is a float, indicates the repeat factor of image i.
```
shaydeci, line 43: Out of curiosity - I guess  this would mean that using Mosaic / mixup would mess things up here?
We use them in most of our recipes, so maybe there's something to tak in account here.

src/super_gradients/training/datasets/detection_datasets/detection_dataset_class_balancing_wrapper.py
```diff
@@ -0,0 +1,34 @@
...
+            num_classes=len(dataset._all_classes),
+            dataset_length=len(dataset),
+            ignore_empty_annotations=dataset.ignore_empty_annotations,
+            oversample_threshold=oversample_threshold,
+        )
+        repeat_indices = []
+        for dataset_idx, repeat_factor in enumerate(repeat_factors):
+            repeat_indices.extend([dataset_idx] * math.ceil(repeat_factor))
+
+        super().__init__(dataset=IndexMappingDatasetWrapper(dataset, repeat_indices), *args, **kwargs)
```
shaydeci, line 34: Following my precious comment regarding the dataset wrapper, I am in favor moving the logic introduced there (mapping etc) to this class.
Also the helper methods which I don't think have much context outside this speific sampler.

shaydeci (changes requested):

src/super_gradients/training/datasets/balancing_classes_utils.py
NatanBagrov: So, you are half correct :) I believe things should work ok because the balancing is done (currently) in the sampler-level, while mixup/mosaic are dataset-level. On the other hand, the sampler will "balance" and ask for more indices with scarce classes (this is good), but the dataset does not know that it should sample non uniformly, thus 3/4 images will be taken at random, regardless of class scarcity. Hope it makes sense...

tests/unit_tests/class_balancing_test.py
NatanBagrov: Pff.. I agree. I'll try to implement it once we agree on the approach (i.e., is will be redundant if we decide on option 4)

BloodAxe:
First of all - great initiative on adding a samplers support! I feel this can be really helpful in many cases.
Regarding the suggested implementation - I do have a few comments:

1) I will start with a most simple one. I think we should have a generic wrapper around any sampler for DPP case: `DistributedSamplerWrapper`

Suppose we have a sampler (any subclass of a `Sampler`), then for DDP case we would do:
`sampler = DistributedSamplerWrapper(sampler)` and that would make `sampler` compatible with DDP.

The suggested design of wrapper was proposed on pytorch forums, been tested for years and myself and does not require any special knowledge of underlying sampler implementation.

```
class DatasetFromSampler(Dataset):
    """Dataset to create indexes from `Sampler`.

    Args:
        sampler: PyTorch sampler
    """

    def __init__(self, sampler: Sampler):
        """Initialisation for DatasetFromSampler."""
        self.sampler = sampler
        self.sampler_list = None

    def __getitem__(self, index: int):
        """Gets element of the dataset.

        Args:
            index: index of the element in the dataset

        Returns:
            Single element by index
        """
        if self.sampler_list is None:
            self.sampler_list = list(self.sampler)
        return self.sampler_list[index]

    def __len__(self) -> int:
        """
        Returns:
            int: length of the dataset
        """
        return len(self.sampler)


class DistributedSamplerWrapper(DistributedSampler):
    """
    Wrapper over `Sampler` for distributed training.
    Allows you to use any sampler in distributed mode.

    It is especially useful in conjunction with
    `torch.nn.parallel.DistributedDataParallel`. In such case, each
    process can pass a DistributedSamplerWrapper instance as a DataLoader
    sampler, and load a subset of subsampled data of the original dataset
    that is exclusive to it.

    .. note::
        Sampler is assumed to be of constant size.
    """

    def __init__(
        self,
        sampler,
        num_replicas: Optional[int] = None,
        rank: Optional[int] = None,
        shuffle: bool = True,
    ):
        """

        Args:
            sampler: Sampler used for subsampling
            num_replicas (int, optional): Number of processes participating in
              distributed training
            rank (int, optional): Rank of the current process
              within ``num_replicas``
            shuffle (bool, optional): If true (default),
              sampler will shuffle the indices
        """
        super(DistributedSamplerWrapper, self).__init__(
            DatasetFromSampler(sampler),
            num_replicas=num_replicas,
            rank=rank,
            shuffle=shuffle,
        )
        self.sampler = sampler

    def __iter__(self):

        self.dataset = DatasetFromSampler(self.sampler)
        indexes_of_indexes = super().__iter__()
        subsampler_indexes = self.dataset
        return iter(itemgetter(*indexes_of_indexes)(subsampler_indexes))
```

BloodAxe:
Secondly, I think we should introduce some sort of the interface:

```
class HasSamplingInformation(ABC):
   def getLabelsPresense() -> np.ndarray:
      """
      :returns: A Numpy array of [Dataset Length, Num Classes] with values corresponding to
                a number of objects at the current sample index.
      """
```

Why we want to have interface? 
* We can check whether dataset we are trying to use with sampler implements it (If not - we show nice error).
* Sampler implementation gets a presence matrix which contains ALL necessary information to do sampling. No remapping indexes and other stuff.

shaydeci:
> First of all - great initiative on adding a samplers support! I feel this can be really helpful in many cases. Regarding the suggested implementation - I do have a few comments:
> 
> 1. I will start with a most simple one. I think we should have a generic wrapper around any sampler for DPP case: `DistributedSamplerWrapper`
> 
> Suppose we have a sampler (any subclass of a `Sampler`), then for DDP case we would do: `sampler = DistributedSamplerWrapper(sampler)` and that would make `sampler` compatible with DDP.
> 
> The suggested design of wrapper was proposed on pytorch forums, been tested for years and myself and does not require any special knowledge of underlying sampler implementation.
> 
> ```
> class DatasetFromSampler(Dataset):
>     """Dataset to create indexes from `Sampler`.
> 
>     Args:
>         sampler: PyTorch sampler
>     """
> 
>     def __init__(self, sampler: Sampler):
>         """Initialisation for DatasetFromSampler."""
>         self.sampler = sampler
>         self.sampler_list = None
> 
>     def __getitem__(self, index: int):
>         """Gets element of the dataset.
> 
>         Args:
>             index: index of the element in the dataset
> 
>         Returns:
>             Single element by index
>         """
>         if self.sampler_list is None:
>             self.sampler_list = list(self.sampler)
>         return self.sampler_list[index]
> 
>     def __len__(self) -> int:
>         """
>         Returns:
>             int: length of the dataset
>         """
>         return len(self.sampler)
> 
> 
> class DistributedSamplerWrapper(DistributedSampler):
>     """
>     Wrapper over `Sampler` for distributed training.
>     Allows you to use any sampler in distributed mode.
> 
>     It is especially useful in conjunction with
>     `torch.nn.parallel.DistributedDataParallel`. In such case, each
>     process can pass a DistributedSamplerWrapper instance as a DataLoader
>     sampler, and load a subset of subsampled data of the original dataset
>     that is exclusive to it.
> 
>     .. note::
>         Sampler is assumed to be of constant size.
>     """
> 
>     def __init__(
>         self,
>         sampler,
>         num_replicas: Optional[int] = None,
>         rank: Optional[int] = None,
>         shuffle: bool = True,
>     ):
>         """
> 
>         Args:
>             sampler: Sampler used for subsampling
>             num_replicas (int, optional): Number of processes participating in
>               distributed training
>             rank (int, optional): Rank of the current process
>               within ``num_replicas``
>             shuffle (bool, optional): If true (default),
>               sampler will shuffle the indices
>         """
>         super(DistributedSamplerWrapper, self).__init__(
>             DatasetFromSampler(sampler),
>             num_replicas=num_replicas,
>             rank=rank,
>             shuffle=shuffle,
>         )
>         self.sampler = sampler
> 
>     def __iter__(self):
> 
>         self.dataset = DatasetFromSampler(self.sampler)
>         indexes_of_indexes = super().__iter__()
>         subsampler_indexes = self.dataset
>         return iter(itemgetter(*indexes_of_indexes)(subsampler_indexes))
> ```

I like this ^^

NatanBagrov:
> Secondly, I think we should introduce some sort of the interface:
> 
> ```
> class HasSamplingInformation(ABC):
>    def getLabelsPresense() -> np.ndarray:
>       """
>       :returns: A Numpy array of [Dataset Length, Num Classes] with values corresponding to
>                 a number of objects at the current sample index.
>       """
> ```
> 
> Why we want to have interface?
> 
> * We can check whether dataset we are trying to use with sampler implements it (If not - we show nice error).
> * Sampler implementation gets a presence matrix which contains ALL necessary information to do sampling. No remapping indexes and other stuff.

This is a nice idea, I wonder if an array is too much compared to a generator. A fixed array, when we take Objects365 a an example is of size: `1.7M dataset size * 365 classes * 4 bytes ~= 2.5GB`
Another possible downside is cases where SG users have their dataset, and will not get this feature. On the other hand, perhaps it is fair to ask to implement something to get value.

NatanBagrov:
Closing because this PR has turned into two: #1865 #1856

PR #1849, opened 2024-02-20
Feature/sg 000 fix yolo nas scores

BloodAxe:
Fixed bug introduced in #1791 that did not scale bbox coordinates when `input_dim` argument for dataset was set.


shaydeci (approved):
LGTM

PR #1856, opened 2024-02-22
Added DistributedSamplerWrapper to automatically wrap non-dist samplers in cases we use dist mode

NatanBagrov:
This feature allows to wrap a non-distributed sampler and use it in a distributed setting. The wrapping happens automatically under-the-hood in case use is running DDP-mode and passed a non-distributed sampler.


BloodAxe (approved):
LGTM

PR #1861, opened 2024-02-25
fix vulnerabilities

ofrimasad:
removed the aws secrets manager (no longer used and contains logging issues)
upgraded onnx to 1.16 to avoid:
Versions of the package onnx before and including 1.15.0 are vulnerable to Directory Traversal as the external_data field of the tensor proto can have a path to the file which is outside the model current directory or user-provided directory. The vulnerability occurs as a bypass for the patch added for https://github.com/advisories/GHSA-ffxj-547x-5j7c.


shaydeci (approved):
LGTM

PR #1862, opened 2024-02-25
added registries

shaydeci:
Added registreis for the coco and yolo format detection classes.
Once registered users can simply plug them into their yaml files if they have the right file structure without the need to specify an entire datalloader (like we do in our main config i.e stating coco2017_train...):
```yaml
train_dataloader_params:
  dataset: COCOFormatDetectionDataset
  batch_size: 25
  num_workers: 8
  shuffle: True
  drop_last: True
  pin_memory: True
  collate_fn: DetectionCollateFN
```


ofrimasad (approved):
LGTM

BloodAxe (approved):
LGTM

PR #1865, opened 2024-02-25
Added class balancing sampler

NatanBagrov:
**Context:** in cases where some classes are less frequent than others, we'd want to sample images containing these classes more often.

Current implementation:
* `HasClassesInformation` Protocol - Datasets can choose to support this protocol. Supporting it will allow the class balancing our-of-the-box.
* `ClassBalancingSampler` - A sampler that oversamples scarce classes (based on `RandomWeightedSampler`). The default sampling strategy follows https://arxiv.org/pdf/1908.03195.pdf. There are two degrees of freedom which are the frequency threshold for the oversampling, and the oversampling aggressiveness.

Experiments and limitations (**over COCO-val set, to speed-up the process**):
The default values are `oversample_threshold=None, oversample_agressiveness=0.5`, where `oversample_threshold=None` means classes that are more frequent than the *median* frequency will not we oversampled. The value of `oversample_agressiveness=0.5` means that the repeat factor for the oversampled classes is `repeat_factor = (oversample_threshold / class_frequency) ** 0.5`

Setting a value of `oversample_agressiveness=0.0` means no oversampling at all. Setting a value of `oversample_threshold=1.0` means all classes will be oversampled (according to their scarcity)

```
BEFORE BALANCE: Most frequent class (#0) appears 10777 times, which is 1197.44x more frequent than the least frequent class (#70) that appears only 9 times!

AFTER BALANCE (oversample_threshold=None, oversample_agressiveness=0.5): Most frequent class (#0) appears 10351 times, which is 265.41x more frequent than the least frequent class (#70) that appears only 39 times!

AFTER BALANCE (oversample_threshold=None, oversample_agressiveness=1.0): Most frequent class (#0) appears 9133 times, which is 71.35x more frequent than the least frequent class (#31) that appears only 128 times!

AFTER BALANCE (oversample_threshold=None, oversample_agressiveness=1.5): Most frequent class (#0) appears 8147 times, which is 103.13x more frequent than the least frequent class (#37) that appears only 79 times!

AFTER BALANCE (oversample_threshold=1.0, oversample_agressiveness=0.5): Most frequent class (#0) appears 10605 times, which is 303.00x more frequent than the least frequent class (#70) that appears only 35 times!

AFTER BALANCE (oversample_threshold=1.0, oversample_agressiveness=1.0): Most frequent class (#0) appears 9887 times, which is 73.24x more frequent than the least frequent class (#22) that appears only 135 times!

AFTER BALANCE (oversample_threshold=1.0, oversample_agressiveness=1.5): Most frequent class (#0) appears 7835 times, which is 89.03x more frequent than the least frequent class (#19) that appears only 88 times!
```

Usage:
```yaml
train_dataloader_params:
    dataset: my_dataset
    sampler:
       ClassBalancingSampler:
           oversample_threshold:  # default
           oversample_aggressiveness: 0.5
    batch_size: 16
    ...
```

**Discussion**
* The current implementation does not change the size of the *epoch*. **It is not-deterministic**. One can implement a sampler that would *repeat* the scarce images, changing the epoch's size.


src/super_gradients/training/datasets/samplers/class_balanced_sampler.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        Wrap WeightedRandomSampler with weights that are computed from the class frequencies of the dataset.
+        """
+
+        if dataset is None and precomputed_factors_file is None:
+            raise ValueError("`dataset` and `precomputed_factors` cannot be both None.")
+
+        if dataset is not None and precomputed_factors_file is not None:
+            # this logic is to simplify `_instantiate_sampler` method.
+            warnings.warn("Both `dataset` and `precomputed_factors_file` are passed. `dataset` WILL BE IGNORED!")
```
NatanBagrov, line 139: Because dataset is injected by `_instantiate_sampler`, there's a problem when `precomputed_factors_file` is passed, because then both appear.
We can either add logic to `_instantiate_sampler` (I don't think it should know the internals), or to prioritize `precomputed_factors_file` like implemented here.

src/super_gradients/training/datasets/samplers/class_balanced_sampler.py
```diff
@@ -0,0 +1,151 @@
...
+
+        The repeat factor is computed as followed:
+        1. For each class c, compute the fraction # of images that contain it (its frequency): :math:`f(c)`
+        2. For each class c, compute the category-level repeat factor: :math:`r(c) = max(1, aggressiveness(threshold/f(c)))`
+        3. For each image I, compute the image-level repeat factor: :math:`r(I) = max_{c in I} r(c)`
+
+        Returns a list of repeat factors (length = dataset_length). How to read: result[i] is a float, indicates the repeat factor of image i.
+        """
+
+        class_information = class_information_provider.get_dataset_classes_information()
```
BloodAxe, line 66: A shape hint of `class_information` would help. E.g is it `[Num Samples, Num Classes] or vice versa

src/super_gradients/training/datasets/samplers/class_balanced_sampler.py
```diff
@@ -0,0 +1,151 @@
...
+            cat_ids = categories[sample_cat_freq != 0]
+            if len(cat_ids) == 0:  # in case image doesn't have annotations, we will not over-sample nor ignore it
+                repeat_factors.append(1.0)
+            else:
+                repeat_factors.append(max({category_repeat[cat_id] for cat_id in cat_ids}))
+
+        return repeat_factors  # len = dataset_length
+
+    @staticmethod
+    def precompute_sample_repeat_factors(
```
BloodAxe, line 91: What is the point of this method? I don't quite get the motivation to save precompute factors in some file.

src/super_gradients/examples/class_balancing/coco_class_balancing_example.py
```diff
@@ -0,0 +1,71 @@
...
+from torch.utils.data import DataLoader
+from tqdm import tqdm
+
+from super_gradients.training.datasets.detection_datasets.coco_format_detection import COCOFormatDetectionDataset
+from super_gradients.training.datasets.samplers.class_balanced_sampler import ClassBalancedSampler
+from super_gradients.training.transforms import DetectionPaddedRescale
+from super_gradients.training.utils.collate_fn import DetectionCollateFN
+
+
+def get_class_frequency(dataloader):
```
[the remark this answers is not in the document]
NatanBagrov: Done, [db2f477](https://github.com/Deci-AI/super-gradients/pull/1865/commits/db2f477deb3b310c30c275e56eeaae7ef71979f2)

NatanBagrov:
Addressed comments.

src/super_gradients/training/datasets/samplers/class_balanced_sampler.py
NatanBagrov: Done, [db2f477](https://github.com/Deci-AI/super-gradients/pull/1865/commits/db2f477deb3b310c30c275e56eeaae7ef71979f2)

src/super_gradients/examples/class_balancing/coco_class_balancing_example.py
```diff
@@ -35,15 +35,15 @@ def main():
     initial_discrepancy = initial_class_balance.max() / initial_class_balance.min()
 
     print(
```
BloodAxe, line 37: Please use `logger.info` or `logger.debug` instead of regular prints 🙏

BloodAxe (approved):
LGTM

PR #1874, opened 2024-02-28
fix bug

ofrimasad:


src/super_gradients/training/transforms/detection/legacy_detection_transform_mixin.py
```diff
@@ -26,7 +26,7 @@ def __call__(self, sample: Union["DetectionSample", Dict[str, Any]]) -> Union["D
         else:
             sample = self.convert_input_dict_to_detection_sample(sample)
             sample = self.apply_to_sample(sample)
-            return self.convert_detection_sample_to_dict(sample, include_crowd_target="crowd_targets" in sample)
+            return self.convert_detection_sample_to_dict(sample, include_crowd_target=hasattr(sample, "crowd_targets"))
```
BloodAxe, line 29: Aaaactulyyy (c)

I believe this code should look like this:
```
has_crowd_target = "crowd_targets" in sample
sample = self.convert_input_dict_to_detection_sample(sample)
sample = self.apply_to_sample(sample)
return self.convert_detection_sample_to_dict(sample, include_crowd_target=has_crowd_target)
```

So when we convert from dict to a Detection sample we keep track whether or not `crowd_target` key was present in the input sample. And so even if crowd targets gets removed by some transforms, we would retain `crowd_target` key as it may be expected to be present by a collate function.

BloodAxe: Also, it is `crowd_target` not `crowd_targets`.

ofrimasad:
addressed

BloodAxe (approved):
LGTM

PR #1875, opened 2024-02-29
Bugfix/sg 000 fix test for deprecated

ofrimasad:
remove test and name for FunctionLRScheduler (was removed on December)
fix crash in YoloNAS_Pose_Fine_Tuning_Animals_Pose_Dataset.ipynb (missing import)

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -5,3 +5,4 @@ gitpython>=3.1.0
 ipykernel==6.25
 nbconvert==7.8.0
 pycocotools==2.0.6
+google-colab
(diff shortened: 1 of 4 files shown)


requirements.dev.txt
```diff
@@ -5,3 +5,4 @@ gitpython>=3.1.0
 ipykernel==6.25
 nbconvert==7.8.0
 pycocotools==2.0.6
+google-colab
```
BloodAxe, line 8: I looked where it is used and it seems we can simply remove the import `from google.colab.patches import cv2_imshow` from the notebook `YoloNAS_Pose_Fine_Tuning_Animals_Pose_Dataset.ipynb` as it is not used at all.
And same is true for `from IPython.display import YouTubeVideo` import from the sample notebook.

Issue #1834, opened 2024-02-13
Perform inference on an opencv image loaded from memory


M3nxudo:
### 💡 Your Question

Is it possible to perform inference on an image already in memory (through Opencv videocapture)?
I have tried to do so but get no detections when trying to perform inference in the following way:
1. Defining a class that has the model info and a method for performing the detection and returns the predictions
```
class objectNAS():
    def __init__(self, confThreshold):
        self.conf = confThreshold
        model_name = "yolo_nas_l"
        self.model =  super_gradients.training.models.get(model_name, pretrained_weights="coco").cuda()

    def detect(self,input_image): 
        model_prediction = self.model.predict(input_image, iou=0.5, conf=self.conf)
        prediction = model_prediction.prediction
        print ('Debugging prediction')
        print (prediction.bboxes_xyxy[0])
        return prediction
```
2. I create an object of the aforementioned class and then try to perform inference with it inside of an OpenCV Videocapture loop

```
self.capVideo = cv2.VideoCapture(self.videofile[0])
        count = 0
        while self.capVideo.isOpened():
            ret, image = self.capVideo.read()
            if (ret):
                image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
                height, width, channel = image.shape
                step = channel * width
                result = objectNAS.detect(image)
```
In doing so and debugging the code i get an error in the line:
print (prediction.bboxes_xyxy[0]) with the following text:
  File "D:\source\repos\libMuinen_UI\nas_interface.py", line 26, in detect
    print (prediction.bboxes_xyxy[0])
IndexError: index 0 is out of bounds for axis 0 with size 0

Which if I understand correctly means that the prediction results are empty and can't access them but from the input image I know i should be getting at least some detections.
Would love to know if i'm using incorrectly the predict method or any tips to make my code work.

BloodAxe:
The `model_prediction = self.model.predict(input_image, iou=0.5, conf=self.conf)` line looks legit.
You can pass your as RGB (not BGR as OpenCV reads) image as numpy array to predict().
It could be you are simply getting no detections (Maybe your confThreshold is too high?).
So if there are no boxes you would get an index error at `bboxes_xyxy[0]` which is expected.

A somewhat related issue where you can find a code snippet of printing all detections:
https://github.com/Deci-AI/super-gradients/issues/1818

M3nxudo:
[...]

PR #1881, opened 2024-03-01
Introduce fp16 flag to enable/disable mixed precision for predict()

BloodAxe:
This PR adds `fp16` flag to `model.predict()` which can be handy to force fp32 computation in predict()

As reported in https://github.com/Deci-AI/super-gradients/issues/1834 there are some GPUs [known](https://github.com/pytorch/pytorch/issues/58123) of not doing correct math in fp16 mode. 
So with `model.predict(..., fp16=False)` users would have the ability to go for fp32 mode as a workaround for that specific GPU.


ofrimasad (approved):
LGTM

PR #1884, opened 2024-03-04
fix a bug when ploting a dataset with images in a range other than 0-255

ofrimasad:


src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -493,6 +505,7 @@ def plot(
                 if image.shape[0] in (1, 3):  # (C, H, W) -> (H, W, C)
                     image = image.transpose((1, 2, 0))
 
+                image = self._standardize_image()
```
BloodAxe, line 508: Function call missing an argument `image`

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
```diff
@@ -449,6 +449,18 @@ def output_target_format(self):
                 target_format = transform.output_format
         return target_format
 
+    @staticmethod
+    def _standardize_image(image):
+        # Normalize the image to have minimum of 0 and maximum of 1
+        image_min = image.min()
+        image_max = image.max()
+        normalized_image = (image - image_min) / (image_max - image_min)
```
BloodAxe, line 457: (max - min + eps) to avoid division by zero if image is happen to be a be flat

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
ofrimasad: 🤦‍♂️

src/super_gradients/training/datasets/detection_datasets/detection_dataset.py
ofrimasad: done

BloodAxe (approved):
LGTm

PR #1895, opened 2024-03-08
Explicitly specify a kernel name that match the environment in CI

BloodAxe:
This is to have correct "pip install" in the notebook


ofrimasad (approved):
LGTM

PR #1904, opened 2024-03-11
Added `Trainer.evaluate_from_config` to align with `Trainer.train_from_config`

NatanBagrov:
Added `Trainer.evaluate_from_config` to align with `Trainer.train_from_config`


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -441,7 +446,7 @@ def evaluate_checkpoint(
         cfg = load_experiment_cfg(ckpt_root_dir=ckpt_root_dir, experiment_name=experiment_name, run_id=run_id)
 
         add_params_to_cfg(cfg, params=["training_hyperparams.resume=True", f"ckpt_name={ckpt_name}"])
-        cls.evaluate_from_recipe(cfg)
+        cls.evaluate_from_config(cfg)
```
BloodAxe, line 449: Add `return` to not lost the evaluation result pls.

NatanBagrov: Added, [0cfa329](https://github.com/Deci-AI/super-gradients/pull/1904/commits/0cfa329dba084337aa8fa1a15a0a9a3c6aa64bb4)

shaydeci (changes requested):
@NatanBagrov What is the context of this PR ?
Its good that we are removing the script from the examples directory, but we did it so that we can use the module way of running it.
So why are we brining it back under a different name?

If you are suggesting to rename the module itself, then that is a different topic.

NatanBagrov:
> @NatanBagrov What is the context of this PR ? Its good that we are removing the script from the examples directory, but we did it so that we can use the module way of running it. So why are we brining it back under a different name?
> 
> If you are suggesting to rename the module itself, then that is a different topic.

We have `train_from_config` and `evaluate_from_recipe`. The goal is to align on the terminology.

shaydeci:
> > @NatanBagrov What is the context of this PR ? Its good that we are removing the script from the examples directory, but we did it so that we can use the module way of running it. So why are we brining it back under a different name?
> > If you are suggesting to rename the module itself, then that is a different topic.
> 
> We have `train_from_config` and `evaluate_from_recipe`. The goal is to align on the terminology.

Sure, no problem with that. But:
1. You are also changing the module/script name. Thats actually not alligned with the other way of launching recipes in sg (i.e python -m super_gradients.train_from_recipe)
2. Please properly deprecate the method (see super_gradients.common.deprecate.deprecated and how it is used in SG - specify version deprecation, version removal etc)

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -407,14 +407,19 @@ def evaluate_from_recipe(cls, cfg: DictConfig) -> Tuple[nn.Module, Tuple]:
 
         return model, valid_metrics_dict
 
+    @classmethod
+    def evaluate_from_recipe(cls, cfg: DictConfig) -> Tuple[nn.Module, Tuple]:
```
shaydeci, line 411: This is not how we deprecate methods in SG.
See super_gradients.common.deprecate.deprecated.

NatanBagrov: Changed,  [ccd7302](https://github.com/Deci-AI/super-gradients/pull/1904/commits/ccd73021509e912577aa3f094aac2754b190db7b)

shaydeci (approved):
LGTM

PR #1908, opened 2024-03-12
Install pytorch-quantization on demand

BloodAxe:
This PR adds adds automatic attempt to import pytorch-quantization when in use (A call to ptq/qat, quantize_from_recipe or call to model.export with quantization requested).

## What is changing

1) pytorch-quantization will be installed if it's not yet installed, during call to ptq/qat, quantize_from_recipe or call to model.export. 
2) importing `super_gradients.training.utils.quantization` outside of the above mentioned functions would raise an ImportError if package is not installed. This is by design since we don't have reasons to expose internal details of quantization outside.

## How

For this reason, imports are reorganized and all quantization-related modules are moved to `super_gradients.training.utils.quantization`. The sub-package itself will raise an error if pytorch quantization is not installed (As it is not meant to be used directly by a user).
However, call to ptq/qat/export will explicitly call `import_pytorch_quantization_or_install` to ensure dependency is installed. This method will also ensure we do monkey-patching of QuantConvTranspose2d to work with newer torch versions.

A CI config is also updated to remove installations pytorch-quantization ahead of time to ensure all works as it should


src/super_gradients/import_utils/pytorch_quantization.py
```diff
@@ -0,0 +1,74 @@
...
+    package = "pytorch_quantization"
+
+    try:
+        importlib.import_module(package)
+
+        patch_pytorch_quantization_modules_if_needed()
+    except ImportError:
+        import pip
+
+        pip.main(["install", "pytorch_quantization==2.1.2", "--extra-index-url", "https://pypi.ngc.nvidia.com"])
```
shaydeci, line 72: May I suggest:

```
import subprocess
import sys

def install_package(package):
    subprocess.check_call([sys.executable, "-m", "pip", "install", package, "--extra-index-url", "https://pypi.ngc.nvidia.com"])

install_package("pytorch_quantization==2.1.2")
```

because of:
https://stackoverflow.com/questions/49870594/pip-main-install-fails-with-module-object-has-no-attribute-main

BloodAxe: Thanks, fixed in 77c8d16520c64f834ec0dfffe04082bff94fd241

shaydeci (dismissed):
LGTM

shaydeci (approved):

PR #1910, opened 2024-03-12
Added support for VOC format detection dataset class

shaydeci:
- Added Pascal VOC format detection dataset class.
- Refactored the VOC dataset class to inherit from it.
- Deprecated the images_sub_directory parameter so that we are alligned with the other COCO/YOLO format classes and also allow more flexibility in the file structure.
- Fixed some failing tests.


src/super_gradients/training/datasets/detection_datasets/pascal_voc_format_detection.py
```diff
@@ -0,0 +1,141 @@
...
+        deprecated_since="3.7.0",
+        removed_from="3.8.0",
+        reason="Support of `images_sub_directory` is removed since it allows less flexibility." " Please use 'images_dir' and 'labels_dir' instead.",
+    )
+    def __init__(
+        self,
+        images_dir: Optional[str],
+        labels_dir: Optional[str],
+        *args,
+        **kwargs,
```
BloodAxe, line 84: A `*args, **kwargs` is a chaotic evil move. Can we have explicit and full list of arguments we can use in this class's c'tor?

shaydeci: done

src/super_gradients/training/datasets/detection_datasets/pascal_voc_format_detection.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def _setup_data_source(self) -> int:
+        """Initialize img_and_target_path_list and warn if label file is missing
+
+        :return: List of tuples made of (img_path,target_path)
+        """
+        if not Path(self.images_dir).exists():
+            raise FileNotFoundError(f"{self.images_dir} not found.")
+
+        img_files = glob.glob(os.path.join(self.images_dir, "*.jpg"))
```
BloodAxe, line 165: Pls add list(sorted(...)) 
Why it is important:
glob return images in no particular order:
> Return a list containing the names of the entries in the directory given by path. The list is in arbitrary order.

So it may happen that in DDP you would get different orders, and thus some images would be excluded from train/validation and some appear more than onces

shaydeci: Nice catch!
done in latest commit.

PR #1918, opened 2024-03-14
Fix the logic how number of calibration batches are computed in PTQ to match the documentation

BloodAxe:
It seems, having a too many calibration batches actually may ruin the model. So put a 16 batches as a default number of calibration batches. Investigating why it is so - is on TODO.

Also, fixed a bug when `.quantize_module` was called a second time

And re-run PTQ/QAT notebook just in case


shaydeci (approved):
LGTM

src/super_gradients/qat_from_recipe.py
```diff
@@ -13,7 +15,16 @@
 
 @hydra.main(config_path="recipes", version_base="1.2")
 def _main(cfg: DictConfig) -> None:
-    Trainer.quantize_from_config(cfg)
+    result = Trainer.quantize_from_config(cfg)
+
+    print("Validation result of quantized model:")
```
shaydeci, line 20: Leftover prints ?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2714,14 +2714,13 @@ def ptq(
             verbose=get_param(calib_params, "verbose"),
         )
         q_util.register_skip_quantization(layer_names=get_param(selective_quantizer_params, "skip_modules"))
-        q_util.quantize_module(model)
 
         model = ptq(
             model,
             selective_quantizer=q_util,
-            calib_loader=calib_loader,
+            calibration_loader=calib_loader,
```
shaydeci, line 2721: Did we rename this parameter and properly deprecated it ?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -2714,14 +2714,13 @@ def ptq(
...
             calibration_method=get_param(calib_params, "histogram_calib_method"),
-            calibration_batches=get_param(calib_params, "num_calib_batches") or len(calib_loader),
+            calibration_batches=get_param(calib_params, "num_calib_batches") or max(1, int(512 // calib_loader.batch_size)),
```
shaydeci, line 2723: Perhpas its a good idea to actually have 16 is the default here as well? WDYT ?

shaydeci (changes requested):

src/super_gradients/training/sg_trainer/sg_trainer.py
NatanBagrov: Basically there's some sort of a mix between samples and batches.
Eventually, you want to see enough samples (a reasonable number can be 512), so you calculate the batches you need to see this number of samples.
For 16 batches it becomes dependent on the batch size and not sample side.
It would be better to standardize it (in terms of: passing num samples and deducing batches, and not a mix).

src/super_gradients/qat_from_recipe.py
BloodAxe: Actually I left then intentionally to print the results of quantize_from_config to console.

src/super_gradients/training/sg_trainer/sg_trainer.py
BloodAxe: It isn't rename.
The former PR was merged with an error

shaydeci (approved):
LGTM

PR #1923, opened 2024-03-15
Pyparsing version more flexible

AlanBlanchet:
* Conflict when installing dvc. It requires a more recent version

Hey. Is it possible to make this more flexible ?
Some other versions for other packages seem to be fixed in your requirements.txt.
It makes it hard to install other packages...
If 2 other entities create a requirements.txt and freezes the version of a package in the requirements.txt. If they are different the install will crash...
For instance there might be problems with the torchmetrics, onnx... packages since their version are fixed. If someone else also freezes those versions, then no one would be able to install the whole package.

Thanks

### DIFF
--- a/src/super_gradients/recipes/dataset_params/pascal_voc_detection_dataset_params.yaml
+++ b/src/super_gradients/recipes/dataset_params/pascal_voc_detection_dataset_params.yaml
@@ -7,6 +7,8 @@ train_dataset_params:
     - DetectionTargetsFormatTransform:
         input_dim: ${dataset_params.train_dataset_params.input_dim}
         output_format: LABEL_CXCYWH
+  images_dir: images
+  labels_dir: labels
   class_inclusion_list:
   max_num_samples:
   download: True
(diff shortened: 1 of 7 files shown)


BloodAxe:
I think we can safely remove `pyparsing` dependency completely as it seems to be a leftover from older SG versions. Please follow https://github.com/Deci-AI/super-gradients/blob/master/CONTRIBUTING.md to set up signed commits if you are interested in contributing this improvement. Thanks!

AlanBlanchet:
Is there anything else to completely remove from the requiements ? @BloodAxe 
I'll format my PR with your contribution guidelines.

BloodAxe:
I don't have a list of outdated requirements from top of my head. Removing even one not used requirement is already good.
But if you willing to go above and beyond - feel free to use some package scanning tools to check for unused dependencies

AlanBlanchet:
I'll make another PR following the guildelines and removing pyparsing then. Thanks

PR #1932, opened 2024-03-20
Added validation step before resuming when failing to fetch the metric to watch

shaydeci:
Also fixes earlier bug that always loaded the Trainer.best_metric from the "acc" entry that actually holds latest metric_to_watch rather then best.


src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1877,7 +1912,12 @@ def _load_checkpoint_to_model(self):
                 )
 
         # UPDATE TRAINING PARAMS IF THEY EXIST & WE ARE NOT LOADING AN EXTERNAL MODEL's WEIGHTS
-        self.best_metric = self.checkpoint["acc"] if "acc" in self.checkpoint.keys() else -1
+        _best_ckpt_metrics = get_param(self.checkpoint, "_best_ckpt_metrics")
+        if self.load_checkpoint or resume_path:
```
BloodAxe, line 1916: Are `load_checkpoint` or `resume_path` synonims? Why are we using the same flow for both these cases? IMO - when resuming we load everything and resuming experiment (in same folder).
When `load_checkpoint` - we simply load weights to a model, but starting from scratch and not looking at the old weights. Am I making a right assumption here?

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1653,6 +1660,34 @@ def get_finetune_lr_dict(self, lr: float) -> Dict[str, float]:
...
+            self.net.eval()
+            self._reset_metrics()
+            self.valid_metrics.to(device_config.device)
+            valid_metrics_dict = self.evaluate(
+                data_loader=self.valid_loader,
+                metrics=self.valid_metrics,
+                evaluation_type=EvaluationType.TEST,
+                silent_mode=silent_mode,
+            )
+            self._save_checkpoint(
```
BloodAxe, line 1680: Why _save_checkpoint here? We should just obtain the metrics. If the goal is to overwrite the previous best checkpoint with a new metrics - I think this is very unexpected behavior and I personally against such move.

src/super_gradients/training/sg_trainer/sg_trainer.py
```diff
@@ -1877,7 +1912,12 @@ def _load_checkpoint_to_model(self):
                 )
 
         # UPDATE TRAINING PARAMS IF THEY EXIST & WE ARE NOT LOADING AN EXTERNAL MODEL's WEIGHTS
-        self.best_metric = self.checkpoint["acc"] if "acc" in self.checkpoint.keys() else -1
+        _best_ckpt_metrics = get_param(self.checkpoint, "_best_ckpt_metrics")
+        if self.load_checkpoint or resume_path:
+            if _best_ckpt_metrics is None or self.metric_to_watch != _best_ckpt_metrics["tracked_metric_name"]:
+                self._eval_before_resume = True
```
BloodAxe, line 1918: Is it possible to call `eval` right here and not splitting logic into two parts (detecting whether evaluation is needed and actual evaluation)?

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: The `if self.load_checkpoint or resume_path:` can indeed be simplified to `if self.load_checkpoint`.
Loading weights only is not handled since long time in the Trainer itself, but through models.get(), which is used by Trainer and one can pass certain weights through checkpoint_params. 
Long story short- the load_checkpoint attribute is used to load weights for resuming training.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: It is necessary to save the checkpoint with these metrics, as continuing the training might have inferior results, for example:
- epoch 0, completed with XXX accuracy score.
Resuming training, and we change metric_to_watch to top5
- We evaluate top5, we get 0.3, **dont save checkpoint**  like you suggested.
- Then in the end of epoch 1, we get 0.2, then it would be considered best and we would save the non optimal checkpoint as best.

Why is it unexpected behaviour?
The checkpoint_best.pth should always hold the best checkpoint according to the metric_to_watch at that time.

src/super_gradients/training/sg_trainer/sg_trainer.py
shaydeci: Problem is there's alot more that needs to be done before we can call `eval` (instantiating metric objects, loss etc).
I rather not change the order of things in Trainer for this cause.
I can move this check right before calling eval (see why we need it prior to resuming in previous comment) if the flag is bothering.

PR #1936, opened 2024-03-22
Fixed a bug in visualization of the worst batch for OD task

BloodAxe:
There was a big in slicing code that used a wrong offset in `targets` which lead to rendering of GT boxes with wrong labels and wrong coordinates.


ofrimasad (approved):
LGTM

PR #1942, opened 2024-03-31
unwrap when deprecating decorated methods

NatanBagrov:
When deprecating a class method (or a static method), there's no `__module__` in `target`, since it is decorated (e.g., with `@classmethod`). This PR unwraps it.


shaydeci (approved):
LGTM

PR #1946, opened 2024-04-02
Optimize SG dependencies

BloodAxe:
1) Sphinx & Coverage dependencies moved to requirements.dev.txt
2) Removed protobuf
3) Removed numpy dependency as it will come from pytorch automatically and at this point <1.23 should not be relevant as we got rid of all numpy deprecation warnings
4) Removed loose omegaconf requirement since it will come from hydra-core anyway
5) Lift up onnx & onnxruntime versions (Can't remove since we have ModelConversionCallback that uses it)


BloodAxe:
A setuptools 67 was released in January 2023 which is eternity ago .
Can't we somehow exclude only affected version and not the whole two ears of releases?

ofrimasad (approved):
LGTM

ofrimasad (approved):

PR #1951, opened 2024-04-08
Abstract optical flow dataset class and KITTI dataset class

Yael-Baron:


shaydeci (changes requested):
There are quite alot of irrelated changes which stemm from the fact that the optical_flow_master isnt updated with master.

Yael-Baron:
See PR Feature/dle 599 optical flow integration #1984

PR #1959, opened 2024-04-11
Added imagesize dependency

BloodAxe:
After merging https://github.com/Deci-AI/super-gradients/pull/1946 an issue popped up with missing `imagesize` dependency in Yolo format dataset.
Probably it came as a bonus from `sphinx`. Since `sphinx` is not with us anymore `imagesize` was not installed and caused test to fail. Added latest `imagesize` dependency


ofrimasad (approved):
LGTM

PR #1962, opened 2024-04-15
OpenVINO-friendly heads  for YoloNAS & YoloNAS-Pose

BloodAxe:
It seems existing call to `F.conv2d` to compute weighted sum of box size distributions predictions confuses OpenVINO quantizers. This PR replaces this operation with simple multiplication which is equivalent op and works well for both TRT & Vino.


shaydeci (approved):
LGTM

BloodAxe:
The speed difference is not statistically significant
image

PR #1976, opened 2024-05-01
Fix average best models to false when save model is false

shaydeci:
If average_best_models=True which is the default, and save_model=False - the last validation will crash since no snapshots are saved throughout training.


BloodAxe (approved):
LGTM

PR #1979, opened 2024-05-05
Feature/sg 1442 sliding window inference for yolonas

shaydeci:
I am opening this WIP PR as a draft, to discuss some small issues.
Implementing sliding window inference inside CustomizableDetector, it happens that a few arch params are now intersecting with some of the predict() ones.
As @BloodAxe suggested, I run NMS on each tile and finally run NMS (just the IOU part) on the aggregated predictions.
But for predict() we can also pass NMS parameters.
As a temporary solution I just added an IdentityPostPredictionCallback that in the case of sliding window would be used, and internally the user's specified NMS params that were passed in __ init __ would be used for the tiles.
What would be your take on how to handle both predict() specified NMS params and the ones passed in init for the sliding window ?
Is having a different IOU threshold for the final NMS also desired?
@BloodAxe @ofrimasad


src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -89,7 +107,71 @@ def __init__(
         self._default_multi_label_per_box = True
         self._default_class_agnostic_nms = False
 
+    def forward_sliding_window(self, images):
```
BloodAxe, line 110: IDK, I think this overcomplicate things.
What I was thinking about in my thoughts was an additional parameter to `predict` method that would enable tiled inference:
`model.predict(..., tiled=True)` while `tile_size` would be the default image size (we have this info already and `tile_step` would be equal to `tile_size` by default. Sure, user can override these params if he wants to. `tile=True` would be mutually exclusive with `skip_image_resizing` parameter for obvious reasons.

Moving this to `predict` would allow us to offer this feature to other OD models. WDYT?

src/super_gradients/training/models/detection_models/customizable_detector.py
```diff
@@ -89,7 +107,71 @@ def __init__(
...
+            else:
+                final_detections.append(torch.empty(0, 6).to(images.device))  # Empty tensor for images with no detections
+        return final_detections
+
+    def _generate_tiles(self, image, tile_size, tile_step):
+        _, _, h, w = image.shape
+        tiles = []
+
+        # Calculate the end points for the grid
+        max_y = h if (h - tile_size) % tile_step < self.min_tile_threshold else h - (h - tile_size) % tile_step + tile_size
```
BloodAxe, line 151: Ok, I'm already lot in this logic. I probably can grasp what is going on here if I take a sheet of paper and a pen. But is there a specific reason we want to do it this way and not like so:

```
def _generate_tiles(self, image, tile_size, tile_step):
  image_height, image_width = image.shape[:2]
  start_y = 0
  while start_y > image_height:
     start_x = 0
     while start_x > image_width:
       crop = image[start_y: start_y+tile_size, start_x:start_x+tile_size]
       yield crop, (start_y, start_x)
       start_x += tile_step
  start_y += tile_step

```

shaydeci: I think you wouldv'e got pretty easily if I added docs (: 
@ofrimasad suggested we add a min_tile_threshold parameter, so that for example in the case where:
image size = 640,641
tile_size = 640
tile_step = 640
It would completely irrleveant to pass a 1X640 padded to 640 in the second tile, so just skip it.

So the first 2 lines just say:
"calculate the end points of the tile for padding, if the remainder when dividing by the tile size is smaller then min_tile_threshold -> leave that tile out"

src/super_gradients/training/models/detection_models/customizable_detector.py
shaydeci: Putting it all in predict would make everything much easier of course.
On the other hand, this might come in handy when people want to use sliding window for producing benchmark results.

src/super_gradients/training/metrics/detection_metrics.py
```diff
@@ -179,7 +181,8 @@ def update(self, preds, target: torch.Tensor, device: str, inputs: torch.tensor,
         targets = target.clone()
         crowd_targets = torch.zeros(size=(0, 6), device=device) if crowd_targets is None else crowd_targets.clone()
 
-        preds = self.post_prediction_callback(preds, device=device)
```
BloodAxe, line 182: Leftover?

src/super_gradients/training/models/detection_models/sliding_window_detection_forward_wrapper.py
```diff
@@ -0,0 +1,395 @@
...
+from super_gradients.training.utils.media.image import ImageSource
+import torchvision
+from super_gradients.training.utils.detection_utils import DetectionPostPredictionCallback
+
+
+class SlidingWindowInferenceDetectionWrapper(HasPredict, nn.Module):
+    """
+    Implements a sliding window inference wrapper for a customizable detector.
+
+    Parameters:
```
BloodAxe, line 22: A docstring is not following `:param param_name: ` pattern that we use elsewhere

src/super_gradients/training/metrics/detection_metrics.py
shaydeci: Nope, just added the option to pass None as a post_prediction_callback in the case we want to pass the predictions as they are (which is what we need when we use the slidding window since NMS already happens inside).
I am addressing this in the docs in this PR on this page.

src/super_gradients/training/models/detection_models/sliding_window_detection_forward_wrapper.py
```diff
@@ -0,0 +1,395 @@
...
+        final_detections = []
+        for detections in all_detections:
+            if detections:
+                detections = torch.cat(detections, dim=0)
+                # Apply global NMS
+                pred_bboxes = detections[:, :4]
+                pred_cls_conf = detections[:, 4]
+                pred_cls_label = detections[:, 5]
+                idx_to_keep = torchvision.ops.boxes.batched_nms(
+                    boxes=pred_bboxes, scores=pred_cls_conf, idxs=pred_cls_label, iou_threshold=sliding_window_post_prediction_callback.score_threshold
```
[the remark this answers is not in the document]
shaydeci: Indeed mixed them up accidentally.
Fixed.

BloodAxe (approved):
LGTM

PR #1982, opened 2024-05-07
more deprecated default args removed from LaunchConfig

shaydeci:
The defaults were passed through these args so previous versions of torch are still supported.
However, in torch 2.3.0 they are already removed.
Shuold fix the sanity tests (ran DDP training with 2.3.0 and all worked fine)


ofrimasad (approved):
LGTM

Issue #1986, opened 2024-05-09
Recipe should be saved to additional_configs_to_log before instantiating it


simgunz:
### 🐛 Describe the bug

The current `Trainer` code logs the **instantiated** cfg to `additional_configs_to_log` making it impossible to retrieve the original config to re-run the training. By moving up the line that generates `recipe_logged_cfg` the logged config will be exactly the original one and can be used to replicate the training.

In `Trainer.train_from_config` the line:
```
recipe_logged_cfg = {"recipe_config": OmegaConf.to_container(cfg, resolve=True)}
```
should be called before executing
```
cfg = hydra.utils.instantiate(cfg)
```
and probably also before
```
cfg = cls._trigger_cfg_modifying_callbacks(cfg)
```

An example of affected configuration key is `post_prediction_callback`.

### Versions

super_gradients-3.7.1

PR #1988, opened 2024-05-10
Fixed issue of logging hydra-instantiated config

BloodAxe:
This PR fixes an issue in logging of train config object.
For logging we want non-instantiated config to be logged, so the fix is to resolve config first and retain this copy for logging and then instantiate it.

Reference: https://github.com/Deci-AI/super-gradients/issues/1986


shaydeci (approved):
LGTM

PR #1990, opened 2024-05-10
2X training speed improvement: prefetch next batch from the data loader while running current batch

koush:
previous pull request seemed to be failing in circle ci due to a force push causing a missing ref.

### DIFF
--- a/src/super_gradients/training/sg_trainer/sg_trainer.py
+++ b/src/super_gradients/training/sg_trainer/sg_trainer.py
@@ -483,7 +483,31 @@ class Trainer:
 
             context.update_context(loss_avg_meter=loss_avg_meter, metrics_compute_fn=self.train_metrics)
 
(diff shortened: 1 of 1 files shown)


koush:
no idea why `ci/circleci: breaking-change-check` is failing

BloodAxe:
I have tested this PR on YoloNAS-Pose task and found no speed gains in DDP mode (4 GPUs) 
I may try on single gpu if I get some free time. But, by chance do you have any benchmark numers you did to prove this speed gain claim?

koush:
You're correct, on a hunch I dug through the recipe/script I was using from the docs or something, and found that num_workers was set to `0`. Removing that resulted in speeds equivalent to the changes I made. Further investigation and I found that PyTorch has a built in prefetch of batches in the data loader if the num_workers is non-zero.

https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader

PR #1994, opened 2024-05-16
Split processing and default preprocesisng params into separate files

BloodAxe:
This PR contains some groundwork for YoloNAS-R:

1) Moves all default preprocessing params to separate file in order to make `processing.py` smaller and also allow introduction of OBB Processings to be in separate file (Existing structure forces us to have all processing and default preprocessing params in a single file)
2) ClassBalancedSampler not has `shuffle` attribute (Not used, exists for compatibility with dataloaders.get since they forward this attr to sampler)
3) Adds `_shift_bboxes_cxcywhr` helper function


src/super_gradients/training/datasets/samplers/class_balanced_sampler.py
```diff
@@ -127,8 +127,14 @@ def __init__(
         oversample_aggressiveness: float = 0.5,
         num_samples: Optional[int] = None,
         generator=None,
+        shuffle: bool = True,  # noqa
     ) -> None:
         """
+
+        :param shuffle: This parameter is not used, exists for compatibility purposes
```
shaydeci, line 134: Maybe itd be wise to give a warning?
I don't remeber whether with the balanced sampler the repeates keep the same order or not but maybe some users would find it confusing.
Lets also set the default accordingly.

shaydeci (changes requested):
Just a minor comment

src/super_gradients/training/datasets/samplers/class_balanced_sampler.py
BloodAxe: @shaydeci reverted this change. Seems I can simply do this:
```
train_dataloader_params:
   sampler: 
      ClassBalancedSampler:
          ...
```

And not specify `shuffle` key at all and that does the job.

BloodAxe: Fixed in ee48f23f75732b8762de489f5fd30460bc480494

shaydeci (approved):
LGTM

ofrimasad (approved):
LGTM

PR #1995, opened 2024-05-17
Feature/sg 1488 yolo nas r integration sample and transforms

BloodAxe:
Adds OBBSample dataclass and transforms related to OBB sample. 
Plus Albu support


src/super_gradients/training/transforms/obb/obb_random_rotate90.py
```diff
@@ -0,0 +1,81 @@
...
+import numpy as np
+from super_gradients.common.registry import register_transform
+from .obb_sample import OBBSample
+
+from .abstract_obb_transform import AbstractOBBDetectionTransform
+
+
+@register_transform()
+class OBBDetectionRandomRotate90(AbstractOBBDetectionTransform):
+    def __init__(self, prob: float = 0.5):
```
shaydeci, line 13: I think there are some issues here...
This entire transform assumes bboxes_xyxy which is not consistent with what I see in the OBB sample.
Perhaps when you add a unit test lets pass some sample through all of these just see they run smoothly.

src/super_gradients/training/transforms/obb/obb_random_rotate90.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__()
+        self.prob = prob
+
+    def apply_to_sample(self, sample: OBBSample) -> OBBSample:
+        if random.random() < self.prob:
+            k = random.randrange(0, 4)
+            image_shape = sample.image.shape[:2]
+            sample = OBBSample(
+                image=self.apply_to_image(sample.image, k),
+                bboxes_xyxy=self.apply_to_bboxes(sample.bboxes_xyxy, k, image_shape),
```
shaydeci, line 23: See previous comment.

src/super_gradients/training/transforms/obb/obb_random_rotate90.py
```diff
@@ -0,0 +1,81 @@
...
+        :param factor:       Number of CCW rotations. Must be in set {0, 1, 2, 3} See np.rot90.
+        :param image_shape:  Original image shape
+        :return:             Rotated bounding boxes in XYXY format.
+        """
+        rows, cols = image_shape
+        bboxes_rotated = self.xyxy_bbox_rot90(bboxes, factor, rows, cols)
+        return bboxes_rotated
+
+    @classmethod
+    def xyxy_bbox_rot90(cls, bboxes: np.ndarray, factor: int, rows: int, cols: int):
```
shaydeci, line 54: See previous comment.

src/super_gradients/training/transforms/obb/obb_sample.py
```diff
@@ -0,0 +1,110 @@
+import dataclasses
+from typing import Union, List, Optional
+
+import numpy as np
+import torch
+from super_gradients.training.datasets.data_formats.obb.cxcywhr import cxcywhr_to_poly, poly_to_cxcywhr
+
+
+@dataclasses.dataclass
+class OBBSample:
```
shaydeci, line 10: We have a dedicated src/super_gradients/training/samples module which is more appropriate to place this in. Please move it there.

shaydeci (changes requested):
Some short minor comments.
The use of keypoints for rotated boxes for the Albu support is awesome! Lets just add another use case for testing in the albumentations integration test like we have for the other tasks (tests/integration_tests/albumentations_test.py).
Also, pretty sure I found some problems with the rotate90 transform, so just to be safe it might be a good idea to have just one test that forward some ample through all of the newly added transforms.

src/super_gradients/training/transforms/obb/obb_random_rotate90.py
BloodAxe: That's a good catch! I think I copy-pasted this class from DetectionRandomRotate90 but actually never used and forgot it is present 🙈 For training I used Albumentation's random 90 rotation instead.
Removed in abc4e8d6d290b62ad3d1bdd008e59f8ca5180715

src/super_gradients/training/datasets/data_formats/obb/cxcywhr.py
```diff
@@ -0,0 +1,58 @@
...
+
+def poly_to_cxcywhr(poly: np.ndarray) -> np.ndarray:
+    shape = poly.shape
+    if shape[-2:] != (4, 2):
+        raise ValueError(f"Expected last two dimensions to be (4, 2), got {shape[-2:]}")
+
+    flat_polys = poly.reshape(-1, 4, 2)
+    rboxes = np.zeros((flat_polys.shape[0], 5), dtype=np.float32)
+    for i, poly in enumerate(flat_polys):
+        hull = cv2.convexHull(np.reshape(poly, [-1, 2]).astype(np.float32))
```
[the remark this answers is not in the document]
BloodAxe: I think I left this as a precaution. For instance, A.ElasticTransform can move (in theory) individual points such that a polygon is not convex anymore and contain self-intersections.
Additionally, a horizontal flip can change the polygon order (CW <-> CCW).

A convex hull would cover it and additionally it would return a consistent keypoints orientation (CCW) for a hull, which in turn would ensure consistent rotated box parametrization for boxes regardless of points order.

src/super_gradients/training/transforms/obb/obb_sample.py
BloodAxe: Thanks! I forgot we have this folder :/
Fixed in 874119c2f1f7e2855eff8c8b9e07518223f90317

shaydeci (approved):
LGTM

PR #1999, opened 2024-05-21
Added DOTA2 dataset

BloodAxe:
This PR adds 
- DOTA2 dataset implementation 
- Script to prepare dataset files.
- Metric implementation (mAP for rotated boxes)

### DIFF
--- a/src/super_gradients/examples/dota_prepare_dataset/dota_prepare_dataset.py
+++ b/src/super_gradients/examples/dota_prepare_dataset/dota_prepare_dataset.py
@@ -1,0 +1,47 @@
+import argparse
+from pathlib import Path
+
+import cv2
+from super_gradients.training.datasets import DOTAOBBDataset
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Slice DOTA dataset into tiles of usable size for training a model")
+    parser.add_argument("--input_dir", help="Where the full coco dataset is stored", required=True)
+    parser.add_argument("--output_dir", help="Where the resulting data should be stored", required=True)
+    parser.add_argument("--num_workers", default=cv2.getNumberOfCPUs() // 2)
+    args = parser.parse_args()
+    ann_subdir_name = "ann-obb"
+
+    cv2.setNumThreads(cv2.getNumberOfCPUs() // 4)
+
+    input_dir = Path(args.input_dir)
+    output_dir = Path(args.output_dir)
+
+    DOTAOBBDataset.slice_dataset_into_tiles(
+        data_dir=input_dir / "train",
+        output_dir=output_dir / "train",
+        ann_subdir_name=ann_subdir_name,
+        tile_size=1024,
+        tile_step=512,
+        scale_factors=(0.75, 1, 1.25),
+        min_visibility=0.4,
+        min_area=8,
+        num_workers=args.num_workers,
+    )
+
+    DOTAOBBDataset.slice_dataset_into_tiles(
+        data_dir=input_dir / "val",
+        output_dir=output_dir / "val",
+        ann_subdir_name=ann_subdir_name,
+        tile_size=1024,
+        tile_step=1024,
+        scale_factors=(1,),
+        min_visibility=0.4,
+        min_area=8,
+        num_workers=args.num_workers,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/src/super_gradients/examples/dota_prepare_dataset/dota_prepare_dataset.py
+++ b/src/super_gradients/examples/dota_prepare_dataset/dota_prepare_dataset.py
@@ -1,3 +1,31 @@
+"""
+This script slices the DOTA dataset into tiles of a usable size for training a model.
+The tiles are saved in the output directory with the same structure as the input directory.
+
+To use this script you should download the DOTA dataset from the official website:
+https://captain-whu.github.io/DOTA/dataset.html
+
+The dataset should be organized as follows:
+    dota
+    └── train
+        ├── images
+        │   ├─ 000000000001.jpg
+        │   └─ ...
+        └── ann
+            └─ 000000000001.txt
+    └── val
+        ├── images
+        │   ├─ 000000000002.jpg
+        │   └─ ...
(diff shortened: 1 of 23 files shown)


src/super_gradients/training/datasets/obb/dota.py
```diff
@@ -0,0 +1,340 @@
...
+from super_gradients.training.transforms import OBBDetectionCompose
+from super_gradients.training.transforms.obb import OBBSample
+from torch.utils.data import Dataset
+from super_gradients.common.factories.transforms_factory import TransformsFactory
+
+__all__ = ["DOTAOBBDataset"]
+
+
+@register_dataset()
+class DOTAOBBDataset(Dataset, HasPreprocessingParams, HasClassesInformation):
```
shaydeci, line 27: Does inheriting from detectiondataset not make sense ?
In case it doe not, I believe we should introduce a base class for OBB at least.

src/super_gradients/training/datasets/obb/dota.py
```diff
@@ -0,0 +1,340 @@
...
+                min_area=min_area,
+                output_images_dir=output_images_dir,
+                output_ann_dir=output_ann_dir,
+                output_image_ext=output_image_ext,
+            )
+            for _ in tqdm(wp.imap_unordered(worker_fn, payload), total=len(payload)):
+                pass
+
+    @classmethod
+    def _worker_fn(cls, args, tile_size, tile_step, min_visibility, min_area, output_images_dir, output_ann_dir, output_image_ext):
```
shaydeci, line 307: Im not really following what's going on here...

src/super_gradients/training/metrics/obb_detection_metrics.py
```diff
@@ -0,0 +1,631 @@
...
+        :param device:          Device to run on
+        :param inputs:          Input image tensor of shape (batch_size, n_img, height, width)
+        :param crowd_targets:   Crowd targets for all images of shape (total_num_targets, 6), LABEL_CXCYWH
+        """
+        preds: List[OBBPredictions] = self.post_prediction_callback(preds)
+        output_device = "cpu" if self.accumulate_on_cpu else None
+        matching_strategy = OBBIoUMatching(self.iou_thresholds.to(preds[0].scores.device))
+
+        for pred, trues in zip(preds, gt_samples):
+            image_mathing = compute_obb_detection_matching(
```
shaydeci, line 475: Isn't update the only different method between this implementation and the original DetectionMetrics?
Seems to me like an overkill to re-write the entire metric...
We can probably even shrink it even more (not gonna be annoying about this though) and extract the matching parts + iou claculations to some methods. But only if it isn't too much work.

src/super_gradients/examples/dota_prepare_dataset/dota_prepare_dataset.py
```diff
@@ -0,0 +1,47 @@
+import argparse
+from pathlib import Path
+
+import cv2
+from super_gradients.training.datasets import DOTAOBBDataset
+
+
+def main():
```
shaydeci, line 8: An example is good.
However I think it's best to add some basic instructions like we have for the rest of our datasets in our src/super_gradients/training/datasets/Dataset_Setup_Instructions.md

Side note - we should probably move it to the docs directory (I guess that might be the reason you missed it).

src/super_gradients/training/datasets/obb/collate.py
```diff
@@ -0,0 +1,33 @@
...
+            all_crowd_masks.append(torch.from_numpy(sample.is_crowd.reshape((-1, 1))))
+            sample.image = None
+
+        images = torch.stack(images)
+
+        boxes = flat_collate_tensors_with_batch_index(all_boxes).float()
+        labels = flat_collate_tensors_with_batch_index(all_labels).long()
+        is_crowd = flat_collate_tensors_with_batch_index(all_crowd_masks)
+
+        extras = {"gt_samples": batch}
```
shaydeci, line 32: I am missing some documentation here so its a bit hard to follow...
Why are we returning extras = {"gt_samples": batch} ?
Can't we just use the regular detection collate ?
Is the reason we don't because the detection collate works on the tensors rather then the sample but have the exact same functionality ?

src/super_gradients/training/datasets/obb/dota.py
```diff
@@ -0,0 +1,340 @@
...
+        is_crowd = difficult.reshape(-1) if self.difficult_labels_are_crowd else np.zeros_like(difficult, dtype=bool)
+        sample = OBBSample(
+            image=image,
+            rboxes_cxcywhr=cxcywhr.reshape(-1, 5),
+            labels=classes.reshape(-1),
+            is_crowd=is_crowd,
+        )
+        return sample.sanitize_sample()
+
+    def __getitem__(self, index) -> OBBSample:
```
shaydeci, line 88: This is not consistent with how we work in DetectionDataset, where we return tensors and the collate works on them.
This is fine of course, but I would like to understand why, since like I stated earlier I didn't understand why we need to untroduce new collate.

shaydeci (changes requested):
Some initial comments - inline.
Also, we are missing some tests...

src/super_gradients/examples/dota_prepare_dataset/dota_prepare_dataset.py
BloodAxe: Fixed in 619e26bfa69c20fb83a76a10c113f618d185844c

src/super_gradients/training/metrics/obb_detection_metrics.py
BloodAxe: Fixed in 75a47e1761d8c7131436532ec0d39a9c2ea4f0fb

src/super_gradients/training/datasets/obb/dota.py
BloodAxe: Fixed in 3427148c1f8114a7252e2f0597cd746de00d2bb9

shaydeci (approved):
LGTM

PR #2007, opened 2024-05-28
Moved logic of update of _best_ckpt_metrics before we build state dict

BloodAxe:
This PR fixes a bug when _best_ckpt_metrics key was not updated in `state` dictionary of saved checkpoint


shaydeci (approved):
LGTM

Issue #2011, opened 2024-05-31
DetectionRandomAffine target-size is in wrong format


cansik:
### 🐛 Describe the bug

During training of a detection model with `input_dim` not a 1:1 ratio (640x480 = 4:3), I experience a strange behaviour of the [DetectionRandomAffine](https://github.com/Deci-AI/super-gradients/blob/217353a3a2fc919d2330d9c74f78e40d96d72b35/src/super_gradients/training/transforms/transforms.py#L603) transform. It seems that even when all parameters are set to have no change on the actual image, the image seems translated after the transform:

```yaml
dataset_params:
  train_dataset_params:
    input_dim: [ 480, 640 ]

train_transforms:
  - DetectionRandomAffine:
      # rotation degrees, randomly sampled from [-degrees, degrees]
      degrees: [ 0, 0 ]
      # image translation fraction
      translate: [ 0.5, 0.5 ]
      # random rescale range (keeps size by padding/cropping) after mosaic transform.
      scales: [ 1.0, 1.0 ]
      # shear degrees, randomly sampled from [-degrees, degrees]
      shear: [ 0.0, 0.0 ]
      target_size: ${dataset_params.train_dataset_params.input_dim}
      # whether to filter out transformed bboxes by edge size, area ratio, and aspect ratio.
      filter_box_candidates: false
  - DetectionTargetsFormatTransform:
      input_dim: ${dataset_params.train_dataset_params.input_dim}
      output_format: LABEL_CXCYWH
```

<img width="845" alt="image" src="https://github.com/Deci-AI/super-gradients/assets/5220162/650d1137-f18a-426e-8ace-2741e2550e5f">

*Figure 1: First 4 images of coco with wrong aspect ratio*

It looks like (*Figure 1*) the aspect ratio of the image is wrong, so I changed the `target_size` of the `DetectionRandomAffine` to `[640, 480]`:

```yaml
train_transforms:
  - DetectionRandomAffine:
      target_size: [640, 480]
```

<img width="890" alt="image" src="https://github.com/Deci-AI/super-gradients/assets/5220162/cb621bb6-03ae-4a90-a200-045afba4a0a6">

*Figure 2: First 4 images of coco with correct aspect ratio but wrong translation*

### Solution

[...]

PR #2012, opened 2024-05-31
Fix DetectionRandomAffine target-size to be in row-column format

cansik:
This fixed the `target-size` for DetectionRandomAffine as explained in #2011. It already takes into account if a specific `target-size` is set or the image shape is used.

### DIFF
--- a/src/super_gradients/training/transforms/transforms.py
+++ b/src/super_gradients/training/transforms/transforms.py
@@ -609,7 +609,7 @@ class DetectionRandomAffine(AbstractDetectionTransform, LegacyDetectionTransform
                                     (center-translate, center+translate)
      :param scales:                 Values for random rescale, when float the random values are drawn uniformly from (1-scales, 1+scales)
      :param shear:                  Degrees for random shear, when float the random values are drawn uniformly from (-shear, shear)
-     :param target_size:            Desired output shape.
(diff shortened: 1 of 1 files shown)


src/super_gradients/training/transforms/transforms.py
```diff
@@ -1510,7 +1510,7 @@ def random_affine(
 
     M = get_affine_matrix(img.shape[:2], target_size, degrees, translate, scales, shear)
 
-    img = cv2.warpAffine(img, M, dsize=target_size, borderValue=(border_value, border_value, border_value))
+    img = cv2.warpAffine(img, M, dsize=target_size[::-1], borderValue=(border_value, border_value, border_value))
```
BloodAxe, line 1513: I suggest to add an explicit message to a method's docstring to indicate whether `target_size` represents size as (rows, cols) or (cols, rows).
And unpack these to add more clarity to the code. Eg:
```
rows, cols = target_size[:2]
img = cv2.warpAffine(img, M, dsize=(cols, rows), ...)

```

cansik: Good point, I've adjusted it.

BloodAxe (approved):
LGTM

BloodAxe:
Thank you for your contribution! Merged to master 🎉

